PR #1, opened 2015-08-26
Remove superfluous line

stig:
Another possibility is that the full stop on the previous line should be removed.


mnencia:
Thanks for the pull request. The documentation has just been converted to Markdown fixing this issue.

PR #7, opened 2015-11-05
Add a sleeptime argument

florentx:
It lowers the load on the network for machines which can stay few minutes behind the master server.

A similar option can be found in the upstream `pg_standby` tool:
http://www.postgresql.org/docs/9.4/interactive/pgstandby.html


florentx:
Note: the Travis builder is red for an unrelated reason. See issue #10

PR #15, opened 2015-12-05
use python internal modules for gzip and bzip2

moench-tegeder:
this avoids forking and shell execution, and simplifies
fsyncing of compressed files for these common compression
cases.


gbartolini:
Merged into d7fdf67ee0fc7a34d911bed188175a1840271dac

PR #18, opened 2015-12-09
add native support for the pigz compressor

zacchiro:
The same result achieved by this commit can be achieved, using today's barman, specifying custom (de)compressor filters. And it works just fine.

But in write-intense workloads (been there, done that) the difference between workloads might be relevant enough (it's up to you to judge, of course) to justify having *native* pigz support rather then requiring user to customize it using filter.


gbartolini:
Closed by 3435c097ba91f3f644748566f4e43acd3b34539f

Grazie Stefano! ;)

PR #23, opened 2016-01-08
allow archive_mode=always setting

moench-tegeder:
archive_mode always is new in PostgreSQL 9.5.


gbartolini:
Part of cdf77eb2b3177e86feabf5fe6ec54eb0e89f397f

PR #26, opened 2016-02-02
Update docs - Add example command on remote recovery section

aydantasdemir:


mnencia:
Thanks for the contribution

PR #33, opened 2016-03-10
Added the -p flag to recursively add parent directories that do not exist

funzie19:
I added the -p flag to resolve the issue when restoring and I have tablespaces two levels deep. For example here is a failure:

```
EXCEPTION: unable to prepare 'drupal_dbspace' tablespace (destination '/usr/local/pgsql/data/tablespaces/drupal_dbspace'): mkdir execution failed
```

In the above case the table space lives under `/usr/local/pgsql/data/tablespaces`. Currently Barman does not support this.

Adding the `-p` allows for the parent directories to be created. In my test case it created the tablespaces correctly
```
Destination directory: /usr/local/pgsql/data
        27837, drupal_dbspace, /usr/local/pgsql/data/tablespaces/drupal_dbspace
        27838, drupal_indexspace, /usr/local/pgsql/data/tablespaces/drupal_indexspace
        27839, drupal_zlog_dbspace, /usr/local/pgsql/data/tablespaces/drupal_zlog_dbspace
Copying the base backup.
```

### DIFF
--- a/barman/fs.py
+++ b/barman/fs.py
@@ -73,7 +73,7 @@ class UnixLocalCommand(object):
             else:
                 return False
         else:
-            mkdir_ret = self.cmd('mkdir %s' % dir_path)
+            mkdir_ret = self.cmd('mkdir -p %s' % dir_path)
             if mkdir_ret == 0:
                 return True
             else:


gcalacoci:
Hi!

the `-p` flag was something that purposely is not used during a recovery.

For a matter of safety and security of the server where the recovery is happening, we want the users to be responsible for the creation destination directories.

Regards

funzie19:
In see your point. Let me explain the issue and perhaps there might be a way to solve it. Creating the restore location is perfectly fine. The problem is when having tablespaces that are within another directory. 

Here is my setup:
```
/usr/local/pgsql/data
/usr/local/pgsql/tablespaces
/usr/local/pgsql/archive_spaces/orders
/usr/local/pgsql/archive_spaces/users
```

The above in a restore works fine until it reaches the `/orders and /users` tablespaces which is one level above. It stops the restore because `/usr/local/pgsql/archive_spaces` does not exists as the mkdir will try to create `/usr/local/pgsql/archive_spaces/users`. That's where the `-p` flag comes in.

Perhaps a restore option to allow automatic parent directory creation or make it smart enough to know it needs to create a parent directory.

gcalacoci:
As said before, we want the users to be responsible for the creation of destination directories. 
This includes tablespaces directories.

To solve your problem you simply need to execute `mkdir -p /usr/local/pgsql/archive_spaces`  by yourself, then barman will take care of the rest during the recovery.

Basically if you have the right to a perform a recovery you have the right to execute a mkdir, even through ssh.

PR #34, opened 2016-03-12
mark "failed backups" check as non-fatal

moench-tegeder:
allows taking a backup when the previous backup has failed

After one failed backup, further backups are denied with message
barman.server ERROR: Check 'failed backups' failed for server '<server>'
That's not helpful, as I'd rather take a fresh backup soon after the process failed for some reason (e.g. network issues, server down, ...)


gbartolini:
We had noticed this issue last Friday evening and were going to act this week, starting with a bug submission. You anticipated us. 

I have applied your patch here: 72ba9e4d0afe366d68954eb84ae8381cfdf58c5e

Issue #32, opened 2016-03-09
Problem with archive_mode = always setting in PG 9.5


koichi-szk:
Barman does not work with archive_mode = always setting.   It is needed to archive WAL for cascading slave or warm standby setting from a slave.

I had the following message with this setting:
```
2016-03-08 05:56:31 UTC [2181-1] postgres@postgres ERROR:  invalid input syntax for type boolean: "always"
2016-03-08 05:56:31 UTC [2181-2] postgres@postgres STATEMENT:  SELECT *, current_setting('archive_mode')::BOOLEAN AND (last_failed_wal IS NULL OR last_failed_wal LIKE '%.history' AND substring(last_failed_wal from 1 for 8) <= substring(last_archived_wal from 1 for 8) OR last_failed_wal <= last_archived_wal) AS is_archiving, CAST (archived_count AS NUMERIC) / EXTRACT (EPOCH FROM age(now(), stats_reset)) AS current_archived_wals_per_second FROM pg_stat_archiver
2016-03-08 05:56:31 UTC [2181-3] postgres@postgres ERROR:  current transaction is aborted, commands ignored until end of transaction block
2016-03-08 05:56:31 UTC [2181-4] postgres@postgres STATEMENT:  SELECT count(*) FROM pg_extension WHERE extname = 'pgespresso'
```

This is because trying to convert archive_mode into boolean.   Archive_mode changed from boolean to enum at PG 9.5.

I'd suggest to correct postgres.py, line from 380 to 391, as well as its caller.

gbartolini:
Hi Koichi,

  It's such a pleasure to see you using Barman. We do hope you find it useful. Could you please let us know which version of Barman you are using?

  This behaviour should have already been fixed in version 1.6.0 (see commit:cdf77eb2b3177e86feabf5fe6ec54eb0e89f397f). If you are using a previous version, please update to 1.6.0 (it is safe to do it) and let us know if the problem persists. Otherwise, it seems that the bug still exists.

Thanks,
Gabriele

gbartolini:
Hang on ... I think I have spotted the issue. I confirm this is a bug in the get_archiver_stats function of the postgres.py module, issued by "barman status".

koichi-szk:
I'm using 1.6.0.   As you suggested, wal_archiver.pl seems to handle
"always" value correctly but postgres.py does not pass this value
correctly.    Attached is my quick hack for the fix.   My colleague will
[...]

PR #35, opened 2016-03-13
unbreak backups when archive_mode is "always"

moench-tegeder:
cannot cast 'always' to BOOLEAN, this will break the whole chain.
the error handling here needs some attention, too

The query in question tried to read the archive_mode setting into a boolean variable by casting it as a BOOLEAN in the first place. While being a nice trick, this is not safe: archive_mode is not a boolean parameter anymore in 9.5.

Errors from the database connection are only logged at DEBUG level and then ignored, but as psycopg does not use autocommit by default (as it would be the sane thing to do) all later statements will ERROR out. I haven't checked yet if it's safe to force autocommit here.

### DIFF
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -378,7 +378,9 @@ class PostgreSQLConnection(PostgreSQL):
             # also evaluating if the server is archiving without issues
             # and the archived WALs per second rate
             cur.execute(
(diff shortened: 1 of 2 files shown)


gbartolini:
Duplicates #32

gbartolini:
Already fixed by ﻿﻿5d34d6a56564b49d6dbe8a6ec6daee25752698c3

PR #43, opened 2016-06-09
Add note about switch-xlog --force for new 1.6.1 installs

jamonation:


mnencia:
The documentation has been completely rewritten in Barman 2.0, so this patch is not applicable.

PR #49, opened 2016-08-19
Split the connstring on whitespaces intead of one space

chtitux:
With `.split(' ')`, the connstring can't contain multiple spaces or tab between the arguments.
Therefore, Barman does not consider `host=localhost  port=5432` as valid.

The postgresql libraries do accept multiple whitespaces between parameters, so passing `"host=localhost\tport=5432      user=postgres"` is valid.

Using `.split()` will not reject valid conninfo string.


chtitux:
Fixed by 7c602fdd924a8637a4a963effbc7d46bf084a680

PR #51, opened 2016-08-29
Enable barman-wal-restore.py to pass -c/--config to Barman

ibarwick:
This enables barman-wal-restore.py to be used if the Barman server
uses a non-default configuration file.


gbartolini:
Merged in fcb0d595c3303c2fe5f94c740888d4006033665b

PR #52, opened 2016-09-19
update barman.5, add conninfo information.

csawyerYumaed:
Added link to libpq documentation, and give common keys to use in conninfo strings.


csawyerYumaed:
Feel free to modify/replace my documentation if you do not like it, but I felt like the docs needed some help around conninfo strings, and how to use them.

PR #53, opened 2016-09-19
update barman.5.md - add conninfo

csawyerYumaed:
add additional information to documentation around conninfo


csawyerYumaed:
Updated MD file instead.

PR #98, opened 2017-04-03
Clarifications on `cron` command in docs

mgalgs:
It can be a bit surprising to discover that barman doesn't have any sort of
daemon program, as many long-running services do.  Make this clear up front
to avoid confusion.

Also, it's not immediately clear from the docs how to stop all of the processes
that get started by the `cron` command.  Add some instructions to the
section on `cron` documenting how to stop things properly.

### DIFF
--- a/doc/manual/41-global-commands.en.md
+++ b/doc/manual/41-global-commands.en.md
@@ -15,6 +15,11 @@ The following list includes the general commands.
 
 ## `cron`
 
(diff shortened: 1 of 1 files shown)


gbartolini:
commit 9c236c16382e313963d6213455a9d6f9b4b1ae3c
commit c20cd2bf18ac872ee77c23237af42dc37761e437

PR #106, opened 2017-05-18
init ru

bessonovevgen:
Hi, I want to help to translate documentation to Russian language. It is help other people from Russia to use barman for backup PostrgeSQL server.


martinmarques:
We don't have intentions of having documentation in other languages, as this would bring in the need to keep them up to date.
We suggest having this in a separate place and make sure to update it with each release.

Issue #111, opened 2017-07-14
Barman 'backup' Errors Not Properly Logged in 2.1


damnski:
Howdy,

I posted this on the mailing list, and wanted to verify it got logged properly.

We are running barman 2.1 on centos 7 with packages from https://yum.postgresql.org/9.6/redhat/rhel-7-x86_64/.  We are using PG96 + pgespresso to do basebackups on a replica.  Since we started running pgbarman 2.1, we are not seeing 'backup' errors properly logged; instead we see an error trace, "'NoneType' object has no attribute 'find'".  If we turn on log_level=DEBUG we see the error trace and the actual error message.   We previously were using barman 1.6.1 and the error showed in the log file with default log_level=INFO.

The error is "canceling statement due to statement timeout" as we are attempting to perform a backup on a replica, which is a known case that happens on replicas, but that is secondary to not logging the error without configuring log_level to debug.

Error log:

2017-06-27 00:01:01,732 [64963] barman.config DEBUG: Including configuration file: exampleserver.conf
2017-06-27 00:01:01,733 [64963] barman.cli DEBUG: Initialised Barman version 2.1 (config: /etc/barman.conf, args: {'reuse_backup': 'link', 'server_name': ['exampleserver'], 'format': 'console', 'quiet': True, 'retry_times': 10, 'command': 'backup', 'retry_sleep': 60, 'debug': False})
2017-06-27 00:01:01,735 [64963] barman.server DEBUG: Retention policy for server exampleserver: RECOVERY WINDOW OF 3 DAYS
2017-06-27 00:01:01,735 [64963] barman.server DEBUG: WAL retention policy for server exampleserver: MAIN
2017-06-27 00:03:02,462 [64963] barman.postgres DEBUG: Error retrieving PostgreSQL total size: canceling statement due to statement timeout
2017-06-27 00:03:02,463 [64963] barman.postgres DEBUG: Error retrieving PostgreSQL configuration files location: current transaction is aborted, commands ignored until end of transaction block
2017-06-27 00:03:02,463 [64963] barman.postgres DEBUG: Error retrieving PostgreSQL setting 'synchronous_standby_names': current transaction is aborted, commands ignored until end of transaction block
2017-06-27 00:03:02,473 [64963] barman.cli ERROR: 'NoneType' object has no attribute 'find'
See log file for more details.
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/barman/cli.py", line 1041, in main
[...]

PR #115, opened 2017-07-21
Fix for #111

steamraven:
postgres.get_setting() returns None on error.   This fix catches it and returns the appropriate empty list


mnencia:
merged with some changes  823acdb

PR #131, opened 2017-10-27
Add --wal-method option to pg_basebackup ver>=10.

S3RK:
By default pg_basebackup 10 launch additional wal sender process to include
all WAL generated during backup into the base bakcup.
Because Barman already stores full WAL stream we don't need it.


S3RK:
The need to eliminate additional wal sender process is even higher in the light of this issue https://www.postgresql.org/message-id/20160909165846.20024.16221%40wrigleys.postgresql.org
since we already use `--no-slot` option

S3RK:
Fixed in #133

PR #137, opened 2018-01-12
Added ability to run scripts pre/post-recovery

seidelma:
This commit adds the ability to run scripts pre- and post-recovery, and to exclude files/directories from rsync operations.

We use a post-backup script to compress all our backups with xz, and this allows us to recover them without decompressing the backup files manually first. We've been using it in production for nearly a year now, and it works fine.

We also store a massive amount of data with the cstore_fdw foreign data wrapper, and backups would be impossible if we allowed rsync to copy that every time a backup was performed. The `rsync_exclude` option can be used to prevent files/directories from being rsynced.

Tested and working with Python 2.7.14 and 3.6.4.


mnencia:
merged as 0139b7a

PR #142, opened 2018-03-01
FIX obsolete backups check

eLBati:
We need to check obsolescence immediatly, otherwise obsolete backup, typically the last one of the for cycle, will not be reported as obsolete

### DIFF
--- a/barman/retention_policies.py
+++ b/barman/retention_policies.py
@@ -257,6 +257,10 @@ class RecoveryWindowRetentionPolicy(RetentionPolicy):
         for bid in sorted(backups.keys(), reverse=True):
             # We are interested in DONE backups only
             if backups[bid].status == BackupInfo.DONE:
+                # TODO: Currently we use the backup local end time
+                # We need to make this more accurate
+                if backups[bid].end_time < self._point_of_recoverability():
+                    found = True
                 if found:
                     # Check minimum redundancy requirements
                     if valid < self.server.config.minimum_redundancy:
@@ -291,10 +295,6 @@ class RecoveryWindowRetentionPolicy(RetentionPolicy):
                     report[bid] = BackupInfo.VALID
                     self._first_backup = bid
                     valid = valid + 1
-                    # TODO: Currently we use the backup local end time
-                    # We need to make this more accurate
-                    if backups[bid].end_time < self._point_of_recoverability():
-                        found = True
             else:
                 report[bid] = BackupInfo.NONE
         return report


gbartolini:
Hi, can you please elaborate on the use case of your proposal? Indeed, it does not resonate with the way we have developed both automated retention policies and minimum redundancy. A backup cannot be considered obsolete if there are not enough backups according to minimum redundancy.

eLBati:
@gbartolini consider you have 2 backups, 1 obsolete and 1 valid.
Without this changes, the first cycle, for the valid backup, ```found``` is False, ```backups[bid].end_time``` is > ```self._point_of_recoverability()``` and ```found``` remains False.
Second cycle, for the obsolete backup, ```found``` is False and the backup is reported as VALID, while it is not

gbartolini:
Ok, that explains and confirms that the we need to discard the pull request. Remember that you always need a backup to restore - and that is what having the check at the end of the loop grants. The purpose of that piece of code is to find the **first** available backup prior to the point of recoverability, from which to start replay WAL files. That is the only way you can recover at the point in time. And, that backup cannot therefore be considered obsolete according to *recovery window* retention policies. I hope that's clearer.

eLBati:
@gbartolini how can this change produce a "0 backup" situation? This is not what it does.

Also, is it correct, as described by the above use case, to keep 2 backups when ```minimum_redundancy``` is 1 and you 1 have 1 valid backup and 1 obsolete backup? Only 1 backup should be kept in that case

gbartolini:
Consider this scenario.:

- Current time: 11.59AM
- Recovery window: 1 DAY
- point of recoverability: yesterday at 11.59 (PoR from now on)
- minimum redundancy: 1

You have got:

- yesterday's backup, finished at 12.00PM (for simplicity, Backup ID = DAY2)
- the day before yesterday's backup, finished at 12.00PM (for simplicity, Backup ID = DAY1)

For now let's consider the current stable code of Barman.

We need to find the first available backup that has an end time lower than yesterday's at 11.59 (current PoR) and we cycle through the list of backups in reverse order (that is the portion of code in RecoveryWindowRetentionPolicy class, _backup_report() method).

The first one we iterate is backup DAY2. It is a valid backup, but because its end time is after our PoR, the 'found' variable is left unchanged. The backup is classified as VALID.

We keep iterating and we hit DAY1 backup. It is a valid backup and, because its end time is before our PoR, the 'found' variable is left unchanged. This backup is classified as VALID and marked as the first valid one. Without this backup we cannot meet the recovery window retention policy requirements. We exit the loop as no more backups are present.

Following your approach, when we hit DAY1, we immediately set the found variable and we erroneously enter the minimum redundancy checks. Because we already have a "valid" backup based on redundancy, we classify DAY1 backup as OBSOLETE. This is **BAD** and we do not want to remove the only valid backup we have in order to recover at PoR.

If we repeat the operation at 12.02PM, you will see that DAY2 becomes the first available backup and DAY2 is correctly classified as obsolete. Please ack the closure of this issue.

Issue #182, opened 2018-09-19
Using `archive_timeout` to prevent corner case with concurrent backup


gciolli:
We could suggest using this parameter to ensure enough WAL file switching even when using concurrent backups.

mnencia:
It will not work. The corner case can only happen in the first backup when no previous backup is available, and Barman has been just configured. There is no way to prevent that other than failing the backup as soon as we can detect that condition, that unfortunately is at the end of the backup.

PR #183, opened 2018-09-19
Improved comments on concurrent backups (#182)

gciolli:
Added a note suggesting the use of `archive_timeout` to avoid a corner
case, and slightly rephrased an existing footnote.

Signed-off-by: Gianni Ciolli <gianni.ciolli@2ndQuadrant.com>

### DIFF
--- a/doc/manual/50-feature-details.en.md
+++ b/doc/manual/50-feature-details.en.md
@@ -172,14 +172,38 @@ streaming replicated standby server.
 > above)[^CONCURRENT_ARCHIVING].
 
 [^CONCURRENT_ARCHIVING]:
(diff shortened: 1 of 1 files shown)


mnencia:
I'm against implementing this. The archive_timeout setting has a different purpose than making the backup consistent when the server has no write load. Setting that parameter right is not in the scope of barman documentation. We can, at most, warn the user that backups taken from standbys could not be immediately valid.

gbartolini:
I think the patch should be rewritten considering that commit c44ef7dc181e982d080bede92e63817910b7c4fa fixes the corner case. However the context information would improve the documentation. Could you please rewrite the docs considering this? Thanks

PR #203, opened 2019-03-29
Recovery Retry Script doesn't get passed the recovery environment

dragonfyre13:
The current recovery retry script doesn't get passed the recovery environment due to a typo in backup.py, it's reapplying the environment to the original recovery_script's "script" variable, rather than the "retry_script" variable.

I'm creating a pull request to resolve, but this seems a pretty straightforward bug.


dragonfyre13:
Fixed the flake8 errors with line length, should be good to go now.

PR #210, opened 2019-04-07
Improve performance of date parsing

jeffjanes:
Looking up the local timezone once for each src file, and not using
a default object for dateutil.parser.parse both lead to redundant
system calls.  This creates a small but noticable slowdown when
backing up a cluster with a very large number of unchanged files.

### DIFF
--- a/barman/copy_controller.py
+++ b/barman/copy_controller.py
@@ -948,7 +948,7 @@ class RsyncCopyController(object):
         # Cache and reuse datetime parsing objects for speed
         tzlocal = dateutil.tz.tzlocal()
         default = datetime.datetime.now().replace(hour=0, minute=0, second=0,
-                                                  microsecond=0, 
+                                                  microsecond=0,
                                                   tzinfo=tzlocal)
         for line in rsync.out.splitlines():
             line = line.rstrip()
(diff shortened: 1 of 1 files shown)


mnencia:
I've written a small test program to verify the issue, and it looks like that following the suggested path suggested we could get a time reduction of about 10%.
However, the regexp validated the date format, so we can parse it securely using a strptime call. This approach gives a reduction of about 80%.

In the attached test, the file list of an almost empty instance (973 files) is processed100 times. This operation is repeated twice (once for the local reference directory and once for the remote source) during an incremental backup, and on a system with millions of files could require several minutes with the current code (on my laptop, one million files requires about 4 minutes to parse, so doing it twice takes about 8 minutes)

With the strptime approach, it would require about 20 seconds per run, 40 seconds in total.

The test output is:

~~~
$ python test_list_files.py
_list_files_old: 18.748609066
_list_files_new: 17.0424120426 (-9.10039255384%)
_list_files_strptime: 2.75374603271 (-85.3122649098%)

$ python3 test_list_files.py
_list_files_old: 10.679487453075126
_list_files_new: 9.039590489002876 (-15.355577421461799%)
_list_files_strptime: 1.997356345062144 (-81.29726399475275%)
~~~

The test is attached as txt file: [test_list_files.txt](https://github.com/2ndquadrant-it/barman/files/3130555/test_list_files.txt)

PR #216, opened 2019-04-12
Fix "Backup completed" elapsed time

jeffjanes:
The message issued when a backup completes does not include
all of the elapsed time, but rather only a non-intuitive
subset of it. Notably it does not include the time taken
to fsync all of the files, which can be substantial. Make
the message include all time up to when the message is
issued.


mnencia:
closed in 36c47d8

PR #229, opened 2019-06-12
enable python3 for bash completion

frankgard:
This modified bash completion snippet should work for Python3 (preferred, if installed) and Python2 (as a fallback)


mnencia:
merged in b9d9c2f9c8ce83e0daa60dc95f8da61496a0847d

PR #253, opened 2019-11-17
Fix typos

Gelma:
Thanks a lot guys for your work!

Here's a stupid set of fix for typos.


Gelma:
Looking at log, I would say the test failure is not related to my patch!

mnencia:
The error in the test is because one of your fixes made a link longer than 79 characters, but that's not a problem. Unfortunately, we cannot apply the fixes to the Changelog file because it is generated from git commits.

I will merge the changes soon. 

Thanks for your contribution.

PR #266, opened 2020-02-11
Add zstd compression via command line utility only.

dkacar-oradian:
Zstd is a lot faster than gzip or bzip2, so it should be an option. This patch adds zstd compression via command line utility only because I don't know what's the policy on dependencies to external python libraries. I'm guessing it's not very desirable.

Currently zstd (or any other) compression can be used via custom compression setting, but there's a race problem with that. It seems (to me) that when custom compression is turned on all unidentified WAL files are treated as compressed with custom compression. However, files which are still in the `streaming` directory aren't compressed at all. So if a client requests them (my clients do, this is not theoretical) then the decompression utility spits an error and Postgres doesn't start.

So I'd like to add zstd compression as a first class citizen.

### DIFF
--- a/barman/compression.py
+++ b/barman/compression.py
@@ -344,6 +344,20 @@ class PyBZip2Compressor(InternalCompressor):
         return bz2.BZ2File(name, mode='rb')
 
 
+class ZstdCompressor(CommandCompressor):
+    """
+    Predefined compressor with Zstd
+    """
+
+    MAGIC = b'\x28\xb5\x2f\xfd'
+
+    def __init__(self, config, compression, path=None):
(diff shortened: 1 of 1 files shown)


artizirk:
Custom compression bug referenced here seems to be fixed in https://github.com/EnterpriseDB/barman/pull/588 but native `zstd` support would still be nice to have.

martinmarques:
@dkacar-oradian could you rebase this on top of master?

martinmarques:
We have a few internal stories that work on adding new compression algorithms, but we want to deliver them to the cloud and the on-prem barman archiver. This PR solves half of the feature

PR #271, opened 2020-03-27
Remove strict superuser requirement for PG 10+

kmoppel:
For some time already it's possible to pull backups without superuser -
which is actually very desired for security sensitive environments.
Non-superuser backups need quite some roles / grants though - this is
documented now in the 21-preliminary_steps.en.md doc.

### DIFF
--- a/barman/exceptions.py
+++ b/barman/exceptions.py
@@ -240,6 +240,12 @@ class PostgresSuperuserRequired(PostgresException):
     """
 
 
+class PostgresSuperuserOrBackupFunctionsAccessRequired(PostgresException):
+    """
+    Superuser or pg_start_backup(), pg_switch_wal(), ... etc access is required
+    """
+
+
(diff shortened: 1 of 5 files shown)


kmoppel:
There's also an issue on that: https://github.com/2ndquadrant-it/barman/issues/213

kmoppel:
The only visual change for users is `barman check` output where instead of `is_superuser` check there is `user is superuser or has grants to replication and backup functions`:
Screenshot from 2020-03-27 01-39-01

kmoppel:
Tested with some older pre 10 versions also both in streaming only and rsync mode. The only downside of going non-superuser - one cannot user explicit (--force) checkpointing for the `barman switch-wal` command.

gbartolini:
This was in my wishlist! Thanks, we will definitely look into this for 2.11!

PR #285, opened 2020-05-25
List backups of an inactive server

dbonne:
If we temporarily deactivate a server, we can now
see the list of its backups.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -126,6 +126,20 @@ class TestCli(object):
         assert err == ''
         assert not to_be_executed
 
+        # Request to include inactive server
+        to_be_executed = manage_server_command(server, skip_inactive=False)
+        out, err = capsys.readouterr()
+        # Expect no error whatsoever
+        assert err == ''
+        assert to_be_executed
+
+        # Request exclude include inactive server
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -126,6 +126,20 @@ def test_manage_server_command(self, monkeypatch, capsys):
...
         assert not to_be_executed
 
+        # Request to include inactive server
+        to_be_executed = manage_server_command(server, skip_inactive=False)
+        out, err = capsys.readouterr()
+        # Expect no error whatsoever
+        assert err == ''
+        assert to_be_executed
+
+        # Request exclude include inactive server
```
justinclift, line 136: Should this be "Request to exclude inactive server"?

mikewallace1979 (changes requested):
Thanks for making this PR @dbonne - I appreciate you made this some time ago so the need to get it merged has likely passed, however it seems like a generally worthwhile change.

In addition to the comments, there are going to be some conflicts rebasing this branch because the command names for `list-server` and `list-backup` have changed.

mikewallace1979:
Closing due to inactivity.

Issue #292, opened 2020-07-22
identify_compression is unable to identify custom compression


Thoro:
I was looking through the code for #291 and saw that identify_compression (https://github.com/2ndquadrant-it/barman/blob/1bcdd6bcbaed9bd0b5882bc9543036e967ff3368/barman/compression.py#L97) does not support the magic for the custom command.

By implementing this any algorithm could be used to actually compress on the postgres server side and send the compressed file via put-wal.

I was thinking about a "custom_compression_magic" that is provided as a hex key and is then set for the CustomCompressor.

```
custom_compression_magic = 0x425a68   # Bzip2
custom_compression_magic = 0x28b52ffd   # Zstd
custom_compression_magic = 0x377abcaf271c   # 7za
```

Thoro:
I added a PR for this, not totally happy, since identify_compression is currently a class method, and I didn't want to rewrite all uses of it.

mikewallace1979:
Thanks for making the PR @Thoro - I mentioned in a comment over on the PR but in the interests of visibility: The way barman currently "detects" custom compressed files is to try and match the magic for all supported compressions and default to the custom compression if no match is found (and a custom compression filter is configured). Using a `custom_compression_magic` config as suggested here would be an improvement on the current method.

mikewallace1979:
> Using a custom_compression_magic config as suggested here would be an improvement on the current method.
Specifically, it allows us to detect pre-compressed WALs which have been compressed with the custom compression and therefore avoid compressing them a second time during archive.

Issue #291, opened 2020-07-22
barman-wal-archive / put-wal using (custom) compression


thoro:
Hi,

I started using put-wal / barman-wal-archive and checkout out compression (https://www.endpointdev.com/blog/2017/03/28/postgres-wal-files-best-compression)

barman-wal-archive is currently not able to use any compression (via any parameter at least) and I thought for everyone using wal-archive via a internet connection this might actually be very helpful.

I envision something like in the config file, where you can put either a supported algorithm i.e. gzip or you can give the path to a custom script that executes the compression algorithm.

for example:

barman-wal-archive --compression=custom --custom_compression_filter=compress_wal.sh barman pg 

The compression option should also take existing compression options: bzip2, gzip, pybzip2, pygzip, pigz

martinmarques:
Just an update here. The next release of Barman will not have compressed archiving, yet! We are working on new compression algorithms on the Barman server and at restore time. Once that is ready, and we plan on releasing those in the next release, we'll work on giving the user options for compressing before shipping the WALs to the Barman server.

martinmarques:
This has been delivered in 3.14

PR #294, opened 2020-07-22
#292 Introduce custom_compression magic

Thoro:
see #292


mikewallace1979:
Hi @Thoro - thanks for taking the time to make this PR and update the tests.

I rebased your branch onto the current master [here](https://github.com/EnterpriseDB/barman/tree/dev/Thoro-magic_custom_compression) with a view to resolving the issue with the `identify_compression` method, however while I was there I noticed that `barman wal-archive` and `barman get-wal` are already able to handle custom compression albeit through a simpler method than the one you are proposing. Specifically, if a custom compression algorithm is in the configuration then it is used in the case that no recognised compression magic is found.

In the case of wal-archive, the compression is actually determined after the call to [WalFileInfo.from_file](https://github.com/EnterpriseDB/barman/pull/294/files#diff-65121e20a7f921ae67836deb2ebccaae8ae35211b28b1284730bf65c96e5e08fL507) when we call [CompressionManager.get_default_compressor](https://github.com/EnterpriseDB/barman/blob/33ce3d35fe72458ba8e36fae1e8b492109743a17/barman/wal_archiver.py#L141) which will return the configured compression.

In the code path for get-wal, the CompressionManager [get_wal_file_info](https://github.com/EnterpriseDB/barman/blob/33ce3d35fe72458ba8e36fae1e8b492109743a17/barman/server.py#L1858) function is called which will result in the custom compressor being used [if the compression cannot be identified](https://github.com/EnterpriseDB/barman/blob/33ce3d35fe72458ba8e36fae1e8b492109743a17/barman/infofile.py#L348-L351).

So although allowing the custom compression magic to be specified is a worthy goal it is not necessary for the existing barman functions that support compression (though as you note in #291 there are other functions which do not currently allow compression to be specified).

That being the case I don't think we want to merge this work in its current form. If we could come up with an implementation which avoided having multiple `identify_compression` methods or multiple `from_file*` methods then we could reconsider however I suspect that requires more rework of the existing barman code.

Thoro:
@mikewallace1979 

> I rebased your branch onto the current master [here](https://github.com/EnterpriseDB/barman/tree/dev/Thoro-magic_custom_compression) with a view to resolving the issue with the `identify_compression` method, however while I was there I noticed that `barman wal-archive` and `barman get-wal` are already able to handle custom compression albeit through a simpler method than the one you are proposing. Specifically, if a custom compression algorithm is in the configuration then it is used in the case that no recognised compression magic is found.

This is correct, but does not allow the actual use-case I was proposing, the proposed use case was to pre-compress the wal files, so that barman is not compressing them a second time. (although I barely remember doing that PR) Since barman does not recognize the custom magic, it will just happily compress the files a **second** time.

mikewallace1979:
@Thoro thanks for clarifying - I see it now, down on [L349](https://github.com/EnterpriseDB/barman/blob/33ce3d35fe72458ba8e36fae1e8b492109743a17/barman/wal_archiver.py#L349-L350). Currently `wal_info.compression` is going to have `None`, even if the file has already been compressed with the custom compression algorithm. With this patch the custom compression will have been detected and therefore that second compression will be avoided.

mikewallace1979:
Merged in #415 - thanks for the contribution @Thoro

Thoro:
Wow thanks, didn‘t expect that to still happen 👍

have to upgrade to master now 😁

mikewallace1979:
No worries and thanks again - this is going out in the 2.16 release which, if things go to plan, should be out by the end of the week.

Issue #333, opened 2021-04-13
barman backups failing due to idle in transaction termination


martinmarques:
When `idle_in_transaction_session_timeout` is set to a low value, you can get failed backups because barman leaves connections opened with a live transaction.

Internally reported in RT70653

PR #345, opened 2021-07-12
Enable autocommit everywhere

mikewallace1979:
This PR proposes a less problematic fix for #333 which does not cause breaking behavior with psycopg2 >=2.9.

The original fix merged in #340 is reverted and `autocommit = True` is enabled for all PostgreSQL connections.


amenonsen (approved):

amenonsen:
Merged, thanks.

PR #329, opened 2021-03-14
Fix missing WAL information in xlog.db

albatross0:
closes #328

PR #352, opened 2021-07-19
Hold the xlogdb lock while replacing the contents of xlog.db

### LINKED ISSUE #328: WAL information may not be added to the xlog.db file
barman v2.11 (RPM-installed on CentOS 8)

The following lines are part of my xlog.db file.

```
000000020002E59C000000BC        3089819 1615159937.4639714      pybzip2
000000020002E59C000000BD        3071435 1615159938.1930323      pybzip2
000000020002E59C000000BE        3092476 1615159938.944095       pybzip2
000000020002E59C000000C0        3136510 1615159940.656238       pybzip2
000000020002E59C000000C1        3286609 1615159944.1185274      pybzip2
000000020002E59C000000C2        3201442 1615159944.4345536      pybzip2
```

The line of 000000020002E59C000000BF is missing, but actually the archived WAL file exists.

```
# cd /var/lib/barman/dbname/wals/000000020002E59C/
# ls -l 000000020002E59C000000BF
-rw-------. 1 barman barman 3148731  3月  8 08:32 000000020002E59C000000BF
```

amenonsen:
BackupManager.rebuild_xlogdb() would open the server.xlogdb() as fxlogdb
and then loop over the files on disk, writing an entry for each one into
fxlogdb_new. At the end, it would replace xlog.db with xlog.db.new using
shutil.move.

The problem is that it replaced the old xlog.db after the `with …` block
that held ServerXLOGDBLock, which led to the problem reported in #328:

    The archive-wal process (PID:4011354) was running. (forked from
    barman cron)

    The backup process (PID:4015392) started deleting WAL segments.
    (forked from barman backup)

    This process opened xlog.db (inode: 1613568054) and xlog.db.new
    (inode: 7058292408)

    Then the archive-wal process (PID:4011354) opened xlog.db (inode:
    1613568054) and wrote the line of 000000020002E59C000000BF to (old)
    xlog.db file.

Because we replaced xlog.db with xlog.db.new after releasing the lock,
the sequence of events above could lose valid new entries in xlog.db.

The reporter proposed to solve this problem in PR #329 by just calling
shutil.move() while holding the lock, but I was not comfortable mixing
this call with server.xlogdb()'s flush/fsync operations (which would be
on the original fxlogdb fd).

Although there shouldn't be a problem in practice, because we were not
writing to fxlogdb in the loop, I chose a slightly different fix:

1. Convert fxlogdb_new into an unnamed TemporaryFile in the same
   directory (which we don't need to worry about removing, nor
   about fsync'ing the parent directory afterwards)

2. Accumulate new entries in fxlogdb_new as before (we don't want to do
   this in memory because xlog.db could be large)

3. At the end of the loop, copy the contents of fxlogdb_new into fxlogdb
   directly using shutil.copyfileobj instead of shutil.move; this is the
   same function that .move would use internally, but after reopening
   both source and destination files.

With this arrangement of the code, it doesn't matter that server.xlogdb
would flush/fsync the "old" fd when exiting the context, because that's
the file we will continue to use anyway.

With thanks to KUWAZAWA Takuya for the report and original analysis.

Closes #328

(This is a proposed replacement for PR #329)

### DIFF
--- a/barman/backup.py
+++ b/barman/backup.py
@@ -24,6 +24,7 @@ import datetime
 import logging
 import os
 import shutil
+import tempfile
 from contextlib import closing
 from glob import glob
 
@@ -871,8 +872,8 @@ class BackupManager(RemoteStatusMixin):
         wal_count = label_count = history_count = 0
         # lock the xlogdb as we are about replacing it completely
         with self.server.xlogdb("w") as fxlogdb:
-            xlogdb_new = fxlogdb.name + ".new"
-            with open(xlogdb_new, "w") as fxlogdb_new:
+            xlogdb_dir = os.path.dirname(fxlogdb.name)
+            with tempfile.TemporaryFile(mode="w+", dir=xlogdb_dir) as fxlogdb_new:
                 for name in sorted(os.listdir(root)):
                     # ignore the xlogdb and its lockfile
                     if name.startswith(self.server.XLOG_DB):
@@ -921,9 +922,11 @@ class BackupManager(RemoteStatusMixin):
                                 "unexpected file " "rebuilding the wal database: %s",
                                 fullname,
                             )
-                os.fsync(fxlogdb_new.fileno())
-        shutil.move(xlogdb_new, fxlogdb.name)
-        fsync_dir(os.path.dirname(fxlogdb.name))
+                fxlogdb_new.flush()
+                fxlogdb_new.seek(0)
+                fxlogdb.seek(0)
+                shutil.copyfileobj(fxlogdb_new, fxlogdb)
+                fxlogdb.truncate()
         output.info(
             "Done rebuilding xlogdb for server %s "
             "(history: %s, backup_labels: %s, wal_file: %s)",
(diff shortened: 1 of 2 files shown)


mikewallace1979 (approved):

mnencia:
I think you should add `os.fsync(fxlogdb.fileno())` at the end of the copy procedure, otherwise you could get into weird corner cases.

amenonsen:
> I think you should add `os.fsync(fxlogdb.fileno())` at the end of the copy procedure, otherwise you could get into weird corner cases.

I was counting on server.xlogdb()'s `finally:` to do it.

gcalacoci:
> > I think you should add `os.fsync(fxlogdb.fileno())` at the end of the copy procedure, otherwise you could get into weird corner cases.
> 
> I was counting on server.xlogdb()'s `finally:` to do it.

I honestly cannot think of a way for "escape" the `finally`, as it should be executed also in case of exceptions and abrupt quits... so makes sense to have it there...

mnencia:
you are right, it is already handled.

amenonsen:
Merged, thanks to everyone for the comments.

PR #354, opened 2021-07-20
Add -d option to specify database name for barman-cloud-backup.

jthreefoot-edb:


mikewallace1979 (approved):
:+1:

I was wondering about validating the conninfo but that happens later on when `postgres.PostgreSQL` attempts to parse it, so I think we're fine here.

amenonsen:
Merged. Thank you @jthreefoot-edb !

PR #358, opened 2021-07-22
Encrypt backup contents on s3

### LINKED ISSUE #350: Have barman-cloud encrypt the backup content in S3
Currently when the `--encryption` is used in barman-cloud it is only used to encrypt the `backup.info` file and the archived WALs. The actual backup content is not encrypted.

We need to pass in the specified encryption value as the `ServerSideEncryption` argument in the call to [client.create_multipart_upload](https://github.com/EnterpriseDB/barman/blob/f99ab593f290151403cfdbdb623d3f123dc4e425/barman/cloud.py#L903-L905).

While we're making this change we should also remove the `--encryption` argument from cloud_backup_list, cloud_restore and cloud_walrestore, as it is not used by barman-cloud for those actions.

mikewallace1979:
Updates the way we handle the `--encryption` option in barman-cloud so that it is passed to boto3 when creating a multipart upload. This means that S3 will now apply the selected encryption to the backup content in addition to the backup.info file and the WALs which were already being encrypted.

We also clean up the documentation of the option and remove it from commands that do not use it (barman-cloud-backup-list, barman-cloud-restore and barman-cloud-wal-restore).

Note: This is based on branch 347-support-encryption-scopes-in-azure which is the subject of https://github.com/EnterpriseDB/barman/pull/357

Closes #350


amenonsen:
Merged.

PR #364, opened 2021-07-23
Don't try to store a top-level PGDATA symlink in data.tar

### LINKED ISSUE #351: barman-cloud-restore "seeking backwards is not allowed"
Hello,
when running `barman-cloud-restore` I get the following error:

`ERROR: Barman cloud restore exception: seeking backwards is not allowed`

I researched online this error and found that it is actually a python error in the tarfile module. There is [a bug report detailing this error](https://bugs.python.org/issue12800) and also a "test.py" to verify the bug. 

Using the provided "test.py" I verified that I had a faulty python version. I have upgraded to Ubuntu 20.04 with Python 3.8.10. Now running the "test.py" doesn't get me the error. But barman-cloud-restore keeps returning the same error.

The archive command we are using is `archive_command = 'barman-cloud-wal-archive -P barman-cloud -z s3://bucket-name/ server-name %p'`

I tried manually running a restore, without the help of barman-cloud-restore, and found that the backups are correct.

Has any one using `barman-cloud-restore` experienced this same issue?

Thanks
Francesco

amenonsen:
It's perfectly valid to have PGDATA, say /var/lib/postgresql/13/main, be
a symlink to /mnt/somedir. But there were two significant problems with
how barman-cloud-backup handled this case.

1. If you run `barman-cloud-restore … /some/dir`, you would expect that
   directory to BE your data directory, not a symlink to somewhere else.
   Even if you were OK with the symlink, there would be no way to allow
   you to specify a different directory, the way it is now possible to
   do for tablespaces.

2. As reported in #351, there are Python-version-dependent problems with
   extracting from a tarfile that contains symlinks.

   Using Python 3.8.10 or newer may fix the problem or some part of it,
   but barman-cloud-restore can still trigger the "seeking backwards is
   not allowed" error under 3.8.10 with a backup from a server where
   PGDATA is a symlink.

The existing behaviour being unreasonable and problematic—and clearly
not anticipated by the code—we change _backup_copy to check if PGDATA
is a symlink, and if so, just backup whatever it points to.

Closes #351

### DIFF
--- a/barman/cloud.py
+++ b/barman/cloud.py
@@ -1128,7 +1128,11 @@ class CloudBackupUploader(with_metaclass(ABCMeta)):
                     include=["/PG_%s_*" % server_major_version],
                 )
 
-        # Copy PGDATA directory
+        # Copy PGDATA directory (or if that is itself a symlink, just follow it
+        # and copy whatever it points to; we won't store the symlink in the tar
+        # file)
+        if os.path.islink(pgdata_dir):
+            pgdata_dir = os.path.realpath(pgdata_dir)
         controller.upload_directory(
             label="pgdata",
             src=pgdata_dir,


mikewallace1979 (approved):

amenonsen:
Merged.

Issue #325, opened 2021-03-03
barman-cloud-wal-restore Sometimes Unable to Restore Compressed Files


bonesmoses:
We have encountered an issue where restoring from an Amazon S3 location can result in this Boto3 error:

```
2021-03-02 14:53:35,174 [36298] ERROR: Barman cloud WAL restore exception: 'StreamingBody' object has no attribute 'tell'
```

And this does to be the case. In the Boto3 code, the `StreamingBody` object does not provide a `tell` method. Even though it is a "file-like" object, passing this to a method that expects certain callable routines will result in an error. We apparently do this in `barman/cloud.py`:

```python
        # Write the dest file in binary mode
        with open(dest_path, 'wb') as dest_file:
            # If the file is not compressed, just copy its content
            if not decompress:
                shutil.copyfileobj(remote_file, dest_file)
                return

            if decompress == 'gzip':
                source_file = gzip.GzipFile(fileobj=remote_file, mode='rb')
            elif decompress == 'bzip2':
                source_file = bz2.BZ2File(remote_file, 'rb')
            else:
                raise ValueError("Unknown compression type: %s" % decompress)

            with source_file:
                shutil.copyfileobj(source_file, dest_file)
```

As the `StreamingBody` object only offers a `read` method, the "correct" use would be to avoid `shutil.copyfileobj` and write the file directly. I.e.:

```python
dest_file.write(source_file.read())

# Or this

for chunk in source_file.read(1048576): # or some other configurable buffer size
    dest_file.write(chunk)
```

This appears to only happen when the base class or various superclasses (`gzip.GzipFile`) when streamed from S3 lack the `tell` method, but this can happen with other methods as well, such as `seek`. This may also be boto3 version dependent, as there's currently a bug discussion in [issue 879](https://github.com/boto/botocore/issues/879) regarding adding `IOBase` to add these missing methods. The issue is still open, but some pull requests have been accepted through its lifetime.

Either we should require a certain Boto3 version, wrap the compressed IO classes to include `IOBase` ourselves or add the methods `shutil.copyfileobj` expects, or write files directly.

I'll also note that in the same `barman/cloud.py` we create a `StreamingBodyIO`  class that extends `RawIOBase`, but this is not utilized for compressed streams.

bonesmoses:
[...]

PR #369, opened 2021-07-27
Document compression issues with barman-cloud-wal-archive and older python versions

mikewallace1979:
Adds warnings around using `--gzip` and `--bzip2` on barman-cloud-wal-archive for older python versions due to issue #325.

I stopped short of automatically refusing to run with those options on affected python versions because it's possible someone might want to archive WALs from an old machine running python 2 and restore onto a newer machine running a current python 3 - such a scenario would not be affected by issue #325 at all and would work correctly.

### DIFF
--- a/barman/clients/cloud_walarchive.py
+++ b/barman/clients/cloud_walarchive.py
@@ -155,7 +155,8 @@ def parse_arguments(args=None):
     compression.add_argument(
         "-z",
         "--gzip",
-        help="gzip-compress the WAL while uploading to the cloud",
+        help="gzip-compress the WAL while uploading to the cloud "
+        "(should not be used with python < 3.2)",
         action="store_const",
         const="gzip",
         dest="compression",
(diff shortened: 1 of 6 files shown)


amenonsen:
I had initially thought that we should warn people who try to *restore* compressed WAL with python 2.7 and tell them to use 3.x instead (for the restore). There's some sense to warning them against getting into the situation to begin with… but I'll think about this a little more.

mikewallace1979:
> I had initially thought that we should warn people who try to _restore_ compressed WAL with python 2.7 and tell them to use 3.x instead (for the restore). There's some sense to warning them against getting into the situation to begin with… but I'll think about this a little more.

Warning on restore makes a lot of sense - we could check for python 2 once we have determined whether compression is involved, e.g. [after finding the WAL file](https://github.com/EnterpriseDB/barman/blob/master/barman/clients/cloud_walrestore.py#L194-L227) and exit with a "use Python 3.x" message there.

mikewallace1979:
@amenonsen Added a commit to error cleanly if restoring compressed WALs on python 2

PR #380, opened 2021-08-19
Optimize _create_dir_and_purge for empty destination

### LINKED ISSUE #378: Skip potentially expensive RSync on copy when target dir is empty
As reported in #377 barman will run an RSync command with a filter file containing two entries for each file in PGDATA as a preparatory step against the target directory for a restore (the same also applies to a backup). Even when running against an empty directory this command has been reported to take over an hour when there are hundreds of thousands of files in PGDATA.

When the target directory is empty during a copy we should always skip this command.

mikewallace1979:
Speeds up _create_dir_and_purge for empty destination directories
by ensuring exclude_and_protect_file is only populated if the
destination directory has content.

The motivation for this patch is: In cases where tablespaces
are not used the number of entries in exclude_and_protect_file
will be 2x the number of files in the PGDATA directory. As this
number enters the hundreds of thousands it can cause RSync to take
hours to execute the command created by _create_dir_and_purge
even if the destination directory is empty (as is the case when
a user is restoring to an empty directory, or barman is taking a
backup into an empty directory).

We can therefore speed up backup and restore times in these cases
by not having RSync apply the filter rules in
exclude_and_protect_file.

This is achieved by updating _analyze_directory so that it
only populates exclude_and_protect_file if the destination
directory has content. This is determined by setting
ref_has_content to True if _list_files returns *anything* which
isn't the current directory. This meant doing away with the dict
comprehension and replacing it with a for loop since the presence
of a directory counts as content. We therefore must check for
directories - if we filter them out in a comprehension then we
would have to do another pass over the output of _list_files in
order to determine whether there is any content.

Closes #378

### DIFF
--- a/barman/copy_controller.py
+++ b/barman/copy_controller.py
@@ -870,13 +870,13 @@ class RsyncCopyController(object):
         # Build a hash containing all files present on reference directory.
         # Directories are not included
         try:
-            ref_hash = dict(
-                (
-                    (item.path, item)
-                    for item in self._list_files(item, ref)
-                    if item.mode[0] != "d"
-                )
-            )
+            ref_hash = {}
+            ref_has_content = False
+            for file_item in self._list_files(item, ref):
+                if file_item.path != ".":
+                    ref_has_content = True
+                if file_item.mode[0] != "d":
+                    ref_hash[file_item.path] = file_item
         except (CommandFailedException, RsyncListFilesFailure) as e:
             # Here we set ref_hash to None, thus disable the code that marks as
             # "safe matching" those destination files with different time or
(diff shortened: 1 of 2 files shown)


mikewallace1979:
Timings for backup and recover with ~375000 files in PGDATA and no tablespaces:

| Operation | Patched? | Time   |
| --------- | -------- | ------ |
| backup    | no       | 16m34s |
| backup    | yes      | 3m7s   |
| recover   | no       | 22m32s |
| recover   | yes      | 3m38s  |

(Caveat: This was with a test dataset of mostly empty tables with an ad-hoc test environment - no science was involved and YMMV)

amenonsen (approved):

PR #386, opened 2021-09-03
Move pg_receieve{wal,xlog} after the connection check

mikewallace1979:
Moves the server version check for determining whether we log
pg_receivewal or pg_receivexlog so that it occurs after the initial
connection check.

This fixes the "Barman using replication slots server unreachable fails
running receive-wal" integration test which was breaking because the
call to self.server.streaming.server_version was raising an exception
directly and the expected log messages were not being logged.


mikewallace1979:
Closing because I didn't prefix the branch therefore the acceptance tests won't actually run. Sorry!

PR #389, opened 2021-09-03
Add support for Azure SAS token at blob container level

### LINKED ISSUE #388: Barman Cloud: Add support for Azure Storage Account SAS Token at Blob Container level
The current version of Barman Cloud doesn't support the [SAS Tokens generated at the Blob Container level](https://docs.microsoft.com/en-us/cli/azure/storage/container?view=azure-cli-latest#az_storage_container_generate_sas), it only supports [SAS Tokens generated at the Storage Account level](https://docs.microsoft.com/en-us/cli/azure/storage/account?view=azure-cli-latest#az_storage_account_generate_sas). This feature will allow a more granular set of credentials accessing the object storage.

mikewallace1979:
Replaces the implementation of AzureCloudInterface._check_bucket_existence
so that it does not use ContainerClient.exists. Instead we use
ContainerClient.list_blobs as a proxy for existence - if we can call
next() on the return value of list_blobs and trigger either no exception
or a StopIteration then we know the bucket exists.

This fixes the only issue which prevented the azure blob storage cloud
provider from working with SAS tokens created at the blob container
level, specifically that calling ContainerClient.exists requires storage
account permissions, not container permissions.

We also change the logic around creating the ContainerClient so that we
instantiate it directly rather than first creating a BlobServiceClient.
This isn't strictly required to fix the SAS token issue but since it
results in a simpler connection initialisation it is worth doing anyway.

Closes #388

### DIFF
--- a/barman/cloud_providers/azure_blob_storage.py
+++ b/barman/cloud_providers/azure_blob_storage.py
@@ -35,7 +35,7 @@ except ImportError:
 try:
     from azure.storage.blob import (
         BlobPrefix,
-        BlobServiceClient,
+        ContainerClient,
         PartialBatchErrorException,
     )
     from azure.core.exceptions import (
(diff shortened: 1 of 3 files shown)


mikewallace1979:
Will rebase and resolve conflicts...

amenonsen (approved):

Issue #373, opened 2021-08-05
Barman Monitoring REST API version 1 MVP


jthreefoot-edb:
In order for barman clients to programmatically monitor backups, we want to expose a REST API via a Flask app. 

For this initial prototype, I will be implementing the following endpoints
- /version: provides the current running version of barman
- /diagnose: provides the output of the barman --diagnose option

These will return json blobs.

The prototype will also have basic auth, and be set up so that OpenAPI specs can be generated from the code.

The purpose of this initial prototype is to make a minimal “vertical slice” of the REST app to determine how to implement the whole thing, and to have a sample interface clients can give feedback on.

PR #399, opened 2021-09-16
Dev/restapi diagnose

jthreefoot-edb:
For Issue #373


jthreefoot-edb:
Note that the json decoding stuff was autogenerated by openapi-generator. My actual code is utility_controller.py, run.py, and the OAS spec file.

PR #400, opened 2021-09-16
Add barman keep functionality

mikewallace1979:
Adds the ability to tag backups for retention beyond the specified retention policies.


tests/test_annotations.py
```diff
@@ -0,0 +1,506 @@
...
+    def _get_annotation_from_filesystem(self, base_dir, backup_id, key):
+        """
+        Helper which retrieves the value of an annotation from the filesystem.
+        """
+        with open(
+            "%s/%s/annotations/%s" % (base_dir, backup_id, key), "r"
+        ) as annotation:
+            return annotation.read()
+
+    def test_get_annotation_missing_backup(self, tmpdir):
```
didiermichel, line 62: I did not know about pytest tmpdir. it seems quite useful.

tests/test_cloud.py
```diff
@@ -1564,3 +1597,73 @@ def test_get_backup_succeeds_with_missing_main_file(self):
...
+        Verifies that KeepManagerMixinCloud methods are available in CloudBackupCatalog
+        and that they work as expected.
+
+        We deliberately do not test the functionality at a more granular level as
+        KeepManagerMixin has its own tests and CloudBackupCatalog adds no extra
+        functionality.
+        """
+        test_backup_id = "20210723T095432"
+
+        # Create a minimal in-memory CloudInterface implementation
```
didiermichel, line 1615: Maybe this minimal  implementation could be created in an external function to keep a better understanding of the test

tests/test_annotations.py
```diff
@@ -0,0 +1,506 @@
...
+    def test_delete_is_idempotent(self, tmpdir):
+        """Tests a single annotation can be deleted multiple times with the same result"""
+        base_backup_dir = tmpdir.mkdir("base")
+        os.makedirs("%s/%s/annotations" % (base_backup_dir, test_backup_id))
+        self._create_annotation_on_filesystem(
+            base_backup_dir, test_backup_id, "test_annotation", "annotation_value"
+        )
+        annotation_manager = AnnotationManagerFile(base_backup_dir)
+        annotation_manager.delete_annotation(test_backup_id, "test_annotation")
+        annotation_manager.delete_annotation(test_backup_id, "test_annotation")
```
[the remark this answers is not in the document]
mikewallace1979: I was happy that the not-deleting-other-files case was covered by the previous test - here we are only testing that delete_annotation will return successfully if it is run multiple times against the same annotation

tests/test_cloud.py
mikewallace1979: That's a good idea - I'll add a commit to do that

PR #402, opened 2021-09-20
set up unit tests workflow in github

didiermichel:
Add initial workflow to run unit tests with gitlab


amenonsen:
Running Black should be a separate workflow, and it should just create a PR and merge whatever changes it wants to make. We're never not going to want to apply the diff, so having to look at `--check --diff` reports will just be a waste of time.

PR #408, opened 2021-09-24
Avoid parsing backup label when backup_method = postgres

### LINKED ISSUE #206: Timezones of "Begin time" and "End time" are inconsistent
## The Problem

I ran into a small display issue, where the reported timezone of the timestamp of the "Begin time" is different than the timezone of the timestamp of the "End time" of a backup:

```
    Begin time           : 2019-04-01 21:21:05+02:00
    End time             : 2019-04-01 20:06:36.513013+00:00
```

I don't believe this has any serious consequences, but it would be nice to display this consistently.

## Expected behavior

I would expect both timestamps to use the same timezone.
It could be either the timezone of the server, or both could be normalized to UTC (probably best).

mikewallace1979:
Removes the call to `_backup_info_from_backup_label` in
`PostgresBackupStrategy` because all the fields read from the backup
label have already been set in `backup_info` during an earlier call to
`_backup_info_from_start_location` which retrieves them directly from
PostgreSQL.

The practical effect of this change is that it resolves a discrepancy
between the `begin_time` and `end_time` timezones in the backup info
because both `begin_time` and `end_time` are now read directly from
PostgreSQL and therefore have the same timezone info as the PostgreSQL
server.

The justification for the change is as follows:

1. When `PostgressBackupStrategy` was added in 1dd8ea8 the standard way
   to populate backup_info in Barman was to read from the backup label.
2. In b9233e4 the behaviour was changed such that backup_info fields
   were read directly from PostgreSQL but the subsequent call to read
   the values from the backup label was not removed.
3. In 8be0c15 the call to read the values from the backup label was
   removed for `ConcurrentBackupStrategy` but not for
   `PostgresBackupStrategy`.

By removing the call to `_backup_info_from_backup_label` we make it
consistent with the other backup strategies. The only remaining user of
`_backup_info_from_backup_label` is `ConcurrentBackupStrategy` when
running against PostgreSQL < 9.6 (i.e. it is using pgespresso). In these
cases the API to provide the backup_info fields from PostgreSQL is not
available and we must continue to read them from the backup label.

Closes #206

### DIFF
--- a/barman/backup_executor.py
+++ b/barman/backup_executor.py
@@ -1568,9 +1568,6 @@ class PostgresBackupStrategy(BackupStrategy):
 
         :param barman.infofile.LocalBackupInfo backup_info: backup information
         """
-        self._read_backup_label(backup_info)
-        self._backup_info_from_backup_label(backup_info)
-
         # Set data in backup_info from current_xlog_info
         self.current_action = "stopping postgres backup_method"
         output.info("Finalising the backup.")
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -852,31 +852,15 @@ class TestPostgresBackupExecutor(object):
             begin_offset=28,
             copy_stats=dict(copy_time=100, total_time=105),
         )
-        timestamp = datetime.datetime(2015, 10, 26, 14, 38)
+        start_timestamp = datetime.datetime(2015, 10, 26, 14, 38)
+        start_wal = "000000010000000000000012"
         backup_manager.server.postgres.current_xlog_info = dict(
             location="0/12000090",
-            file_name="000000010000000000000012",
+            file_name=start_wal,
             file_offset=144,
-            timestamp=timestamp,
+            timestamp=start_timestamp,
         )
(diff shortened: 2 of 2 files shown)


mikewallace1979:
The acceptance tests failed because this breaks backing up a standby server which is in recovery mode.

The reason is because PostgresBackupStrategy cannot determine the WAL name and offset on the standby server when it is in recovery mode - the [current_xlog_info function returns nulls for those values](https://github.com/EnterpriseDB/barman/blob/33ce3d35fe72458ba8e36fae1e8b492109743a17/barman/postgres.py#L615-L632) which means that we [cannot determine the timeline from the WAL name](https://github.com/EnterpriseDB/barman/blob/33ce3d35fe72458ba8e36fae1e8b492109743a17/barman/backup_executor.py#L1448-L1451) and therefore we cannot [determine the end WAL file](https://github.com/EnterpriseDB/barman/blob/33ce3d35fe72458ba8e36fae1e8b492109743a17/barman/backup_executor.py#L1475-L1479).

This is a long way of saying that PostgresBackupStrategy still depends on being able to read the backup label when backing up a server in recovery mode.

PR #416, opened 2021-10-01
Convert list-backup, list-server, show-server to plural forms

### LINKED ISSUE #413: Support plural form of barman commands
E.g. `barman list-backups <server>` in addition to `barman list-backup <server>`.

richyen:
Now these functions will be called `list-backups`, `list-servers`, `show-servers`
Support singular form for backward compatibility

Closes #413


doc/barman.1.d/50-list-backup.md
```diff
@@ -1,4 +1,4 @@
-list-backup *SERVER_NAME*
```
mikewallace1979, line 1: I think it's worth renaming the file too so it matches the non-legacy command name.

(Edit: same applies to `50-list-server.md` and `50-show-server.md`)

richyen: Good idea -- I've added a commit 👍

PR #417, opened 2021-10-01
Dev/411 refactor xlog functions

didiermichel:


didiermichel:
I am not that happy with the merged function. It should probably need a larger refactoring, but since we want to get rid of xlogdl use, it does not seem to be worth the effort.

tests/test_server.py
```diff
@@ -210,32 +219,117 @@ def test_get_wal_full_path(self, tmpdir):
...
+                ],
+                [1, 2, 3, 4, 6],
+            ),
+        ],
+    )
+    def test_get_required_xlog_files(self, wal_info_files, expected_indices, tmpdir):
+        # Prepare input string
+        walstring = ""
+        for i in wal_info_files:
+            walstring += i.to_xlogdb_line()
```
mikewallace1979, line 254: Minor nitpick: I think using the name `wal_info` instead of `i` would be more expressive here.

tests/test_server.py
```diff
@@ -210,32 +219,117 @@ def test_get_wal_full_path(self, tmpdir):
...
+            global_conf={"barman_lock_directory": tmpdir.mkdir("lock").strpath},
+            main_conf={"wals_directory": wals_dir.strpath},
+        )
+
+        wals = []
+        for wal_file in server.get_required_xlog_files(backup, 2, 41):
+            # get the result of the xlogdb read
+            wals.append(wal_file.name)
+        # Check for the presence of expected files
+        assert expected_wals == wals
```
mikewallace1979, line 280: It took me a little while to figure out that we are verifying that get_required_xlog_files returns all WALs on the target timeline along with the history files - it could be helpful for future maintainers if the test explained this somewhere, though I'm not sure where best. Maybe in the parameterize function or the test docstring?

mikewallace1979 (approved):
Looks good overall. I left a couple of small comments but the increase in test coverage is great and it's a good example of using pytest.parametrize to increase the test surface area.

PR #435, opened 2021-10-21
Replace argh with argparse

### LINKED ISSUE #398: argh dependency is unmaintained
Hello,

The argh dependency for barman is unmaintained since 2016: https://github.com/neithere/argh/

Would you consider migrating to something else, such python's stdlib argparse?

stratakis:
This is an initial attempt to replace argh with the stdlib argparse library to resolve #398 

Should be compatible across all python's supported by barman (at least locally it was). The tests for the cli pass as well. However there are still some issues to resolve, like the help command showing properly the positional arguments.

Let me know what you think of that approach.

### DIFF
--- a/barman/cli.py
+++ b/barman/cli.py
@@ -24,11 +24,14 @@ import json
 import logging
 import os
 import sys
-from argparse import SUPPRESS, ArgumentTypeError
-from contextlib import closing
-from functools import wraps
+from argparse import SUPPRESS, ArgumentTypeError, ArgumentParser, HelpFormatter
 
-from argh import ArghParser, arg, expects_obj, named
+if sys.version_info.major < 3:
+    from argparse import Action, _SubParsersAction, _ActionsContainer
+import argcomplete
+from collections import OrderedDict
+from contextlib import closing
+
 
 import barman.config
 import barman.diagnose
@@ -52,6 +55,161 @@ from barman.utils import (
 _logger = logging.getLogger(__name__)
 
 
+# Support aliases for argparse in python2.
+# Derived from https://gist.github.com/sampsyo/471779 and based on the
+# initial patchset for CPython for supporting aliases in argparse.
+# Licensed under CC0 1.0
+if sys.version_info.major < 3:
+
+    class AliasedSubParsersAction(_SubParsersAction):
+        old_init = staticmethod(_ActionsContainer.__init__)
+
+        @staticmethod
+        def _containerInit(
+            self, description, prefix_chars, argument_default, conflict_handler
+        ):
+            AliasedSubParsersAction.old_init(
+                self, description, prefix_chars, argument_default, conflict_handler
+            )
+            self.register("action", "parsers", AliasedSubParsersAction)
+
+        class _AliasedPseudoAction(Action):
+            def __init__(self, name, aliases, help):
+                dest = name
+                if aliases:
+                    dest += " (%s)" % ",".join(aliases)
+                sup = super(AliasedSubParsersAction._AliasedPseudoAction, self)
+                sup.__init__(option_strings=[], dest=dest, help=help)
+
+        def add_parser(self, name, **kwargs):
+            aliases = kwargs.pop("aliases", [])
+            parser = super(AliasedSubParsersAction, self).add_parser(name, **kwargs)
+
+            # Make the aliases work.
+            for alias in aliases:
+                self._name_parser_map[alias] = parser
+            # Make the help text reflect them, first removing old help entry.
+            if "help" in kwargs:
+                help_text = kwargs.pop("help")
+                self._choices_actions.pop()
+                pseudo_action = self._AliasedPseudoAction(name, aliases, help_text)
+                self._choices_actions.append(pseudo_action)
+
+            return parser
+
+    # override argparse to register new subparser action by default
+    _ActionsContainer.__init__ = AliasedSubParsersAction._containerInit
+
+
+class OrderedHelpFormatter(HelpFormatter):
+    def _format_usage(self, usage, actions, groups, prefix):
+        for action in actions:
+            if not action.option_strings:
+                action.choices = OrderedDict(sorted(action.choices.items()))
+        return super(OrderedHelpFormatter, self)._format_usage(
+            usage, actions, groups, prefix
+        )
+
+
+p = ArgumentParser(
+    epilog="Barman by EnterpriseDB (www.enterprisedb.com)",
+    formatter_class=OrderedHelpFormatter,
+)
+p.add_argument(
+    "-v",
+    "--version",
+    action="version",
+    version="%s\n\nBarman by EnterpriseDB (www.enterprisedb.com)" % barman.__version__,
+)
+p.add_argument(
+    "-c",
+    "--config",
+    help="uses a configuration file "
+    "(defaults: %s)" % ", ".join(barman.config.Config.CONFIG_FILES),
+    default=SUPPRESS,
+)
+p.add_argument(
+    "--color",
+    "--colour",
+    help="Whether to use colors in the output",
+    choices=["never", "always", "auto"],
+    default="auto",
+)
+p.add_argument(
+    "--log-level",
+    help="Override the default log level",
+    choices=list(get_log_levels()),
+    default=SUPPRESS,
+)
+p.add_argument("-q", "--quiet", help="be quiet", action="store_true")
+p.add_argument("-d", "--debug", help="debug output", action="store_true")
+p.add_argument(
+    "-f",
+    "--format",
+    help="output format",
+    choices=output.AVAILABLE_WRITERS.keys(),
+    default=output.DEFAULT_WRITER,
+)
+
+subparsers = p.add_subparsers(dest="command")
+
+
+def argument(*name_or_flags, **kwargs):
+    """Convenience function to properly format arguments to pass to the
+    command decorator.
+    """
+
+    # Remove the completer keyword argument from the dictionary
+    completer = kwargs.pop("completer", None)
+    return (list(name_or_flags), completer, kwargs)
+
+
+def command(args=None, parent=subparsers, cmd_aliases=None):
+    """Decorator to define a new subcommand in a sanity-preserving way.
+    The function will be stored in the ``func`` variable when the parser
+    parses arguments so that it can be called directly like so::
+        args = cli.parse_args()
+        args.func(args)
+    Usage example::
+        @command([argument("-d", help="Enable debug mode", action="store_true")])
+        def command(args):
+            print(args)
+    Then on the command line::
+        $ python cli.py command -d
+    """
+
+    if args is None:
+        args = []
+    if cmd_aliases is None:
+        cmd_aliases = []
+
+    def decorator(func):
+        parser = parent.add_parser(
+            func.__name__.replace("_", "-"),
+            description=func.__doc__,
+            help=func.__doc__,
+            aliases=cmd_aliases,
+        )
+        parent._choices_actions = sorted(parent._choices_actions, key=lambda x: x.dest)
+        for arg in args:
+            if arg[1]:
+                parser.add_argument(*arg[0], **arg[2]).completer = arg[1]
+            else:
+                parser.add_argument(*arg[0], **arg[2])
+        parser.set_defaults(func=func)
+        return func
+
+    return decorator
+
+
+@command()
+def help(args=None):
+    """
+    show this help message and exit
+    """
+    p.print_help()
+
+
 def check_target_action(value):
     """
     Check the target action option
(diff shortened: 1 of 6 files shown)


stratakis:
Using aliases for python2 is non-trivial but it should be possible by utilizing something like https://gist.github.com/sampsyo/471779 which is based on the initial patchset for [CPython](https://bugs.python.org/issue9234)  adding aliases to argparse. I'll see if I can integrate it with the current approach.

stratakis:
> Thank you @stratakis for this PR - the overall approach looks reasonable and is definitely heading in the right direction.
> 
> There are a couple of blocking issues:
> 
>     * The barman commands which rely on `barman.__config__` being available are currently failing, e.g.:
> 
> 
> ```
> barman show-servers all
> EXCEPTION: 'NoneType' object has no attribute 'server_names'
> See log file for more details.
> 'NoneType' object has no attribute 'server_names'
> See log file for more details.
> Traceback (most recent call last):
>   File "/Users/michael.wallace/src/EnterpriseDB/barman/barman/cli.py", line 1495, in main
>     args.func(args)
>   File "/Users/michael.wallace/src/EnterpriseDB/barman/barman/cli.py", line 669, in show_servers
>     servers = get_server_list(args)
>   File "/Users/michael.wallace/src/EnterpriseDB/barman/barman/cli.py", line 1350, in get_server_list
>     available_servers = barman.__config__.server_names()
> AttributeError: 'NoneType' object has no attribute 'server_names'
> ```
> 

This should hopefully be fine after @jthreefoot-edb's recommendation but please verify if it looks ok.

>     * We would lose the singular forms of `show-servers`, `list-servers` and `show-backups` in python 2 - it's unfortunate argparse doesn't support the `aliases` keyword argument for python 2 but we need to keep those singular forms for backward compatibility.
> 

Working on it per https://github.com/EnterpriseDB/barman/pull/435#issuecomment-956858179

> 
> Also a couple of smaller issues:
> 
>     * There are still a few references to argh in the comments of the `pretty_args` down on [L1250](https://github.com/EnterpriseDB/barman/pull/435/files#diff-d91ba4117ac85c960aa4252ff9b8df565954676cf8148c1eb07cde33f2734ee4L1250).
> 

Done.

>     * argh added a `help` subcommand implicitly which does the same as what running `barman` on its own does with these changes - it would be nice if we could still have `barman help` just to keep the CLI consistent with previous versions.
>

I'll check that out as well.
 
> 
> If we can get those two blockers resolved then we should be able to get this merged and say goodbye to argh.

stratakis:
Aliases should now work for python2.

barman/cli.py
```diff
@@ -52,6 +54,116 @@
 _logger = logging.getLogger(__name__)
 
 
+# Support aliases for argparse in python2
+if sys.version_info.major < 3:
+    class AliasedSubParsersAction(_SubParsersAction):
```
mikewallace1979, line 59: Nice hack and thank you for putting the python2 workaround in.

This looks like it was mostly derived from https://gist.github.com/sampsyo/471779 - if so it's worth referencing that gist in a comment and including a note that the code is public domain. Although the CC0 1.0 license doesn't require attribution I think it's still important to note where the code came from just in case there are some issues with automated code scans in the future.

mikewallace1979:
Another thing I just noticed testing locally is that the `diagnose` command is gone - it didn't have any argh decorators but it's the `cli.diagnose` function. It doesn't take any arguments so probably just needs a `@command()` decorator.

stratakis:
A help command was added. Since I've found no easy way unfortunately to add an alias to --help I've simply created a new subparser which prints out the help. However this does not work the same as argh:

`$ barman help` will print out the help but only if we are under the barman user, since the global_config is called

argh gave the possibility to also do something like `$ barman help list-servers` which would be equivalent to `$ barman list-servers --help`. That is not the case with the new subparser adding the `help` command. Possibly I'll have to delve again into the private attributes or argparse, however the complexity seems quite big if the argh compatibility is to be maintained 100%.

mikewallace1979:
> Possibly I'll have to delve again into the private attributes or argparse, however the complexity seems quite big if the argh compatibility is to be maintained 100%.

Agreed - this might be somewhere we need to break compatibility for the sake of keeping complexity under control. Given that the `help` command is nearly always going to be used interactively (I can't think of any use cases for running help in scripts) it doesn't seem too bad to change behaviour here - and the `--help` option is always there to provide subcommand-specific help when needed.

barman/cli.py
```diff
@@ -67,8 +186,11 @@ def check_target_action(value):
     raise ArgumentTypeError("'%s' is not a valid recovery target action" % value)
 
 
-@named("list-servers")
-@arg("--minimal", help="machine readable output")
+@command([
+        argument("--minimal",
+                 help="machine readable output")
+        ],
```
mikewallace1979, line 192: I think we need an `action="store_true"` here (there wasn't one in the original code, I suppose argh must have been adding that automatically).

barman/cli.py
```diff
@@ -67,8 +186,11 @@ def check_target_action(value):
...
+        cmd_aliases=["list-server"])
 def list_servers(minimal=False):
```
mikewallace1979, line 194: This should accept `args` and then down on L208 we can reference `args.minimal`. Currently its getting the args namespace object as the `minimal` argument which is truthy and therefore forcing the minimal output even if the option isn't set.

barman/cli.py
```diff
@@ -691,42 +731,28 @@ def show_servers(args):
...
+                  dest="archive",
+                  action="store_true",
+                  default=False),
+         argument("--archive-timeout",
+                  help="the time, in seconds, the archiver will wait for a new WAL file "
+                  "to be archived before timing out",
+                  metavar="TIMEOUT",
+                  default="30",
+                  type=check_non_negative)],
+         cmd_aliases=["switch_xlog"])
```
mikewallace1979, line 755: This should be `switch-xlog`

mikewallace1979 (commented):
A few extra comments following testing locally - this is looking pretty good though! We'll give it a final review tomorrow.

barman/cli.py
stratakis: Hope I understood it correctly, check the latest commit.

stratakis:
Some more caveats that I've noticed when comparing the new and old help output side by side:

Running `--help` now will show the dict of the positional arguments including the aliases and in a non-alphabetical order. argh would sort the arguments and would not show the aliases.

Also argh would show a list of all the subparsers as positional arguments on the help text along with their help. In the case of argparse this is for explicitly set arguments on the main parser, but not for the subparsers.

And I haven't looked yet on the shell completion with arghcomplete if things work there.

barman/cli.py
```diff
@@ -790,6 +804,7 @@ def check(args):
     output.close_and_exit()
 
 
+@command()
 def diagnose():
```
mikewallace1979, line 808: Just noticed this still needs a placeholder arg as argparse is going to give it a namespace.

barman/cli.py
mikewallace1979: Yep, looks good :+1:

stratakis:
By the way I'll be squashing the commits to one when everything is reviewed and ready.

barman/cli.py
```diff
@@ -52,6 +54,127 @@
...
+    Usage example::
+        @command([argument("-d", help="Enable debug mode", action="store_true")])
+        def command(args):
+            print(args)
+    Then on the command line::
+        $ python cli.py command -d
+    """
+    def decorator(func):
+        parser = parent.add_parser(func.__name__.replace('_','-'), description=func.__doc__,
+                                   help=func.__doc__, aliases=cmd_aliases)
```
didiermichel, line 161: I spent time on  add the help section yesterday but I realized you just add it :) Nice job.

didiermichel (commented):
Hi @stratakis, 
I had a look at the code. I like the decorator.

You should try run `tox -e black` to format your code before merging.

barman/cli.py
```diff
@@ -790,7 +808,8 @@ def check(args):
     output.close_and_exit()
 
 
-def diagnose():
+@command()
+def diagnose(args):
```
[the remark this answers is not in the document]
didiermichel: Also, maybe rebase on master to resolve PR conflict (`tests/requirements_minimal.txt`).
This will allow CI to run tests.

mikewallace1979:
Something like this would fix ordering of the subcommands in both the positional arguments and the expanded help, though it is yet more coupling to argparse internals: https://gist.github.com/mikewallace1979/8d9ae2fa84ce5f4fc0b70999fbd5b04c

barman/cli.py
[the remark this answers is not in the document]
jthreefoot-edb: > Hi, does we need args as it is not used(or maybe in decorator)?
> It also this breaks tests/test_diagnose.py where diagnose() is used.
> maybe set args=[]

It would need to be `args=None` actually. [https://docs.python-guide.org/writing/gotchas/](https://docs.python-guide.org/writing/gotchas/) 

I don't think this particular case would ever cause an actual problem bc diagnose doesn't have any arguments, but may as well set it to None.

stratakis:
> Something like this would fix ordering of the subcommands in both the positional arguments and the expanded help, though it is yet more coupling to argparse internals: https://gist.github.com/mikewallace1979/8d9ae2fa84ce5f4fc0b70999fbd5b04c

This is quite elegant, the solution I was trying was far more complex than that. Added it.

stratakis:
Also another thing I noticed is that the doctext shown for each command when running `$ barman --help` is not properly wrapped on the terminal as it happens with argh, or argparse by default for that matter. A newline is added after ~50 chars instead of extending the line up to where the terminal is.

mikewallace1979:
Giving this PR a final check now - we can get autocompletion working by adding `argcomplete.autocomplete(p)` just before the call to `p.parse_args` on L1786.

stratakis:
I'll squash the commits after the final review.

barman/cli.py
```diff
@@ -98,16 +254,13 @@ def list_servers(minimal=False):
...
-    help="Keep the stdout and the stderr streams attached to Barman subprocesses.",
+@command(
+    [
+        argument(
+            "--keep-descriptors",
+            help="Keep the stdout and the stderr streams attached to Barman subprocesses",
+        )
+    ]
 )
 def cron(keep_descriptors=False):
```
mikewallace1979, line 265: This is another place we were relying on argh magic which I missed in the last review - we need to do the same change we did for `list-servers`, specifically: accept `args` instead of `keep_descriptors`, add `action="store_true"` the `--keep-descriptors` argument, and refer to `args.keep_descriptors` inside the `cron` function.

mikewallace1979 (approved):

mikewallace1979:
@stratakis Thanks for making all the changes. I'm going to push a copy of this branch directly to the EnterpriseDB repo so that the full CI pipeline can run (there are some behind-the-scenes integration which will take around 20 minutes) and if that all succeeds then I think we're good to squash and merge.

barman/cli.py
```diff
@@ -52,6 +56,156 @@
...
+    """Convenience function to properly format arguments to pass to the
+    command decorator.
+    """
+
+    # Remove the completer keyword argument from the dictionary
+    completer = kwargs.pop("completer", None)
+    return (list(name_or_flags), completer, kwargs)
+
+
+def command(args=[], parent=subparsers, cmd_aliases=[]):
```
mikewallace1979, line 168: We should avoid using mutables as defaults for `args` and `cmd_aliases` here, as we did with `args=None` earlier in the PR.

mikewallace1979 (commented):
One last round of feedback - sorry!

barman/cli.py
mikewallace1979: Yes - we'd also need to test for `None` and assign an empty list for both parameters at the top of the function.

mikewallace1979:
@stratakis Ok I think you're all good to squash this down to one commit ready for merging.

stratakis:
Squashed and rebased on top of master. Edited the commit message as well. Should I also add anything else, such as a changelog entry?

mikewallace1979:
> Squashed and rebased on top of master. Edited the commit message as well. Should I also add anything else, such as a changelog entry?

I think we've got everything we need for this PR. We'll update the NEWS at release time and we retired the ChangeLog since it was just duplicating the commit history.

Thanks again for taking this on and sticking with it (and indeed identifying the issue in the first place) - it's great that we've been able to wave goodbye to unmaintained argh so quickly! 🎉🎉🎉

stratakis:
Thanks as well for reviewing the PR, addressing all the issues with it and merging it in such a swift manner :)

Is there an ETA for the next release?

mikewallace1979:
The next release should be out by the end of this coming week - 19th November or earlier.

Issue #423, opened 2021-10-07
Make Config Loading Re-runnable


jthreefoot-edb:
Currently, Barman operations are self-contained, loading its configs each time something is run. As a result, the config-related code does things like clobber the simple int or string loaded from the config file with an object reference, and assumes that at such points in the code, it will only be dealing with the initial types/values and never the updated ones. When this code is rerun in the same process (as it will have to in order to support the backup REST API), these assumptions cause crashes.

Ultimately, it would be good to refactor the config loading to be cleaner overall, but for now a few minimal fixes are all that's needed to support the REST API.

Split from #373 
Needed for https://github.com/EnterpriseDB/pg-backup-api/issues/1

PR #437, opened 2021-10-26
Make barman diagnose rerunnable in the same session.

jthreefoot-edb:
Missing fix for #423


jthreefoot-edb:
I checked the SonarQube failures, and it's just complaining about duplicated code in the /build directory, so I think this is fine to go ahead and merge anyway.

PR #441, opened 2021-11-02
Provide commands to check WAL archive destination is usable

### LINKED ISSUE #432: Implement new barman and/or barman-cloud subcommands to test the archive-destination contents
We could add an option to `barman-cloud-wal-archive` to enable the following behavior:

1. when archiving the very first WAL file (000000010000000000000000), `barman-cloud-wal-archive` should fail if the target position contains any backup or WAL
2. when archiving the first file of a WAL segment (XXXXXXXXXXXXXXX0000000), `barman-cloud-wal-archive` should fail if the target contains any WAL in that segment

the rationale is to prevent mixing WALs from different instances.

mikewallace1979:
Adds commands to barman and barman cloud which check that a barman server
or cloud location is safe to use as an archive destination for a new
PostgreSQL server.

A location is considered safe if either:

1. There are no WAL files at all in the archive.
2. Any existing WAL files meet one of the following criteria:
  a. They belong to an older timeline than that specified by the
     current_timeline argument.
  b. They are on the same timeline as the current_timeline argument but
     the segment is less than that specified in the current_wal_segment
     argument.

A file is considered a WAL file if it passes the `is_any_xlog_file` check
in `barman/xlog.py` so this applies to WAL files, history files, partial
WAL files and backup labels.

The commands added are:

* barman check-wal-archive
* barman-cloud-check-wal-archive

The motivation for this patch is to provide a way that external
orchestration tools can validate the WAL archive destination is safe for
a newly provisioned PostgreSQL cluster, given such a cluster may use the
exact same name as an old cluster.

In such scenarios, any WAL files which have a higher timeline or segment
than the WALs being written by the new cluster will cause any attempt to
restore from a backup to fail.

Reasons why external orchestration tooling may re-use the same cluster
name and archive destination include (but are not limited to):

* A new cluster is created via initdb with the same name as the old one.
  The sysid will be different but this does not affect the archive
  destination so any archived WALs relating to the older cluster will be
  present in the same location.
* A cluster is restored from a base backup and uses the same name as the
  old cluster. The cluster has the same sysid and starts with a segment
  ID > 1 and timeline > 1. The same archive destination used by the old
  cluster will be used for the restored cluster.
* A new cluster is started which happens to re-use the same name and
  archive destination.

All of these cases lead to the situation where WAL archiving and backup
is functioning normally *but* any attempts to restore from those backups
will fail. This is dangerous for anyone relying on the databases managed
by external orchestration/automation.

The commands provided by this patch do not solve the problem alone
because neither Barman nor PostgreSQL have the necessary context. The
commands can, however, be added to external automation in order to
catch archive safety issues at the provisioning stage.

Closes #432

### DIFF
--- a/barman/cli.py
+++ b/barman/cli.py
@@ -35,8 +35,13 @@ import barman.diagnose
 from barman import output
 from barman.annotations import KeepManager
 from barman.config import RecoveryOptions
-from barman.exceptions import BadXlogSegmentName, RecoveryException, SyncError
-from barman.infofile import BackupInfo
+from barman.exceptions import (
+    BadXlogSegmentName,
+    RecoveryException,
+    SyncError,
+    WalArchiveContentError,
+)
+from barman.infofile import BackupInfo, WalFileInfo
 from barman.server import Server
 from barman.utils import (
     BarmanEncoder,
@@ -48,6 +53,7 @@ from barman.utils import (
     get_log_levels,
     parse_log_level,
 )
+from barman.xlog import check_archive_usable
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


mikewallace1979:
Closing because sonarqube seems to think I've touched all the code.

PR #443, opened 2021-11-02
Provide commands to check WAL archive destination is usable

### LINKED ISSUE #432: Implement new barman and/or barman-cloud subcommands to test the archive-destination contents

mikewallace1979:
Adds commands to barman and barman cloud which check that a barman server
or cloud location is safe to use as an archive destination for a new
PostgreSQL server.

A location is considered safe if either:

1. There are no WAL files at all in the archive.
2. All existing WAL files belong to an older timeline than that
   specified by the --timeline argument.

A file is considered a WAL file if it passes the `is_any_xlog_file` check
in `barman/xlog.py` so this applies to WAL files, history files, partial
WAL files and backup labels.

The commands added are:

* barman check-wal-archive
* barman-cloud-check-wal-archive

The motivation for this patch is to provide a way that external
orchestration tools can validate the WAL archive destination is safe for
a newly provisioned PostgreSQL cluster, given such a cluster may use the
exact same name as an old cluster.

In such scenarios, any WAL files on the same or higher timeline as the
WALs being written by the new cluster will cause any attempt to restore
from a backup to fail.

Reasons why external orchestration tooling may re-use the same cluster
name and archive destination include (but are not limited to):

* A new cluster is created via initdb with the same name as the old one.
  The sysid will be different but this does not affect the archive
  destination so any archived WALs relating to the older cluster will be
  present in the same location.
* A cluster is restored from a base backup and uses the same name as the
  old cluster. The cluster has the same sysid and starts with a segment
  ID > 1 and timeline > 1. The same archive destination used by the old
  cluster will be used for the restored cluster.
* A new cluster is started which happens to re-use the same name and
  archive destination.

All of these cases lead to the situation where WAL archiving and backup
is functioning normally *but* any attempts to restore from those backups
will fail. This is dangerous for anyone relying on the databases managed
by external orchestration/automation.

The commands provided by this patch do not solve the problem alone
because neither Barman nor PostgreSQL have the necessary context. The
commands can, however, be added to external automation in order to
catch archive safety issues at the provisioning stage.

Closes #432


barman/clients/cloud_check_wal_archive.py
```diff
@@ -0,0 +1,143 @@
...
+
+import barman
+from barman.cloud import configure_logging, CloudBackupCatalog
+from barman.cloud_providers import get_cloud_interface
+from barman.exceptions import WalArchiveContentError
+from barman.utils import force_str, check_positive
+from barman.xlog import check_archive_usable
+
+try:
+    import argparse
```
jthreefoot-edb, line 29: Why does the argparse import need to be wrapped in a try-except?

jthreefoot-edb (approved):

barman/clients/cloud_check_wal_archive.py
mikewallace1979: This is a good question and I think the answer is "there is no longer any good reason".

The first commit to introduce this was 83e9979a70d8caae5a65ad1b195960cb01ec4cd5 where the `barman-wal-restore.py` script is added. This would have needed to support python 2.6 (and presumably older versions of 3.x) which didn't ship with argparse so there was a genuine possibility it could run without argparse being available, therefore providing a friendly error message was useful.

Now we only support python versions which include argparse I don't think there is a need to do this at all. I'll remove it from this PR and while we're at it I'll update the other scripts too.

ringerc:
After a bit more thinking I think it's probably sufficient to be able to say "no archives or timeline history files on timeline X or any greater timeline". Archives on the same timeline as end-of-recovery only matter if we're not doing a TLI increment, and in this use case I don't see why we wouldn't always be doing one.

So it's probably sufficient to say "it's an error if a timeline history file or any WAL seg with timeline (end-of-recovery-tli)+1 exists".

That's up to the tooling invoking this though. And being able to filter by LSN on specified timeline is still useful. You can just specify 0/0 if you don't care. We should default to that, so that specifying just a timeline makes it an error if any history files or WAL segs on or after that timeline exist.

Reading code now

mikewallace1979:
@ringerc So we could maybe simplify this to just passing in a `--current-timeline` and not require a WAL segment at all?

mikewallace1979:
(just rebased and updated now that the argh->argparse work has landed - everything else remains the same in the PR for now but I'll squash what needs to be squashed before merging)

mikewallace1979:
@ringerc I've just pushed a commit which removes `--current-wal-segment` and changes the behaviour so if `--timeline` is provided then it passes only if all WAL files in the archive relate to an earlier timeline.

@jthreefoot-edb I know you've already reviewed this but could you please take another look now the options have changed?

jthreefoot-edb:
> @jthreefoot-edb I know you've already reviewed this but could you please take another look now the options have changed?

Looks good to me.

PR #445, opened 2021-11-05
Dev/documentation update

didiermichel:
Those are small fixes to appease sonarqube warnings
Also update documentation


didiermichel:
Maybe it is worth checking some Code Smells as well.

mikewallace1979:
> Maybe it is worth checking some Code Smells as well.

I've started going through the sonarqube smells and have been picking off the low hanging fruit. I have not been addressing the "cognitive complexity" smells because there is more potential for introducing new bugs, but I'm working my way through the simple fixes on https://github.com/EnterpriseDB/barman/tree/dev/sonarqube-smells

Patches very welcome on that branch...!

PR #452, opened 2021-11-17
Dev/release 2.16

jthreefoot-edb:
NEWS and version number updates for release 2.16


jthreefoot-edb:
Closing to re-regenerate the docs with the right version of pandoc, and because I missed a merge.

PR #455, opened 2021-11-18
Remove the changelog file generation from scripts/release.sh

jthreefoot-edb:


didiermichel:
This change suggests that `scripts/gitlog-to-changelog` will not be needed either.

mikewallace1979 (approved):
Nice :+1:

We could also remove `scripts/gitlog-to-changelog` since we don't have a use for that anymore.

Issue #458, opened 2021-11-24
 barman.cli ERROR: local variable 'line' referenced before assignment


sebasmannem:
I ran into this error while running a restore.
Command:
   `barman -c ~/barman-restore.conf recover --get-wal --remote-ssh-command 'ssh postgres@server' server 20211122T010151 /postgres/12/recover`
Error in the logfile:

> 2021-11-24 12:13:10,614 [1418729] barman.cli ERROR: local variable 'line' referenced before assignment
> See log file for more details.
> Traceback (most recent call last):
>   File "/usr/lib/python3.6/site-packages/barman/cli.py", line 1640, in main
>     p.dispatch(pre_call=global_config)
>   File "/usr/lib/python3.6/site-packages/argh/helpers.py", line 55, in dispatch
>     return dispatch(self, *args, **kwargs)
>   File "/usr/lib/python3.6/site-packages/argh/dispatching.py", line 174, in dispatch
>     for line in lines:
>   File "/usr/lib/python3.6/site-packages/argh/dispatching.py", line 277, in _execute_command
>     for line in result:
>   File "/usr/lib/python3.6/site-packages/argh/dispatching.py", line 231, in _call
>     result = function(namespace_obj)
>   File "/usr/lib/python3.6/site-packages/barman/cli.py", line 647, in recover
>     standby_mode=getattr(args, "standby_mode", None),
>   File "/usr/lib/python3.6/site-packages/barman/server.py", line 1679, in recover
>     backup_info, dest, tablespaces, remote_command, **kwargs
>   File "/usr/lib/python3.6/site-packages/barman/backup.py", line 754, in recover
>     **kwargs
>   File "/usr/lib/python3.6/site-packages/barman/recovery_executor.py", line 342, in recover
>     self._analyse_temporary_config_files(recovery_info)
>   File "/usr/lib/python3.6/site-packages/barman/recovery_executor.py", line 1191, in _analyse_temporary_config_files
>     conf_file, self.MANGLE_OPTIONS, "%s.origin" % conf_file, append_lines
>   File "/usr/lib/python3.6/site-packages/barman/recovery_executor.py", line 1281, in _pg_config_mangle
>     if line[-1] != "\n".encode("utf-8"):
> UnboundLocalError: local variable 'line' referenced before assignment

[...]

PR #459, opened 2021-11-24
#458 fix _pg_config_mangle empty file corner case.

didiermichel:


barman/recovery_executor.py
```diff
@@ -1183,6 +1183,7 @@ def _analyse_temporary_config_files(self, recovery_info):
 
             append_lines = None
             if conf_file.endswith("postgresql.auto.conf"):
+                print("auto file")
```
amenonsen, line 1186: Debugging leftover?

amenonsen (approved):

barman/recovery_executor.py
didiermichel: Indeed I will remove it

PR #469, opened 2021-12-01
Limit sonarqube and blackduck jobs to EDB repo only

mikewallace1979:
Adds a conditional to both the sonarqube-scan and blackduck-scan jobs so
that they only run if the repository is `EnterpriseDB/barman`. The jobs
will fail when running on PRs from other repositories because they do
not have access to the necessary credentials.

This prevents us from automatically failing these two checks on all PRs
from forks.


mikewallace1979:
This was to test blackduck and sonarqube would be skipped when the PR is from a fork.

Issue #474, opened 2021-12-09
Add backup validation mechanism


didiermichel:
# Description
Since PG13 a `backup_manifest` file is created upon `pg_basebackup` call.
This manifest can then be used with `pg_verifybackup` to validate backup integrity.

In order to have a better monitoring on backups, we want to create a new barman command :`barman verify servername backupid`.
This command will call in background `pg_verifybackup backup_path -n` (`n` option  to ignore wall files as we do not save those the same way.)

In case of remote backup using rsync (could excluded),  `pg_basebackup` is not called so there is no `backup_manifest`  file.

# Technical steps:
## 1 Implement verify command
Command: `barman verify servername backupid`

Should be available for PG>= 13. Logs and returns a message.
If manifest missing for any reason (ex: old backup), returns logs and returns a message.
Should call in background: `pg_verifybackup backup_path -n`
Should return pg_verifybackup response as text.
ex:
```bash
# Extra file
pg_verifybackup: error: "backup_manifest_1" is present in the manifest but not on disk
# Inconsistent manifest.
pg_verifybackup: fatal: manifest checksum mismatch
# Something wrong in manifest file
pg_verifybackup: fatal: could not parse backup manifest: unrecognized top-level field
# File mismatch
pg_verifybackup: error: checksum mismatch for file "PG_VERSION"
# Sucess!
backup successfully verified
```

## 2 Implement create manifest command
This can be split into several steps.
### 2.1 Implement `backup_manifest`  file creation from backup directory
Implementation should match original backup_manifest file as closely as possible (some documentation here: https://www.postgresql.org/docs/13/backup-manifest-toplevel.html) to be usable with `pg_verifybackup`.

### 2.2 Implement create manifest command
Command: `barman generate-manifest servername backupid`

Should create a manifest for a backup.
Should abort and return a message if manifest already exists, PG version under 13.
Should use previously implemented manifest creation material.

### 2.3 Allow use create manifest as hookscript 
As we do not want to add additional time to actual backup behavior, manifest creation will be optional. 
activation should be done using post-backup hook in barman configuration to run `barman generate-manifest servername backup_id`

Sources:
* https://www.2ndquadrant.com/en/blog/backup-manifests-and-pg_verifybackup-in-postgresql-13/
* backup_manifest documentation: https://www.postgresql.org/docs/13/backup-manifest-toplevel.html

PR #475, opened 2021-12-13
 #474 implement backup_manifest creation

didiermichel:


barman/storage/file_stats.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, size, last_modified):
+        """
+        Arbitrary timezone set to UTC. There is probably possible improvement here.
+        :param size: file size in bytes
+        :type size: int
+        :param last_modified: Time of last modification in seconds
+        :type last_modified: int
+        """
+        self.size = size
+        self.last_modified = datetime.fromtimestamp(last_modified, tz=utc)
```
mikewallace1979, line 42: My immediate thought here was "we should use whatever time zone pg_basebackup uses which will surely be the time zone of the PostgreSQL server" however a few tests with PG14 showed that the backup manifest produced by pg_basebackup would use GMT regardless of the PostgreSQL time zone or the OS time zone.

The PG docs don't specify the time zone but [this mailing list discussion](https://postgrespro.com/list/id/4cea0d44-7524-dc7d-9043-8c701ccaba0a@oss.nttdata.com) suggests Last-Modified is always GMT.

didiermichel: I indeed noticed that GMT is used in backup manifest. I tested pg_verifybackup with random srting, it is not checked and just informational.
For consistency, I should probably add GMT to formated date.

doc/manual/42-server-commands.en.md
```diff
@@ -53,6 +53,23 @@ barman check <server_name>
 > and monitoring infrastructure. The `--nagios` option allows you
 > to easily create a plugin for Nagios/Icinga.
 
+## `generate-manifest`
+
+This command is useful when backup is created remotely and bg_basebackup is not 
+involved and `backup_manifest` file does not exist in backup.
+It will generate `backup_manifest` from backup_id if it does not already exist.
```
mikewallace1979, line 60: I think it's worth emphasizing here that we generate the backup_manifest from the files already stored on the barman server, since that is a key difference between a barman-generated backup_manifest and a backup_manifest generated by `pg_basebackup`.

tests/requirements_minimal.txt
```diff
@@ -6,4 +6,4 @@ boto3<1.11
 mock
 pytest-timeout==1.4.2
 argcomplete
-python-dateutil==1.5
+python-dateutil
```
mikewallace1979, line 9: Due to the use of [tz.UTC](https://github.com/EnterpriseDB/barman/pull/475/files#diff-bf2bad4e300610a4cebe48674911e98fa7d7a819b133035fb14e8eff6130a0a3R29) there is a version dependency with python-dateutil and python 2. I'm not sure which version `tz.UTC` was added, but `python-dateutil==2.6.1` (the OS package version on Ubuntu 18.04) does not have it (whereas 2.8.1 does).

mikewallace1979: Ok so it's 2.7.0 which introduced `tz.UTC`: https://dateutil.readthedocs.io/en/stable/tz.html#dateutil.tz.dateutil.tz.UTC

I think this can be resolved by using `tz.tzutc()` in place of `tz.UTC`.

mikewallace1979 (commented):
This is coming together nicely - only minor comments from me

tests/requirements_minimal.txt
didiermichel: indeed, tz.UTC exist since 2.7.0.  I actually encounter some issues with barman testing. python-dateutil is installed with apt-get wich provides 2.6.1.  trying pip install  `python-dateutil>=2.7.0` does not change anything whereas  `python-dateutil==2.7.0` does the trick but I find it a bit limiting.
Not sure how to solve this properly yet.

I will have a look at tz.tzutc()

barman/cli.py
```diff
@@ -1419,6 +1422,66 @@ def check_backup(args):
...
+        argument(
+            "server_name",
+            completer=server_completer,
+            help="specifies the server name for the command ",
+        ),
+        argument(
+            "backup_id", completer=backup_completer, help="specifies the backup ID"
+        ),
+    ],
+    cmd_aliases=["generate-manifest"],
```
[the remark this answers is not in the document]
didiermichel: Good point, I was not sure cmd_aliases was optional.

barman/command_wrappers.py
```diff
@@ -1014,6 +1017,58 @@ def __init__(
...
+    Wrapper class for the pg_verify system command
+    todo: findout why script path is:
+    /usr/lib/postgresql/13/bin/
+    and not /usr/bin/
+    """
+
+    # Todo  find a clean way to find exec
+    COMMAND_ALTERNATIVES = [
+        "pg_verifybackup",
+        "/usr/lib/postgresql/13/bin/pg_verifybackup",
```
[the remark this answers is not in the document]
didiermichel: I wish, but for some reason it is not always. I barman-testing docker images for example it is not in the PATH. I am not sure what is the best way to handle it.
Probably something to do at installation time. For docker images  do some fix in docker images to behave like exec is in the path 
`ENV PATH=$PATH:/usr/lib/postgresql/$PG_MAJOR/bin`
Hoping there will be no side effect with other pg binary we usually find in `usr/bin/`

What is your view on this?

tests/requirements_minimal.txt
didiermichel: So tz.tzutc() seems to work fine. I will use it.

barman/command_wrappers.py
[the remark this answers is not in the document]
mikewallace1979: You're right - `postgresql-client-13` provides the `pg_verifybackup` command in `/usr/lib/postgresql/$PG_MAJOR/bin` (along with various other client applications) but it's the `postgresql-client-common` package which adds the `pg_*` commands which Barman uses into `/usr/bin` and `postgresql-client-common` does not include `pg_verifybackup`.

This feels like something which should be fixed at the packaging level, i.e. the barman OS package would make sure pg_verifybackup is in the path (and anyone installing from source would need to make sure it is in the path themselves).

doc/manual/42-server-commands.en.md
didiermichel: Please check if the updated text is enough.

barman/command_wrappers.py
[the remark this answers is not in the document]
didiermichel: I have added a temporary fix adding pg_verifybackup path to PATH.
Not sure how to fix this at packaging level. I will check the repo.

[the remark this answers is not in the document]
mikewallace1979: For the benefit of the PR, we discussed this synchronously and concluded we'll leave it up to the user to make sure the right version of the `pg_verifybackup` binary is on the path.

We'll update the installation docs to cover this (and also `pg_receivewal` since that has a similar issue) but that does not need to block this PR.

mikewallace1979 (approved):

PR #483, opened 2021-12-30
Pass WAL archive check if bucket is missing

### LINKED ISSUE #477: [barman-cloud-check-wal-archive] Fails with NoSuchBucket
Hello barman team,

When we integrated this new command with Cloud-Native-Postgres we found that the new command is failing like below when there is no bucket (pre-created) in the objectstore.

```
{"level":"info","ts":1639242621.4066403,"msg":"CheckWalArchive() ","logging_pod":"pg-backup-minio-1","walName":"pg_wal/000000010000000000000001"}
{"level":"info","ts":1639242621.6449559,"logger":"barman-cloud-check-wal-archive","msg":"2021-12-11 17:10:21,644 [238] ERROR: Barman cloud WAL archive check exception: An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: The specified bucket does not exist","pipe":"stderr","logging_pod":"pg-backup-minio-1"}
```

This new `cloud-check-` command is run just before we issue `barman-cloud-wal-archive` for the first time or new-segment is being archived.  Hence this is expected to be consistant with `barman-cloud-wal-archive` behaviour. 

Can we not error out for `NoSuchBucket` and let `barman-cloud-wal-archive`  handle it? 

cc: @mnencia @gbartolini

mikewallace1979:
Updates barman-cloud-check-wal-archive such that the check passes if the
target bucket does not exist *and* we were able to successfully connect
to the cloud provider.

If we cannot successfully connect to the cloud provider then the check
errors and exits with status 2.

Closes #477.

### DIFF
--- a/barman/clients/cloud_check_wal_archive.py
+++ b/barman/clients/cloud_check_wal_archive.py
@@ -22,6 +22,7 @@ from barman.clients.cloud_cli import (
     create_argument_parser,
     GeneralErrorExit,
     OperationErrorExit,
+    NetworkErrorExit,
     UrlArgumentType,
 )
 from barman.cloud import configure_logging, CloudBackupCatalog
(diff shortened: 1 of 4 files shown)


barman/clients/cloud_check_wal_archive.py
```diff
@@ -38,6 +38,12 @@ def main(args=None):
 
     try:
         cloud_interface = get_cloud_interface(config)
+        if not cloud_interface.test_connectivity():
+            # Deliberately raise an error if we cannot connect
+            raise CloudProviderError("Cannot connect to cloud provider")
+        if cloud_interface.bucket_exists is not True:
```
didiermichel, line 44: Something like `if not cloud_interface.bucket_exists` feel more readable to me.

didiermichel (commented):
Just this slight personal view to report.

Issue #438, opened 2021-10-27
Add file-not-found dedicated exit code to barman-cloud-wal-restore


mnencia:
It would be helpful if `barman-cloud-wal-restore` could have a dedicated exit code when everything works, but the file is not present in the source bucket/storage container.

This change could help the caller to identify configuration issues over normal conditions.

mikewallace1979:
Follow same approach that `pg_isready` is using

PR #484, opened 2022-01-04
Dev/438 improve wal restore exit codes

mikewallace1979:


didiermichel:
I just read the issue related to exit codes. And reading at changes I was thinking: I will never remember what code number refers to . If I need to extend things and raise an error I will have to search for code references that I can find in  #438.
Sot I thought maybe it is worth having it somewhere in the code. 

Could be comments:
```
Follow same approach that pg_isready is using: https://www.postgresql.org/docs/14/app-pg-isready.html
Specifically:
0 if the WAL is found
1 if the WAL isn't found
2 network errors
3 CLI-level errors
```

Or directly in the code (even with comment in class)
```
class NetworkErrorExit(SystemExit):
    """
    Dedicated to exit on network error context 
    """
    def __init__(self):
        super(SystemExit, self).__init__(2)

# Somewhere else

raise NetworkErrorExit()
```

I find it more easy to read when investigating for example.
SystemExit can still be used on any other context unrelated.

Sot this is just to open a discussion about it. What do you think?

mikewallace1979:
@didiermichel I think that's a good idea. I'm wary about changing error codes and making the intentions clearer to future maintainers is definitely a smart move.

I also realised this commit doesn't communicate the meaning to the users via the documentation so I'll improve things there too.

mikewallace1979:
@didiermichel Ok - what do you think?

If we're happy this is an improvement (I definitely am) I will squash and reword the commit messages to reflect the changes.

barman/clients/cloud_backup_delete.py
```diff
@@ -177,7 +184,7 @@ def _delete_backup(
             cloud_interface.delete_objects([backup_info_path])
         except Exception as exc:
             logging.error("Could not delete backup %s: %s", backup_id, force_str(exc))
-            raise SystemExit(2)
+            raise OperationErrorExit()
```
didiermichel, line 187: exit value changes from 2 to 1. Not sure if it is on purpose.

didiermichel (approved):
It looks good.
I find it more readable. even documentation :)

barman/clients/cloud_backup_delete.py
mikewallace1979: This was deliberate since the error is something which occurred while attempting to perform the deletion (but we don't know anything more than that).

PR #491, opened 2022-01-11
Expose azure upload parameters #472

### LINKED ISSUE #472: Speed up barman cloud WAL archiving
`barman-cloud-wal-archive` can take around 120s to upload a 16MB WAL. This is not amazing performance and can lead to WALs backing up under modest workloads.

The goal of this ticket is to:

1. Understand why WAL upload is taking so long, and:
2. Speed up the WAL upload either by making better/aternative use of the cloud provider API.

mikewallace1979:
Improves the way we use the Azure container client to upload WALs so that concurrent uploading of chunks can be used, if so desired. This can yield more than double the throughput achieved with single-request uploads.

The defaults are set in the cloud interface to the values which have experimentally yielded a measurable improvement however the defaults in the CLI arguments have been set to keep the behaviour consistent with previous versions of Barman Cloud. This is so that users can opt-in to the concurrent upload method. If things go well we can change the CLI argument defaults to match those in the cloud interface.

Closes #472

### DIFF
--- a/barman/clients/cloud_walarchive.py
+++ b/barman/clients/cloud_walarchive.py
@@ -36,7 +36,7 @@ from barman.clients.cloud_cli import (
 from barman.cloud import configure_logging
 from barman.cloud_providers import get_cloud_interface
 from barman.exceptions import BarmanException
-from barman.utils import force_str
+from barman.utils import check_positive, check_size, force_str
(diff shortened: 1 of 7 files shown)


didiermichel (approved):
This looks good to me.

PR #502, opened 2022-01-25
Fix inactive server show-servers output with --format=json #488

### LINKED ISSUE #488: "barman --format json show-server the_server" produces _EXECPTION when server is inactive
Tested wit barman 2.17

    $ barman --format json show-server bdu1
    {"_EXCEPTION": ["'bdu1'\nSee log file for more details."], "bdu1 (inactive)": {}}

    $ barman show-server bdu1                                                                                         
    Server bdu1 (inactive):
            active: False
            archive_timeout: 0
            archiver: False
            ...

mikewallace1979:
Changes the implementation of show-servers so that the status
description is passed to the output writer as a separate variable
instead of appending it directly to the name of the server. This allows
the JsonOutputWriter to use just the name of the server as the key when
creating the entry in the output dict.

This fixes a bug where the key was previously the name *and* the
description, meaning the result_show_server method would then cause a
KeyError when it attempted to access the results dict by the server
name alone.

Because the JSON output is keyed by server name the description is
simply added under the description key in the output. So that this
does not get overwritten the output logic is changed so that message
values of None are skipped if there is already a value for a particular
key in the output dict.

Closes #488.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -18,12 +18,14 @@
 
 from argparse import ArgumentTypeError
 
+import json
 import os
 import pytest
 import sys
-from mock import Mock, patch
+from mock import MagicMock, Mock, patch
 
 import barman.config
+from barman import output
 from barman.cli import (
     ArgumentParser,
     argument,
@@ -36,6 +38,7 @@ from barman.cli import (
     OrderedHelpFormatter,
     recover,
     keep,
+    show_servers,
 )
 from barman.exceptions import WalArchiveContentError
 from barman.infofile import BackupInfo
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -681,3 +684,123 @@ def test_barman_check_wal_archive_content_error(
...
+        with pytest.raises(SystemExit):
+            # AND barman show-servers runs
+            show_servers(mock_args)
+
+        # THEN nothing is sent to stderr
+        out, err = capsys.readouterr()
+        assert "" == err
+
+        # AND the command output includes the description and server name
+        assert "%s%s:" % (self.test_server_name, expected_description)
```
richyen, line 755: PR generally looks good--I'm curious to know what this `assert` is looking for?  Making sure the two args aren't `None`?

mikewallace1979: Good catch - that should be testing that the server and description appear in the output, something like:

    assert "%s%s:" % (self.test_server_name, expected_description) in out

richyen (approved):

PR #505, opened 2022-01-28
Fix barman-cloud-check-wal-archive to observe `-t` option

### LINKED ISSUE #503: `-t` option not respected by barman-cloud-check-wal-archive
All the other barman-cloud commands exit after the connectivity test when the `-t` option is used but barman-cloud-check-wal-archive continues. This should be changed to match the behaviour of the other commands.

richyen:
Closes #503

### DIFF
--- a/barman/clients/cloud_check_wal_archive.py
+++ b/barman/clients/cloud_check_wal_archive.py
@@ -47,6 +47,9 @@ def main(args=None):
         if not cloud_interface.test_connectivity():
             # Deliberately raise an error if we cannot connect
             raise NetworkErrorExit()
+        # If test is requested, just exit after connectivity test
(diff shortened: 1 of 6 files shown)


mikewallace1979 (changes requested):
Would you be able to add a unit test to verify the early exit with status 0? Something similar to [this barman-cloud-keep test](https://github.com/EnterpriseDB/barman/blob/2f92fa26784c67a845a2a6e4c06e0daf523ead4a/tests/test_barman_cloud_backup_keep.py#L57-L74) would be fine.

mikewallace1979 (approved):

mikewallace1979:
Merged in #512

PR #506, opened 2022-01-28
handle 'latest' and 'current' as valid target-tli values 

### LINKED ISSUE #305: target-tli should allow "current" or "latest" in addition to positive integer
https://github.com/2ndquadrant-it/barman/blob/1bcdd6bcbaed9bd0b5882bc9543036e967ff3368/barman/cli.py#L339
defines target-tli as a positive integer but it can also be 'current' or 'latest' (which is the default value).
See [postgresql doc](https://www.postgresql.org/docs/13/runtime-config-wal.html#GUC-RECOVERY-TARGET-TIMELINE).

richyen:
Closes #305

### DIFF
--- a/barman/utils.py
+++ b/barman/utils.py
@@ -620,18 +620,16 @@ def check_tli(value):
 
     :param value: str containing the value to check
     """
+    retval = value
     if value is None:
         return None
     try:
-        int(value)
-    except Exception:
-        if not value.isdigit():
-            assert value in ["current", "latest"], "'%s' is not a valid TLI" % value
-        else:
-            raise ArgumentTypeError("'%s' is not a valid positive integer" % value)
-    if value < 1:
-        raise ArgumentTypeError("'%s' is not a valid positive integer" % value)
-    return value
+        retval = check_positive(value)
+    except ArgumentTypeError:
+        #  Value is not a positive integer so check for the valid string alternatives here
+        if value not in ["current", "latest"]:
+            raise ArgumentTypeError("'%s' is not a valid TLI" % value)
+    return retval
 
 
 def check_size(value):
--- a/barman/utils.py
+++ b/barman/utils.py
@@ -614,6 +614,26 @@ def check_positive(value):
     return int_value
 
 
+def check_tli(value):
+    """
+    Check for a positive integer option, and also make "current" and "latest" acceptable values
+
+    :param value: str containing the value to check
+    """
+    if value is None:
+        return None
+    try:
+        int(value)
+    except Exception:
+        if not value.isdigit():
+            assert(value in ["current", "latest"]), "'%s' is not a valid TLI" % value
+        else:
+            raise ArgumentTypeError("'%s' is not a valid positive integer" % value)
+    if value < 1:
+        raise ArgumentTypeError("'%s' is not a valid positive integer" % value)
+    return value
+
+
 def check_size(value):
     """
     Check user input for a human readable size
--- a/barman/utils.py
+++ b/barman/utils.py
@@ -624,10 +624,8 @@ def check_tli(value):
         return None
     if value in ["current", "latest"]:
         return value
-    try:
+    else:
         return check_positive(value)
-    except ArgumentTypeError as err:
-        raise ArgumentTypeError("%s" % err)
 
 
 def check_size(value):
--- a/barman/utils.py
+++ b/barman/utils.py
@@ -626,7 +626,7 @@ def check_tli(value):
         int(value)
     except Exception:
         if not value.isdigit():
-            assert(value in ["current", "latest"]), "'%s' is not a valid TLI" % value
+            assert value in ["current", "latest"], "'%s' is not a valid TLI" % value
         else:
             raise ArgumentTypeError("'%s' is not a valid positive integer" % value)
     if value < 1:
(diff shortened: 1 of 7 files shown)


mikewallace1979 (changes requested):
Thanks for picking this one up - looks like a good first pass at it.

As well as the in-line comments, we also need to handle the current/latest values in [get_required_xlog_files](https://github.com/EnterpriseDB/barman/blob/8656d01bc82bfafff7f417ff7e3aa43c82b33a04/barman/server.py#L1433) because currently values of `"current"` and `"latest"` mean the `tli > target_tli` check on L1456 will always return `False` which means we'll never skip WALs on a higher timeline than the target_tli.

barman/utils.py
```diff
@@ -614,6 +614,24 @@ def check_positive(value):
...
+    retval = value
+    if value is None:
+        return None
+    try:
+        retval = check_positive(value)
+    except ArgumentTypeError:
+        #  Value is not a positive integer so check for the valid string alternatives here
+        if value not in ["current", "latest"]:
+            raise ArgumentTypeError("'%s' is not a valid TLI" % value)
+    return retval
```
mikewallace1979, line 632: This could be further simplified if we check the allowed string values first. That way we don't have to do the `retval = value` assignment and can also avoid using the exception to handle normal behaviour (apologies, I know that was my suggestion from the last review), e.g.:

```
if value in ["current", "latest"]
    return value
try:
    return check_positive(value)
except ArgumentTypeError:
    raise ArgumentTypeError("'%s' is not a valid TLI" % value)
```

mikewallace1979:
@richyen Thanks for making the suggested changes - it's just handling the target_tli string values in Server.get_required_xlog_files which is left to do I think.

barman/utils.py
```diff
@@ -626,8 +626,8 @@ def check_tli(value):
         return value
     try:
         return check_positive(value)
-    except ArgumentTypeError:
-        raise ArgumentTypeError("'%s' is not a valid TLI" % value)
+    except ArgumentTypeError as err:
+        raise ArgumentTypeError("%s" % err)
```
mikewallace1979, line 630: Since we're now just re-raising the original exception we can do away with the `try/except` block altogether.

mikewallace1979 (approved):

mikewallace1979:
I'm going to take a look at where we could add test for this but I think other than that it's ready for merging.

mikewallace1979:
While adding unit tests for `_set_pitr_targets` I realised there are some subtleties around [L474](https://github.com/EnterpriseDB/barman/pull/506/files#diff-97e4f053b533ec26978215af2e61018adeb2f8863e04500b32b4587db9486a86R474) which need considering:

The lines in question are:

```
        d_tli = target_tli != backup_info.timeline and calculated_target_tli

        # Detect PITR
        if target_time or target_xid or d_tli or target_name or d_immediate or d_lsn:
            recovery_info["is_pitr"] = True
```

The previous behaviour, where we support numeric timelines only, will cause barman to skip configuring PITR options if the requested timeline is the same timeline as the backup being recovered (and no other recovery targets are specified). When recovering PostgreSQL versions < 12 this makes sense because the [default behaviour of PostgreSQL is to recover to the most recent WAL on the timeline which was current when the backup was taken](https://www.postgresql.org/docs/11/recovery-target-settings.html#RECOVERY-TARGET-TIMELINE).

To preserve this behaviour we can compare the calculated tli, not the original `target_tli` value, so that `current` and `latest` have been resolved to the numeric timeline before we decide whether we can skip PITR or not. Specifically we would do the following:

```
        d_tli = calculated_target_tli != backup_info.timeline and calculated_target_tli
```

However, the previous behaviour doesn't seem to make sense with PostgreSQL versions >= 12. With these versions, the [default recovery behaviour is to recover to the latest timeline found in the archive](https://www.postgresql.org/docs/12/runtime-config-wal.html#GUC-RECOVERY-TARGET-TIMELINE). If we were to leave things as they are then it would not be possible to use `--target-tli=current` to recover to the timeline which was current when the backup was taken. Furthermore, using `--target-tli=latest` would cause PITR options to be set, even though they could be skipped since the default behaviour is to recover to the latest timeline.

So we should be comparing `calculated_target_tli` to `backup_info.timeline` if `backup_info.version < 120000` and we should be comparing `calculated_target_tli` to the latest archived timeline if `backup_info.version >= 120000`.

There is a second issue, not so much related to the PR but related to the behaviour of Barman when recovering WALs, which is that if no `target_tli` is specified, Barman will assume [the timeline which was current when the backup was taken is the most recent timeline for which WALs are required](https://github.com/EnterpriseDB/barman/pull/506/files#diff-40d894681de880b0b65b0db09b8c805d98a794a302d7fdcdb82e205f3952a12fL1559-R1573). This means that if we are recovering a backup with no `--target-tli` option and the backup is not on the most recent timeline, then Barman will not copy WALs from newer timelines than the one current at the time of the backup. This will cause the recovered PostgreSQL server to start without all the WALs required to restore to the latest timeline. This isn't an issue if the latest timeline is specified by number and would also not be an issue with this patch if `--target-tli=latest` is used explicitly. It is also not an issue if `--get-wal` is specified, since PostgreSQL will fetch the WALs itself on recovery.

Assuming I'm right about these things (and it'd be great for a second opinion here) then these are both existing bugs in Barman due to it needing to be updated to handle changes in `recovery_target_timeline` defaults introduced in PostgreSQL 12. We could therefore file a new issue and fix them in a separate PR rather than continuing to block this one.

mikewallace1979:
Merged in #537

PR #513, opened 2022-02-02
Minor docs fixes

mikewallace1979:
Fixes a couple of errors in the docs noticed over the last few days


didiermichel (approved):
I find it wierd that black fixes so much things
Maybe we should stop cashing tox contexts. It makes black have big version steps

PR #515, opened 2022-02-03
Capture pipe return code and pass along to

### LINKED ISSUE #279: barman-wal-archive --test and barman receive-wal --reset return exit code 0 even on failure
barman 2.10, PostgreSQL 12, Ubuntu 18.04

As recommended in the barman documentation I run `barman-wal-archive --test` to check if WAL archiving works before setting archive_command in the PostgreSQL configuration. I'd like to automate this and check that `barman-wal-archive` succeeded, but this is difficult, because it returns exit code 0 even if the server is inactive or does not exist at all, e.g.

```
$ barman-wal-archive --test HOSTNAME nosuchserver DUMMY && echo Success?
ERROR: Unknown server 'nosuchserver'
b''
Success?
```

Please make it return exit code 0 only on success.

richyen:
Closes #279

### DIFF
--- a/tests/test_barman_wal_archive.py
+++ b/tests/test_barman_wal_archive.py
@@ -162,11 +162,12 @@ class TestMain(object):
     def test_connectivity_test_ok(self, popen_mock, capsys):
 
         popen_mock.return_value.communicate.return_value = (b"Good test!", b"")
+        popen_mock.return_value.returncode = 255
 
         with pytest.raises(SystemExit) as exc:
             walarchive.main(["a.host", "a-server", "--test", "dummy_wal"])
 
-        assert exc.value.code == 0
+        assert exc.value.code == 255
         out, err = capsys.readouterr()
         assert "Good test!" in out
         assert not err
--- a/tests/test_barman_wal_archive.py
+++ b/tests/test_barman_wal_archive.py
@@ -161,7 +161,10 @@ class TestMain(object):
     @mock.patch("barman.clients.walarchive.subprocess.Popen")
     def test_connectivity_test_returns_subprocess_output(self, popen_mock, capsys):
 
-        popen_mock.return_value.communicate.return_value = (b"Tested subprocess return code percolation", b"")
+        popen_mock.return_value.communicate.return_value = (
+            b"Tested subprocess return code percolation",
+            b"",
+        )
         popen_mock.return_value.returncode = 255
 
         with pytest.raises(SystemExit) as exc:
--- a/tests/test_barman_wal_archive.py
+++ b/tests/test_barman_wal_archive.py
@@ -159,9 +159,9 @@ class TestMain(object):
         assert "Remote 'barman put-wal' command has failed" in err
 
     @mock.patch("barman.clients.walarchive.subprocess.Popen")
-    def test_connectivity_test_ok(self, popen_mock, capsys):
+    def test_connectivity_test_returns_subprocess_output(self, popen_mock, capsys):
 
-        popen_mock.return_value.communicate.return_value = (b"Good test!", b"")
+        popen_mock.return_value.communicate.return_value = (b"Tested subprocess return code percolation", b"")
         popen_mock.return_value.returncode = 255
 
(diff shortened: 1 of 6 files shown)


mikewallace1979 (changes requested):
You'll need to patch `returncode` of `popen_mock.return_value` in [test_barman_wal_archive.TestMain.test_connectivity_test_ok](https://github.com/EnterpriseDB/barman/blob/851ed11d3a9c7841b54783db9bf0caee70143f0b/tests/test_barman_wal_archive.py#L161-L172) so that the test gets the expected exit code.

mikewallace1979:
Looking at #279 it also mentions `barman receive-wal` will exit with status zero in some error cases - the specific case in the ticket is when there's a `LockFileBusy` exception: https://github.com/EnterpriseDB/barman/blob/d8ba7cabab9217534e2f9897be95ef9b3ee29392/barman/server.py#L2571-L2582

Fixing this is a bit trickier because of the way the main Barman program handles output - if we changed the `output.info` calls to `output.error` calls we'd be able to use the `exit_code` kwarg to set a specific error code here though that may not be the best fix.

tests/test_barman_wal_archive.py
```diff
@@ -162,11 +162,12 @@ def test_error_barman(self, rpw_mock, tmpdir, capsys):
     def test_connectivity_test_ok(self, popen_mock, capsys):
 
         popen_mock.return_value.communicate.return_value = (b"Good test!", b"")
+        popen_mock.return_value.returncode = 255
```
mikewallace1979, line 165: Because the test is testing the successful case it would be clearer to use a `0` here instead of `255`.

richyen: hmm, I'm a bit confused here.  If I set `popen_mock.return_value.returncode = 0`, then the subsequent `assert exc.value.code == 255` fails (a few lines later).

`walarchive.main(["a.host", "a-server", "--test", "dummy_wal"]` should return 255, since the host and server don't actually exist.  Here's the output when I manually run it:

```
-bash-4.2$ barman-wal-archive a.host a-server --test dummy_wal 
                                                               
-bash-4.2$ echo $?                                             
255                                                            
```

How should I properly construct the test?  Any way not to assign `returncode` to the mock object and see if `walarchive.main()` will cause it to exit with `255`?

mikewallace1979: The test itself is fine but the name of the test tells us it should be testing the "ok" or successful case, so what I was thinking was that we'd have `popen_mock` return a successful exit code and then leave the rest of the test to do what it did originally, i.e. assert that the exit code returned by the command itself is zero.

Alternatively we could continue to use the `255` error code but change the test name to something like `test_connectivity_test_returns_subprocess_output` so that it better matches the behaviour the test is verifying.

mikewallace1979 (approved):
Looks good.

@richyen Were you going to fix the `barman receive-wal` exit code here or in a separate PR?

richyen:
> Looks good.
> 
> @richyen Were you going to fix the `barman receive-wal` exit code here or in a separate PR?

I can put the `_ok` test back in, but would the test be misleading? i.e., we are telling it to return 0 even though it shouldn't be returning 0 (since we're passing bogus args into it).  Or what would the `_ok` test be confirming in that case? Maybe I'm misunderstanding unit test philosophy?

mikewallace1979:
> I can put the `_ok` test back in, but would the test be misleading? i.e., we are telling it to return 0 even though it shouldn't be returning 0 (since we're passing bogus args into it). Or what would the `_ok` test be confirming in that case? Maybe I'm misunderstanding unit test philosophy?

Sorry Rich - I wasn't referring to that test here (I'm happy with the last change where the name was updated). I was referring to the `barman receive-wal` command which is also mentioned in issue #279 as a command which can return status 0 even though it has failed.

Given it concerns a different part of the codebase I think it makes sense to merge this as it is and make a separate PR for `barman receive-wal`.

mikewallace1979:
Merged in #540

Issue #510, opened 2022-01-31
Update timestamp format in barman diagnose output


mikewallace1979:
Currently `barman diagnose` provides timestamps as [ctime](https://docs.python.org/3/library/datetime.html#datetime.date.ctime) strings which (crucially) lack timezone information.

As a minimum we need to include timezone information (and confirm whether the timezone is of the PostgreSQL server or the Barman server) but we should also consider using something other than ctime (ISO8601 or epoch time for example). If we're going to use epoch time then that should probably be a separate field so that we still have human-friendly timestamps in the output.

didiermichel:
I had some thoughts about it, I like the idea of using ISO8601 date format. But Displaying local server time can be an issue with DST. 
This output will also be used by pg_backup_api, so it is both a system interface and a human interface.

Here are concerned fields with current values:

```
global.system_info.timestamp = "Thu Feb 17 10:53:47 2022"
global.servers.xxx.backups.yyy.begin_time = 'Thu Feb 17 10:45:11 2022'
global.servers.xxx.backups.yyy.end_time = "Thu Feb 17 10:45:21 2022"
global.servers.xxx.status.last_archived_time = "Thu Feb 17 10:45:22 2022"
global.servers.xxx.status. last_failed_time = "Thu Feb 17 10:45:22 2022"
global.servers.xxx.status.stats_reset = "Thu Feb 17 09:11:54 2022",
```

What I suggest is to output UTC time for all fields so there is no data loss. And to keep the knowledge of server timezone 

Have two fields in global.system_info: 
timestamp in local server time
utc_time in UTC

Ex:
```
global.system_info.timestamp = "2022-02-17T11:53:47.944515+01:00"
global.system_info.utc_time = "2022-02-17T10:53:47.944515+00:00"
global.servers.xxx.backups.yyy.begin_time = '2022-02-17T10:45:11.672345+00:00'
global.servers.xxx.backups.yyy.end_time = "2022-02-17T10:45:21.664855+00:00"
global.servers.xxx.status.last_archived_time = "2022-02-17T10:45:22.331255+00:00"
global.servers.xxx.status. last_failed_time = "2022-02-17T10:45:22.644879+00:00"
global.servers.xxx.status.stats_reset = "2022-02-17T09:11:54.156842+00:00",
```

PR #528, opened 2022-02-21
Add NewBarmanEncoder to encode json with date as isoformat (#510)

didiermichel:


barman/diagnose.py
```diff
@@ -80,6 +83,19 @@ def exec_diagnose(servers, errors_list):
         diagnosis["servers"][name]["status"] = server.get_remote_status()
         # backup list
         backups = server.get_available_backups(BackupInfo.STATUS_ALL)
+        # update date format for each backup begin_time and end_time and ensure local timezone.
+        # This code is a duplicate from BackupInfo.to_json()
+        # This should be temporary to keep original behavior for other usage.
```
didiermichel, line 88: Changes in diagnose.py file are really ugly, but it was the easiest way to not impact other `BackupInfo.to_json()`  users.

If we plan to use isoformat() for all `BackupInfo.to_json()`  users then it is a temporary thing. Otherwise, we should consider finding another solution.

mikewallace1979: Ideally this would be a temporary thing - we can file an issue for replacing ctime strings with ISO 8601 format strings in all the output and that could land in a major version change, which would allow us to remove this code and probably revert back to a single Barman encoder.

barman/utils.py
```diff
@@ -297,30 +297,101 @@ class BarmanEncoder(json.JSONEncoder):
...
+        :return: None|str
+        """
         if isinstance(obj, Version):
             return str(obj)
-        # Let the base class default method raise the TypeError
-        return super(BarmanEncoder, self).default(obj)
+
+
+class NewBarmanEncoder(BarmanEncoder):
+    @staticmethod
```
mikewallace1979, line 382: Could you add a comment stating that this is intended to replace `BarmanEncoder` at the next major version change (assuming we're all happy with that approach)?

mikewallace1979 (commented):
Looks great - one minor comment from me about maybe adding a comment to `NewBarmanEncoder`, though maybe that information could go in the commit message instead.

barman/diagnose.py
didiermichel: I like this idea.
Hopping no one will miss ctime output.

barman/utils.py
hlovdal: Speaking of names, "new" tends to get old quite fast and I think something like BarmanEncoderV2 or similar would be better.

didiermichel: Absolutely agreed. I will add comment and rename it.

amenonsen (approved):

mikewallace1979 (approved):

Issue #559, opened 2022-04-24
Some imporvements on barman documentation 


elhananjair:
Hello there, I am trying to set up barman for the first time and its a little bit confusing, the documentation. I have created a user called barman on hostname called 'pg' as the doc says, and then at this point: 
> Make sure you test the following command before proceeding:
  barman@backup$ psql -c 'SELECT version()' -U barman -h pg postgres

I got lost, **barman@backup**, I thought I should test that on **pg** instance not on the **backup** as the documentation says, in addition to that the user must be **postgres** to run the above command. Actually, I tested this as **postgres@pg** and it seems working like that. Here is an output in my case. 
> PostgreSQL 13.4 on x86_64-redhat-linux-gnu, compiled by gcc (GCC) 11.2.1 2021120
3 (Red Hat 11.2.1-7), 64-bit
(1 row)

I just wanna suggest a correction if my correction is right, and a little bit of improvement on the documentation, to grasp the concepts easily. 
Thank you.

mikewallace1979:
The idea behind running `psql -c 'SELECT version()' -U barman -h pg postgres` is to verify that the barman user on the barman host can successfully connect to the "postgres" database on the PostgreSQL host. It therefore should be run as the barman user on the barman host, not as the postgres user on the PostgreSQL host.

We'll update the documentation to make this clearer.

elhananjair:
> The idea behind running `psql -c 'SELECT version()' -U barman -h pg postgres` is to verify that the barman user on the barman host can successfully connect to the "postgres" database on the PostgreSQL host. It therefore should be run as the barman user on the barman host, not as the postgres user on the PostgreSQL host.
> 
> We'll update the documentation to make this clearer.

@mikewallace1979 Thank you so much for clarifying things for me, and now I saw the commit and that's perfect. But I am still little bit confused, am sorry to reply to the closed issue. 

On this line,

> You need to make sure that the backup server can connect to the PostgreSQL server on pg as superuser or, from PostgreSQL 10 or higher, that the correct set of privileges are granted to the user that connects to the database.

You can create a specific superuser in PostgreSQL, named barman, as follows:

postgres@pg$ createuser -s -P barman

--- 
[...]

PR #560, opened 2022-04-25
Clarify docs around verifying barman/pg connection

mikewallace1979:
Updates the documentation around the following command in the docs:

    barman@backup$ psql -c 'SELECT version()' -U barman -h pg postgres

The documentation is updated so that it explains the reason for running
the command as well as stating the user and host which should be used.

Thanks to @elhananjair for reporting the issue.


mikewallace1979:
closes #559

Issue #555, opened 2022-04-13
Compress base backups with pg_basebackup (backup_method = postgres)


theadamwright:
**What?** 

1. Support the compression of base backups when using the `pg_basebackup` method in Barman e.g., `--gzip` 
2. Expose the gzip compression of tar file output by being able to configure the compression level(0 through 9) e.g.,`--compress=level`
3. Support recovery of compressed backups: simple implementation that unzips on the Barman server and copies or preferably decompresses on the fly like `barman-cloud`. 

**Why?** 

Reduce the storage footprint of Postgres backups when using Barman.

snailxr:
Or can send the compressed backups to the target server and decompress them.

PR #571, opened 2022-05-17
Optionally compress pg_basebackup backups #555

mikewallace1979:
This is a work-in-progress PR - the implementation will likely change but the CLI args and backup.info field should be stable unless there is feedback indicating otherwise.


barman/backup.py
```diff
@@ -579,6 +579,15 @@ def backup(self, wait=False, wait_timeout=None):
                 self.server, backup_id=datetime.datetime.now().strftime("%Y%m%dT%H%M%S")
             )
             backup_info.set_attribute("systemid", self.server.systemid)
+            if self.server.config.backup_compression is not None:
+                backup_info.set_attribute(
+                    "compression", self.server.config.backup_compression
+                )
+            elif self.server.config.backup_compression_level is not None:
```
mikewallace1979, line 586: I don't think it should be the job of the BackupManager to know this detail which concerns the way pg_basebackup interprets its arguments.

It would be better to just set `backup_compression` in server.config at an earlier stage, ideally in an abstraction which sits somewhere on top of pg_basebackup.

barman/backup_executor.py
```diff
@@ -1593,6 +1601,60 @@ def stop_backup(self, backup_info):
...
+        elif backup_info.compression == "lz4":
+            import lz4.frame
+
+            yield lz4.frame.open(tar_path, mode="rb")
+        elif backup_info.compression == "zstd":
+            import pyzstd
+
+            yield pyzstd.open(tar_path, "rb")
+
+    def _read_compressed_backup_label(self, backup_info):
```
mikewallace1979, line 1631: Reading a `backup_label` from a backup compressed by pg_basebackup does seem like the job of the `PostgresBackupStrategy` so I think it is reasonable for this code to live here.

barman/cli.py
```diff
@@ -447,6 +489,20 @@ def backup(args):
             server.config.parallel_jobs = args.jobs
         if hasattr(args, "bwlimit"):
             server.config.bandwidth_limit = args.bwlimit
+        if args.compression_type is not None:
```
mikewallace1979, line 492: As well as the validation of individual arguments there is a subsequent validation which needs to happen which is that the resulting `config` fields (which will be a combined result of these arguments and the configuration) combine to an allowed combination.

The PostgresBackupExecutor has a [validate_configuration](https://github.com/EnterpriseDB/barman/blob/9682b304e8ab2a344074d117d062bd668f8d095e/barman/backup_executor.py#L238) method so perhaps that is a good place to do such a thing. We could also have the executor set the backup_compression option explicitly if it is not set but `backup_compression_levell` is set and avoid the special case in BackupManager on L586.

barman/backup_executor.py
```diff
@@ -1593,6 +1601,60 @@ def stop_backup(self, backup_info):
             # Skip switching XLOG if a standby server
             pass
 
+    @contextmanager
+    def _get_tar_fileobj(self, backup_info):
```
mikewallace1979, line 1605: This function is dealing with the fact that the tar file containing the backup_label needs to be opened using a specific compression type - I'm in two minds about whether this is the best place for that knowledge. On one hand, a backup strategy perhaps shouldn't need to care about specific compresion, but on the other hand I don't think there's currently a better place for this knowledge to go.

barman/command_wrappers.py
```diff
@@ -880,6 +884,61 @@ class PgBaseBackup(PostgreSQLClient):
 
     COMMAND_ALTERNATIVES = ["pg_basebackup"]
 
+    def _get_compression_args(
```
mikewallace1979, line 887: This does belong in the command wrapper because it's converting the compression options into CLI args specific to this command. The code itself is a bit hard to follow with all the conditionals but in principle I think it's ok here.

barman/config.py
```diff
@@ -70,6 +70,10 @@
 
 CREATE_SLOT_VALUES = ["manual", "auto"]
 
+# Config values relating to pg_basebackup compression
+BASEBACKUP_COMPRESSIONS = ["gzip", "lz4", "zstd"]
```
mikewallace1979, line 74: We have to define these here because nearly every module imports config and we'd end up with a circular dependency.

barman/cli.py
mikewallace1979: This was a nice idea but it won't work - validate_configuration gets called when the Server is created which is before the arguments are parsed. We would therefore need to put such validation somewhere else.

mikewallace1979:
This PR was useful for exploring ideas however the current implementation has diverged a little. I'm going to push the work to a new branch and close this PR, keeping the branch around for a while for reference.

PR #573, opened 2022-05-19
Dev/555 recover compressed backup

didiermichel:


mikewallace1979 (commented):
Looks like a good start - I had a hack about with the branch locally and was able to copy the backup into the staging area.

I added some comments having read through this PR and the surrounding code a bit more.

One thing which I missed is that the recover method handles [restoring the config files separately to the rest of the backup](https://github.com/EnterpriseDB/barman/blob/9682b304e8ab2a344074d117d062bd668f8d095e/barman/recovery_executor.py#L341-L343) - this is so it can do additional validation, config generation (for recovery options) and comment out unsafe options. Currently the _map_temporary_config_files function builds the list of config files using the backup source directory - of course the files in there are compressed so it finds no config files and therefore the special-case handling of config files doesn't happen. The most obvious result of this is that postgresql.auto.conf does not end up with the PITR settings when doing PITR with remote_ssh_command.

So however we implement restore, _map_temporary_config_files is going to need to change - it must either:
1. Look in the (remote or local) staging directory to find the files (if the backup is uncompressed in the staging directory).
2. Look in the compressed backup tarball to find the files and store them in a temporary directory (if the backup is extracted directly to the destination on the PG server).

PR #578, opened 2022-05-26
Fix google dependencies for python 3.5

mikewallace1979:
Pins protobuf to below 3.18 as python 3.5 support was formally dropped
in 3.18.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,11 @@ setup(
         "snappy": [
             "python-snappy == 0.6.0"
         ],  # version is limited py python2.7 see issue #529
-        "google": ["google-cloud-storage"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -103,7 +103,11 @@
         "snappy": [
             "python-snappy == 0.6.0"
         ],  # version is limited py python2.7 see issue #529
-        "google": ["google-cloud-storage"],
+        "google": [
+            "google-cloud-storage",
+            # python 3.5 support was dropped in 3.18
```
didiermichel, line 108: It's a good habit to explain limiting version and reason for later update.

didiermichel (approved):

PR #579, opened 2022-05-27
Add PostgreSQL 15 support

### LINKED ISSUE #564: PostgreSQL 15 support
Exclusive backups will be completely removed from PG 15 so Barman must not allow exclusive backups to be used.

It is also worth considering making concurrent backups the default in cases where it is not currently so.

Additionally, the [commit which removed exclusive backups](https://github.com/postgres/postgres/commit/39969e2a1e4d7f5a37f3ef37d53bbfe171e7d77a) also renamed the start/stop backup functions from `pg_{start,stop}_backup` to `pg_backup_{start,stop}`.

Also we need to keep a log of changes users need to be aware of:

1. If default backup method remains `exclusive` then they will need to add `backup_options = concurrent_backup`.
2. If they are not using a superuser they will need to update the function names and signatures in the permissions for the barman PG user <-- We will also need to update the Barman docs so that the correct function names/signatures are there too.

mikewallace1979:
Adds support for PostgreSQL 15.

This has been tested locally using a [barman-testing branch](https://github.com/EnterpriseDB/barman-testing/tree/add-pg15-support) and we can also use a [barman-it-set branch](https://github.com/EnterpriseDB/barman-it-set/tree/test-on-pg15) to test this in CI before merging.

Edit: Actually we need to use a [barman-packaging branch](https://github.com/EnterpriseDB/barman-packaging/tree/pg15-test) to test with since the integration tests always pull built packages and it's easier to just point the acceptance tests in barman-packaging to the updated integration tests branch.

### DIFF
--- a/barman/backup_executor.py
+++ b/barman/backup_executor.py
@@ -1669,6 +1669,23 @@ class ExclusiveBackupStrategy(BackupStrategy):
                     False,
                     hint="cannot perform exclusive backup on a standby",
                 )
+        check_strategy.init_check("exclusive backup supported")
+        try:
+            if self.postgres and self.postgres.server_version < 150000:
+                check_strategy.result(self.server_name, True)
+            else:
+                check_strategy.result(
+                    self.server_name,
+                    False,
+                    hint="exclusive backups not supported "
+                    "on PostgreSQL %s" % self.postgres.server_major_version,
+                )
+        except PostgresConnectionError:
+            check_strategy.result(
+                self.server_name,
+                False,
+                hint="unable to determine postgres version",
+            )
 
 
 class ConcurrentBackupStrategy(BackupStrategy):
--- a/barman/exceptions.py
+++ b/barman/exceptions.py
@@ -263,6 +263,13 @@ class PostgresUnsupportedFeature(PostgresException):
     """
 
 
+class PostgresObsoleteFeature(PostgresException):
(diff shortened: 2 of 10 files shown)


mikewallace1979 (commented):
This needs to default to concurrent when no backup_options are set if the server version is >= 15.

We could also take the same approach used with `backup_method = postgres` and replace `backup_options = exclusive` with `backup_options = concurrent` when server version >=15.

---

Edit: This was a nice idea but adding server-version dependent checks at the config validation stage is problematic because it moves the point at which barman must make a postgres connection closer to the beginning of execution. This has a side effect of requiring Barman to connect to the PostgreSQL server for nearly every operation, whether or not the backup_method has any relevance at all. Given the amount of refactoring which would be required to implement this cleanly, we will instead omit it for now and instead require that users set `backup_options = concurrent_backup` in the global/server options when upgrading to PG15.

mikewallace1979:
Successful test run using a [barman-packaging branch](https://github.com/EnterpriseDB/barman-packaging/tree/pg15-test) to run this branch against PG15 using the [PG15 integration test branch](https://github.com/EnterpriseDB/barman-testing/pull/50): https://ci.2ndquadrant.com/jenkins/job/barman/job/barman/3728/

amenonsen (approved):

PR #580, opened 2022-05-27
Set wait_for_archive false when calling pg stop_backup

### LINKED ISSUE #448: Backups waiting to finalize on Standbys
In an specific configuration, when taking `rsync - concurrent` backups on a Standby with `archive_mode = always`, the barman backups taken on this server (specially when the load is low on the primary server) will wait with a message like: 
`Asking PostgreSQL server to finalize the backup.`

Backups would normally end when the primary server switches to a new wal file or in case it is forced with `pg_swithch_wal()`. 

From the documentation PG >= 10:

https://www.postgresql.org/docs/14/functions-admin.html#:~:text=There%20is%20an%20optional%20second%20parameter%20of%20type%20boolean.%20If%20false%2C%20the%20function%20will%20return%20immediately%20after%20the%20backup%20is%20completed%2C%20without%20waiting%20for%20WAL%20to%20be%20archived.%20This%20behavior%20is%20only%20useful%20with%20backup%20software%20that%20independently%20monitors%20WAL%20archiving

For this case, seems the code on `pg_stop_backup` could make a better use of the second option `wait_for_archive=false` in order to finalize the backup command. It can also make a better use of the settings `--wait` and `--wait-timeout`, as we believe that was the intention.

Mentioning @martinmarques and @tureba as they have been involved in some testings.

mikewallace1979:
Based on https://github.com/EnterpriseDB/barman/pull/579 because it touches the code affected by the PG15 changes so it makes sense to update that first, then change the wait_for_archive behaviour.

They should stay as two separate PRs though since they both have the potential to break things and it would be easier to figure out which change is responsible if they're separate PRs.

Closes #448.

### DIFF
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -1240,10 +1240,10 @@ class PostgreSQLConnection(PostgreSQL):
             # want to wait for the required WALs to be archived. This is because
             # barman will do its own waiting at a later stage of the backup process.
             cur.execute(
-                "SELECT location, "
-                "({pg_walfile_name_offset}(location)).*, "
+                "SELECT end_row.lsn AS location, "
+                "({pg_walfile_name_offset}(end_row.lsn)).*, "
                 "now() AS timestamp "
-                "FROM {pg_backup_stop}(TRUE, FALSE) AS location".format(**self.name_map)
+                "FROM {pg_backup_stop}(TRUE, FALSE) AS end_row".format(**self.name_map)
             )
 
             return cur.fetchone()
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -1235,11 +1235,15 @@ class PostgreSQLConnection(PostgreSQL):
             if self.server_version >= 150000:
                 raise PostgresObsoleteFeature("15")
 
+            # Call pg_stop_backup with TRUE, FALSE.
+            # TRUE indicates exclusive backup and FALSE tells PostgreSQL we do not
+            # want to wait for the required WALs to be archived. This is because
+            # barman will do its own waiting at a later stage of the backup process.
             cur.execute(
                 "SELECT location, "
                 "({pg_walfile_name_offset}(location)).*, "
                 "now() AS timestamp "
-                "FROM {pg_backup_stop}() AS location".format(**self.name_map)
+                "FROM {pg_backup_stop}(TRUE, FALSE) AS location".format(**self.name_map)
             )
 
             return cur.fetchone()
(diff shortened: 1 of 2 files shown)


mikewallace1979:
I've started a [draft PR](https://github.com/EnterpriseDB/barman/pull/597/files) for release notes and docs changes which are version-specific - I've added documentation regarding this change in the first commit.

barman/postgres.py
```diff
@@ -1235,11 +1235,15 @@ def stop_exclusive_backup(self):
...
+            # Call pg_stop_backup with TRUE, FALSE.
+            # TRUE indicates exclusive backup and FALSE tells PostgreSQL we do not
+            # want to wait for the required WALs to be archived. This is because
+            # barman will do its own waiting at a later stage of the backup process.
             cur.execute(
                 "SELECT location, "
                 "({pg_walfile_name_offset}(location)).*, "
                 "now() AS timestamp "
-                "FROM {pg_backup_stop}() AS location".format(**self.name_map)
+                "FROM {pg_backup_stop}(TRUE, FALSE) AS location".format(**self.name_map)
```
mikewallace1979, line 1246: The (bool, bool) form of pg_stop_backup returns a record: `record ( lsn pg_lsn, labelfile text, spcmapfile text )` instead of the pg_lsn which the () form returns. We therefore need to extract the pg_lsn in this query to avoid breaking exclusive backup.

mikewallace1979: Also, double check that the permissions in the docs and the permissions check are still correct since as of this commit we will no longer be using the `()` form of `pg_stop_backup`.

mikewallace1979:
Having discussed this internally I'm going to close this PR and this change is not going to ship in the next Barman release.

The reasoning is as follows:

1. The change always had sharp edges because it changes barman backup with `backup_method = rsync` such that when it completes successfully, WALs may still be on the PostgreSQL server and not yet copied onto the Barman server *unless* the `--wait` option is used in the `barman backup` command. Users could unknowingly be depending on the existing behaviour, where `barman backup` won't return until PostgreSQL has successfully run `archive_command` for all WALs required by the backup regardless of whether `--wait` is used.
2. The rationale for doing it was that it would make it consistent with `backup_method = postgres` - however _I was completely wrong about pg_basebackup_. It always makes a final WAL switch and so, if archiving is on, it will always be waiting on archiving. So when `backup_method = postgres` backups complete, all required WALs will have already been copied to the Barman server.
3. The PR is a workaround rather than a solution to the problem - the real solution to being unable to complete backups on standbys is to add a primary_conninfo config option and have Barman call pg_switch_wal on the primary when stopping the backup (thanks to Abhijit for this solution). This would fix the issue for both backup_method = postgres and backup_method = rsync.

Given the potential impact of the change, the additional inconsistency between `backup_method = rsync` and `backup_method = postgres` it would introduce, and the limited benefit, we would be better off closing this PR and instead implementing #596.

PR #588, opened 2022-06-06
Custom compression fixes

### LINKED ISSUE #566: Wrong assumption that every unidentified file is custom compressed
When using custom compression we now have custom_compression_magic that helps us detect custom compressed files. But also [this code block here](https://github.com/EnterpriseDB/barman/blob/master/barman/compression.py#L51) assumes  that every unidentified file (where magic is unknown) is custom compressed. Just this conflicts one with the other. If we know custom compression magic, then proper assumption would be that unidentified file is probably not compressed.
Here is the block for clarity:
```        
# If Barman is set to use the custom compression, it assumes that
# every unidentified file is custom compressed
if self.config.compression == "custom":
    self.unidentified_compression = self.config.compression
```

_How this manifests?_
When DB is producing WALs much faster then barman can archive them, they accumulate in the `streaming` directory; uncompressed. Compression happens when barman moves them into `wals`. If `barman-wal-restore` is used (for example from some replica or any other use-case) that fetches the WAL which is still in `streaming` directory and not yet compressed it will check the magic of that WAL, see that it is undefined, assume that this means custom compressed, try to decompress it, fail with an error and break the `barman-wal-restore`.

My guess is that assumption should just be that `self.unidentified_compression` should stay None as default. Because we have magic now which tells us how to detect custom compressed files. But I may not see the bigger picture.  :) 

In our case i had to comment out the lines above to make ZSTD work properly and not try to decompress uncompressed WALs that have not yet been archived.

### LINKED ISSUE #567: MAGIC_MAX_LENGTH not increased for custom_compression_magic
When using `custom_compression_magic` with more then 3 bytes (ZSTD has magic of 4 bytes for example) barman fails to detect the compression.

`MAGIC_MAX_LENGTH` used in `CompressionManager.identify_compression` remains 3; while `CustomCompressor` does show MAGIC of 4 and does set `sys.modules[__name__].MAGIC_MAX_LENGTH` to 4; but for some reason when calling `barman-wal-restore` which invokes `CompressionManager.identify_compression`, it sees max as 3.

Debugged this fast using additional logging calls to see what `MAGIC_MAX_LENGTH` is set in which part of the code; didn't get the chance to dive more deeper into it.
Reference here:
MAGIC_MAX_LENGTH is 3 in this code block:
https://github.com/EnterpriseDB/barman/blob/21f5e5adb105edc0b91fa158be9512bf2ebe8314/barman/compression.py#L112
MAGIC_MAX_LENGTH is 4 in this code block:
https://github.com/EnterpriseDB/barman/blob/21f5e5adb105edc0b91fa158be9512bf2ebe8314/barman/compression.py#L373

I did workaround by setting custom_compression_magic to only 3 out of its 4 bytes for ZSTD; but this is bad workaround.

This is on RHEL7; using Python 2.7.

mikewallace1979:
Fixes for #566 and #567.

See https://github.com/EnterpriseDB/barman-testing/pull/51 for integration test PR.

Closes #566.
Closes #567.

### DIFF
--- a/barman/compression.py
+++ b/barman/compression.py
@@ -46,9 +46,15 @@ class CompressionManager(object):
         self.path = path
         self.unidentified_compression = None
 
-        # If Barman is set to use the custom compression, it assumes that
-        # every unidentified file is custom compressed
-        if self.config.compression == "custom":
+        # If Barman is set to use the custom compression and no magic is
+        # configured, it assumes that every unidentified file is custom
+        # compressed. If custom_compression_magic is set then this assumption
+        # cannot be made because the custom_compression_magic should allow
+        # the compression to be identified.
+        if (
+            self.config.compression == "custom"
+            and self.config.custom_compression_magic is None
+        ):
             self.unidentified_compression = self.config.compression
 
     def check(self, compression=None):
--- a/barman/compression.py
+++ b/barman/compression.py
@@ -25,7 +25,6 @@ import bz2
 import gzip
 import logging
 import shutil
-import sys
 from abc import ABCMeta, abstractmethod
 from contextlib import closing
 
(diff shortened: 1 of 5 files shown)


mikewallace1979:
When no custom compression magic is set and Barman attempts to decompress the .partial file, instead of letting the `CommandFailedException` bubble up to the user we should be catching it and logging the error, since we know exactly what the problem is.

PR #598, opened 2022-06-14
Refactor/recovery executor

didiermichel:


barman/recovery_executor.py
```diff
@@ -1470,15 +1342,15 @@ def _backup_copy(
...
                 )
-                print(output)
+                print(cmd_output)
         base_src_path = "%s/%s" % (staging_dir, base_file)
-        output = self.compression.uncompress(
+        cmd_output = self.compression.uncompress(
             base_src_path, dest, exclude=["recovery.conf", "tablespace_map"]
         )
-        print(output)
+        print(cmd_output)
```
didiermichel, line 1353: I might need to remove those

mikewallace1979 (approved):

barman/recovery_executor.py
mikewallace1979: perhaps they should be info or debug level logs?

mikewallace1979: if it was useful at dev time chances are it might be useful at debugging-in-production time 😄

PR #599, opened 2022-06-14
Dev/555 recovery 3

### LINKED ISSUE #555: Compress base backups with pg_basebackup (backup_method = postgres)

didiermichel:
This is a Dev/555 recovery-2 rebased on master with unit test fixes

Closes #555


barman/fs.py
```diff
@@ -49,13 +49,15 @@ def get_last_output(self):
...
 
-    def create_dir_if_not_exists(self, dir_path):
+    def create_dir_if_not_exists(self, dir_path, mode=None):
         """
         This method recursively creates a directory if not exists
 
         If the path exists and is not a directory raise an exception.
 
         :param str dir_path: full path for the directory
+        :param mode str|None: Specify the mode to use for creation. Not used If teh directory already exists.
```
mikewallace1979, line 62: Minor typo `teh` in comment

barman/recovery_executor.py
```diff
@@ -1238,12 +1228,311 @@ def close(self):
...
+            dest_prefix = ":"
+
+        # Instead of adding the `data` directory and `tablespaces` to a copy
+        # controller we instead want to copy just the tarballs to a staging
+        # location via the copy controller and then untar into place.
+
+        # Create the staging area
+        staging_dir = os.path.join(
+            self.config.recovery_staging_path,
+            "barman-stating-{}-{}".format(self.config.name, backup_info.backup_id),
```
mikewallace1979, line 1276: s/stating/staging/

barman/fs.py
```diff
@@ -48,25 +52,27 @@ def get_last_output(self):
...
         _logger.debug("Create directory %s if it does not exists" % dir_path)
-        exists = self.exists(dir_path)
-        if exists:
-            is_dir = self.cmd("test", args=["-d", dir_path])
-            if is_dir != 0:
-                raise FsOperationFailed("A file with the same name already exists")
-            else:
-                return False
+        if self.check_directory_exists(dir_path):
+            False
```
mikewallace1979, line 69: Need a `return` here

barman/recovery_executor.py
```diff
@@ -1238,12 +1224,297 @@ def close(self):
...
+            exclude_args.append(name)
+        include_args = [] if include_args is None else include_args
+        args = ["xfz", src, "--directory", dst]
+        args.extend(exclude_args)
+        args.extend(include_args)
+        ret = self.command.cmd("tar", args=args)
+        out, err = self.command.get_last_output()
+        if ret != 0:
+            raise CommandFailedException(
+                "Error decompressiong %s into %s: %s" % (src, dst, err)
```
didiermichel, line 1495: typo: decompressing

barman/backup_executor.py
```diff
@@ -1588,6 +1628,40 @@ def stop_backup(self, backup_info):
             # Skip switching XLOG if a standby server
             pass
 
+    def _read_compressed_backup_label(self, backup_info):
```
mikewallace1979, line 1631: This bothers me because we must scan the tarball in order to find the backup_label - as it happens, the backup label will be one of the first members however this is only by luck and is not a documented behaviour that we should be relying on. If it turned out that backup_label was later in the tarball then Barman could be waiting a long time while we're extracting it.

didiermichel: Does it actually take more time to get the latest file in the archive than the first one using `TarFile.open`?  I was guessing it would behave like `tar -tvf file.tar` that list files to get the expected file, then extract the file. In wich case the only difference regarding file position would be the file list size (that would be max O(n) depending on the search algorithm). I admit that it could still take a while for a large backup.

mikewallace1979: Yes it is just a single scan of the tarfile, since once we have the member the subsequent extract does not require another scan - but indeed with a large enough backup that could take a long time.

Perhaps it's a case of "but a small enough overhead compared to the time taken to copy the file over the network" though.

barman/fs.py
```diff
@@ -48,25 +52,27 @@ def get_last_output(self):
...
         If the path exists and is not a directory raise an exception.
 
         :param str dir_path: full path for the directory
+        :param mode str|None: Specify the mode to use for creation. Not used
+            if the directory already exists.
+        :returns bool: False if the directory already exists True if the
+            directory is created.
         """
         _logger.debug("Create directory %s if it does not exists" % dir_path)
-        exists = self.exists(dir_path)
```
mikewallace1979, line 60: Nice - I didn't spot the duplication here with the `check_directory_exists` method.

barman/recovery_executor.py
```diff
@@ -1238,12 +1224,299 @@ def close(self):
...
+            "barman-staging-{}-{}".format(self.config.name, backup_info.backup_id),
+        )
+        output.info(
+            "Staging compressed backup files on the recovery host in: %s", staging_dir
+        )
+        recovery_info["cmd"].create_dir_if_not_exists(staging_dir, mode="700")
+        recovery_info["cmd"].validate_file_mode(staging_dir, mode="700")
+        recovery_info["staging_dir"] = staging_dir
+        self.temp_dirs.append(
+            fs.UnixCommandPathDeletionCommand(staging_dir, recovery_info["cmd"])
```
mikewallace1979, line 1283: Using `temp_dirs` to ensure the staging dir is cleaned up means it's going to get cleaned up whenever `close()` is called, so we won't keep the staging dir around in the event of failure.

Maybe that's ok for now, in which case we'd just need to update the documentation (since it was written thinking we were going to keep the staging dir around).

barman/recovery_executor.py
```diff
@@ -1238,12 +1224,299 @@ def close(self):
...
+
+        # Execute the copy
+        try:
+            controller.copy()
+        except CommandFailedException as e:
+            msg = "data transfer failure"
+            raise DataTransferFailure.from_command_error("rsync", e, msg)
+
+        # Untar the results files to their intended location
+        # When zstd and lz4 are supported this will need to be a little more
```
mikewallace1979, line 1337: We can probably remove this comment - thanks to @didiermichel 's refactoring we don't need to do anything here to support additional compressions.

barman/recovery_executor.py
```diff
@@ -1238,12 +1224,299 @@ def close(self):
...
+    """
+    if compression is None:
+        return RecoveryExecutor(backup_manager)
+    if compression == GZipCompression.name:
+        return TarballRecoveryExecutor(backup_manager, GZipCompression(command))
+
+    raise AttributeError("Unexpected compression format: %s" % compression)
+
+
+class Compression(with_metaclass(ABCMeta, object)):
```
mikewallace1979, line 1435: I keep wondering whether we could/should combine this with the `PgBaseBackupCompression` class introduced in an earlier commit. On one hand the hierarchy is similar and they both involve compression so it seems logical, but on the other hand they're dealing with different concerns so really they're different things.

tests/test_recovery_executor.py
```diff
@@ -1288,3 +1296,238 @@ def test_recover_waiting_for_wals(
...
+            executor = recovery_executor_factory(
+                mock_backup_manager, mock_command, compression
+            )
+            assert type(executor) is expected_executor
+
+
+class TestGZipCompression(object):
+    # None for src
+    # None for dst
+    # badly formed exclude / include
```
mikewallace1979, line 1428: These comments can be removed.

mikewallace1979 (changes requested):
A few minor changes and questions on my final pass through this.

doc/manual/43-backup-commands.en.md
```diff
@@ -314,6 +314,37 @@ behaviour defined by `recovery_options`. Use `--get-wal` with `barman recover`
...
+to choose a suitable location for the staging directory.
+
+If you are using the `backup_compression` option you *must* therefore
+either set `recovery_staging_path` in the global/server config *or* use
+the `--recovery-staging-path` option with the `barman recover` command. If
+you do neither of these things and attempt to recover a compressed backup
+then Barman will fail rather than try to guess a suitable location.
+
+The staging directory is only removed on successful completion of the
+recovery. This allows subsequent recovery attempts to avoid transferring
```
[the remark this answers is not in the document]
mikewallace1979: No - not any more. We should update the documentation to match the behaviour.

barman/recovery_executor.py
mikewallace1979: (I don't think this is a blocking change for this PR - just keeping the conversation regarding future refactorings alive).

barman/recovery_executor.py
didiermichel: If we want to change the behavior of removing staging dir, it can be done at low cost (I believe).
We can save staging dir clean command to another list attribute. And on close call decide to clean or not this list depending on the recovery status (not sure it is available at that point or if it is a clean and clear status).

didiermichel: But I would keep this as a future improvement.

barman/recovery_executor.py
didiermichel: I think there might be something to do at larger scale. There are indeed those to classes (recovery_executor.Compression and compression.PgBaseBackupCompression) but there is also compression.Compressor
The main issue being in my opinion those class should have pure abstract class interfaces. It might help reveal common aspects while diverging ones should be handled without forcing heritage if it is not appropriate.

mikewallace1979 (approved):

PR #611, opened 2022-06-27
Dev/release 3.0.1

didiermichel:


NEWS
```diff
@@ -1,5 +1,9 @@
 Barman News - History of user-visible changes
 
+Version 3.0.1 - 27 June 2022
+- Bug fixes:
+    - Fix package signing issue in PyPI (same sources as 3.0.0)
```
mikewallace1979, line 5: Could you add line breaks between each line so that it matches the formatting of the Version 3.0.0 release below?

didiermichel: I updated it.

PR #612, opened 2022-06-28
Allow backups against `archive_mode = always` standbys to complete

mikewallace1979:
Fully implemented for both `backup_method = rsync` and `backup_method = postgres`. Unit tests included in this PR and integration tests up for review at https://github.com/EnterpriseDB/barman-testing/pull/59.

### DIFF
--- a/barman/backup_executor.py
+++ b/barman/backup_executor.py
@@ -599,6 +599,7 @@ class PostgresBackupExecutor(BackupExecutor):
             retry_sleep=self.config.basebackup_retry_sleep,
             retry_handler=partial(self._retry_handler, dest_dirs),
             compression=self.backup_compression,
+            err_handler=self._err_handler,
         )
 
         # Do the actual copy
@@ -662,6 +663,32 @@ class PostgresBackupExecutor(BackupExecutor):
         )
         # Remove all the destination directories and reinit the backup
         self._prepare_backup_destination(dest_dirs)
+
+    def _err_handler(self, line):
+        """
+        Handler invoked during a backup when anything is sent to stderr.
+
+        Used to perform a WAL switch on a primary server if pg_basebackup
+        is running against a standby, otherwise just logs output at INFO
+        level.
+
+        :param str line: The error line to be handled.
+        """
+
+        # Always log the line, since this handler will have overridden the
+        # default command err_handler.
+        # Although this is used as a stderr handler, the pg_basebackup lines
+        # logged here are more appropriate at INFO level since they are just
+        # describing regular behaviour.
+        _logger.log(logging.INFO, "%s", line)
+        if (
+            self.server.config.primary_conninfo is not None
+            and "waiting for required WAL segments to be archived" in line
+        ):
+            # If pg_basebackup is waiting for WAL segments and primary_conninfo
+            # is configured then we are backing up a standby and must manually
+            # perform a WAL switch.
+            self.server.postgres.switch_wal()
 
     def _prepare_backup_destination(self, dest_dirs):
         """
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -23,6 +23,12 @@ This module represents the interface towards a PostgreSQL server.
 import atexit
 import logging
 from abc import ABCMeta
+from multiprocessing import Process, Queue
+
+try:
+    from queue import Empty
+except ImportError:
+    from Queue import Empty
 
 import psycopg2
 from psycopg2.errorcodes import DUPLICATE_OBJECT, OBJECT_IN_USE, UNDEFINED_OBJECT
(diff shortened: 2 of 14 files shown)


mikewallace1979 (commented):
Discussed with @amenonsen offline and the following points came up:

1. We need to verify that `primary_conninfo` is pointing to a primary server (i.e. the server is in recovery).
2. We need to verify that `conninfo` is pointing to a standby server.
3. We should also verify that `primary_conninfo` and `conninfo` point to servers with the same system ID.

Once we've done these things we can be reasonably certain that `primary_conninfo` and `conninfo` are pointing to where they should be and that the WAL switch will happen on the correct server.

barman/server.py
```diff
@@ -265,9 +269,20 @@ def __init__(self, config):
...
+                    self.postgres = StandbyPostgreSQLConnection(
+                        config.conninfo,
+                        config.immediate_checkpoint,
+                        config.slot_name,
+                        # The standby needs a reference to the primary so that it can
+                        # perform WAL switches itself when calling pg_backup_stop.
+                        primary_conninfo=config.primary_conninfo,
+                    )
+                else:
+                    self.postgres = PostgreSQLConnection(
```
mikewallace1979, line 283: If a user has not set `primary_conninfo` it is still possible that `PostgreSQLConnection` is a standby server - in this case Barman will do nothing special about it, it will attempt to switch WALs on the server at the usual points and ignore the `PostgresInRecovery` exception from psycopg2 just like it does now.

The user will therefore see the same behaviour as with the current Barman, unless they set `primary_conninfo`.

barman/backup_executor.py
```diff
@@ -648,6 +649,34 @@ def _retry_handler(self, dest_dirs, command, args, kwargs, attempt, exc):
...
+            # Always log the line, since this handler will have overridden the
+            # default command err_handler.
+            # Although this is used as a stderr handler, the pg_basebackup lines
+            # logged here are more appropriate at INFO level since they are just
+            # describing regular behaviour.
+            _logger.log(logging.INFO, "%s", line)
+            # TODO: Does this line change across pg_basebackup versions 10 to 15?
+            if (
+                self.server.config.primary_conninfo is not None
+                and "waiting for required WAL segments to be archived" in line
```
mikewallace1979, line 671: Verified this string is consistent across PostgreSQL versions 10 to 15(beta)

doc/barman.5.d/50-primary_conninfo.md
```diff
@@ -0,0 +1,18 @@
+primary_conninfo
+:   The connection string used by Barman to connect to the primary Postgres
+    server during backup of a standby Postgres server. Barman will use this
+    connection to carry out any required WAL switches on the primary during
+    the backup of the standby. This allows backups to complete even when
+    `archive_mode = always` is set on the standby and write traffic to the
+    primary is not sufficient to trigger a natural WAL switch.
+
+    If this is set then it *must* be pointing to a primary Postgres instance
```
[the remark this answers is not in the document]
mikewallace1979: Thanks for catching the ambiguity - will fix.

barman/backup_executor.py
```diff
@@ -658,6 +659,33 @@ def _retry_handler(self, dest_dirs, command, args, kwargs, attempt, exc):
...
+    def _make_err_handler(self):
+        """
+        Handler invoked during a backup when anything is sent to stderr.
+
+        Used to perform a WAL switch on a primary server if pg_basebackup
+        is running against a standby, otherwise just logs output at INFO
+        level.
+        """
+
+        def err_handler(line):
```
didiermichel, line 671: I am not a huge fan of functions defined within functions. 
I am also wondering if the self within `err_handler` refers to `PostgresBackupExecutor` instance (in wich case the function is like a disguised object).
But I can see this is heavily used already. I think a more appropriate solution would be to use an interface with handle_error function. and passing anything that implements this interface the the handler would be using this function.
It might not be worth refactoring this behavior, and maybe this solution has limitations. We can probably live with that until we can't :)

barman/postgres.py
```diff
@@ -1720,3 +1726,155 @@ def name_map(self):
...
+        super(StandbyPostgreSQLConnection, self).__init__(
+            conninfo,
+            immediate_checkpoint=immediate_checkpoint,
+            slot_name=slot_name,
+            application_name=application_name,
+        )
+        self.primary_conninfo = primary_conninfo
+        # The standby connection has its own connection object used to talk to the
+        # primary when switching WALs.
+        self.primary = PostgreSQLConnection(
```
didiermichel, line 1771: Maybe Instanciate the `PostgreSQLConection instance` in server.Server._init_postgres() and give this instance when initiating `StandbyPostgreSQLConnection ` at the same place. This way both objects are created at the same place and less coupled.

doc/manual/50-feature-details.en.md
```diff
@@ -238,20 +238,35 @@ rules:
...
+> **NOTE:** It is especially important that `primary_conninfo` is
+> set if the standby is to be backed up when there is little or no write
+> traffic on the primary. If `primary_conninfo` is not set then the
+> backup will still run however it will wait at the stop backup stage
+> until the current WAL semgent on the primary is newer than the latest
+> WAL required by the backup.
+
+Barman currently requires that WAL files and backup data come from the
+same PostgreSQL server. In the case that the standby is promoted to primary
+the backups and WALs will continue to be valid however you may wish to update
```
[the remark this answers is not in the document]
mikewallace1979: Yes we'd need to either remove `primary_conninfo` or point it to the new primary. If nothing is updated after the standby is promoted then `barman check` is going to fail because of [Server._check_standby](https://github.com/EnterpriseDB/barman/pull/612/files#diff-40d894681de880b0b65b0db09b8c805d98a794a302d7fdcdb82e205f3952a12fR861) so backups aren't going to run until it is updated.

barman/backup_executor.py
mikewallace1979: On second thoughts, I don't think `partial` is going to let me assign the implicit `self`.

barman/postgres.py
mikewallace1979: What I had intended to achieve here was making the `primary` reference an implementation detail of `StandbyPostgreSQLConnection` - the idea being that nothing which interacts with `StandbyPostgreSQLConnection` should have to care that it is using another PostgreSQL connection to the primary. However given this isn't actually the case (Server is already referencing `primary` directly) it would be better to create is in Server and pass it in.

didiermichel (approved):

PR #614, opened 2022-06-30
Dev/rework compression classes

didiermichel:
This pull request does a bit of refactoring. Its main purpose is to have a common compression base between pg_basebackup compression and recovery

Here is a list of main changes:
- Around pg_basebackup compression classes
- Server_config
- use tmpdir in dummy_dest file creation (test_barman_wal_restore)

### DIFF
--- a/barman/backup_executor.py
+++ b/barman/backup_executor.py
@@ -32,7 +32,6 @@ import logging
 import os
 import re
 import shutil
-from tarfile import TarFile
 from abc import ABCMeta, abstractmethod
 from functools import partial
 
@@ -52,6 +51,7 @@ from barman.exceptions import (
     PostgresConnectionError,
     PostgresIsInRecovery,
     SshCommandException,
+    FileNotFoundException,
 )
 from barman.fs import UnixLocalCommand, UnixRemoteCommand
 from barman.infofile import BackupInfo
@@ -269,9 +269,8 @@ class PostgresBackupExecutor(BackupExecutor):
         # Forbid tablespace_bandwidth_limit option.
         # It works only with rsync based backups.
         if self.config.tablespace_bandwidth_limit:
-            self.server.config.disabled = True
             # Report the error in the configuration errors message list
-            self.server.config.msg_list.append(
+            self.server.config.update_msg_list_and_disable_server(
                 "tablespace_bandwidth_limit option is not supported by "
                 "postgres backup_method"
             )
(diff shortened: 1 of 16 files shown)


barman/backup_executor.py
```diff
@@ -1634,13 +1647,10 @@ def _read_compressed_backup_label(self, backup_info):
...
         :param barman.infofile.LocalBackupInfo backup_info: backup information
         """
-        basename = os.path.join(backup_info.get_data_directory(), "base.tar")
-        with self.backup_compression.open(basename) as f:
-            self.tar = TarFile.open(fileobj=f, mode="r|")
-            for member in self.tar:
-                if member.name == "backup_label":
-                    backup_label_fileobj = self.tar.extractfile(member)
-                    return backup_label_fileobj.read().decode("utf-8")
+        basename = os.path.join(backup_info.get_data_directory(), "base")
```
didiermichel, line 1650: The idea of giving the archive name without the extension feels weird.  especially if we need to get the extension later in case of error (using with_suffix). it seems there is a contradiction here.
We could either discover the full name of the archive and give it to the `get_file_content`. then we do not need with_suffix method later as we would have the full archive name.
Or we could  let the `backup_compression` find the archive full name as it is actually done. Then rely on the `FileNotFoundException` to forge the new message and we would not need the actual `with_suffix` method.

barman/compression.py
```diff
@@ -447,53 +597,200 @@ def with_suffix(self, basename):
...
+        ret = self.command.cmd("tar", args=args)
+        out, err = self.command.get_last_output()
+        if ret != 0:
+            raise CommandFailedException(
+                "Error decompressing %s into %s: %s" % (src, dst, err)
+            )
+        else:
+            return self.command.get_last_output()
+
+    def get_file_content(self, filename, archive):
```
mikewallace1979, line 752: Does this specific function get called anywhere? I'm having trouble finding a reference to it being called on `GZipCompression` (other than in the tests) and I'm wondering whether it is used at all, since by the end of the #555 PR we had switched to copying the config files from the recovery target directory rather than extracting them from the staged tarball.

mikewallace1979 (commented):
I had one minor in-line question about a function which I'm not sure is still being used but generally this looks like an improvement on what we had before. It seems pretty clear where we would add in additional compression.

barman/compression.py
mikewallace1979: Thinking about this some more, perhaps we would be better off using this `get_file_content` method to read the backup_label, since we would then be relying only on external tools to handle compression-specific concerns rather than python libraries *and* external tools. Not something which needs action now but when we do add zstd and lz4 it might be a worthwhile change.

barman/compression.py
```diff
@@ -447,53 +599,205 @@ def with_suffix(self, basename):
...
+        """
+
+        :param filename: str file to search for in the archive (requires its full path within the archive)
+        :param archive: str archive path/name without extension
+        :return: string content
+        """
+        full_archive_name = "%s.%s" % (archive, self.file_extension)
+        args = ["-xzf", full_archive_name, "-O", filename]
```
mikewallace1979, line 762: Is it worth using the `--occurrence` option here so that tar exits as soon as it finds the file rather than continuing to scan the tarball?

One downside is that it would break on systems which don't have GNU tar.

barman/backup_executor.py
```diff
@@ -269,56 +269,68 @@ def validate_configuration(self):
...
+            self._validate_compression(remote_status)
+
+    def _validate_compression(self, remote_status):
+        """
+        In charge of validating compression options
+        :param remote_status:
+        :return:
+        """
+        # Todo: config server change was tested within backup_compression.validate function.
+        #       It is not tested anymore. Probably worth testing this method.
```
mikewallace1979, line 328: Is it still worth adding a test for this?

barman/compression.py
```diff
@@ -447,53 +597,200 @@ def with_suffix(self, basename):
...
+        return self.options.validate(pg_server_version, remote_status)
 
-        # plain backup format is only allowed when compression is on the server
-        if self.format == "plain" and self.location != "server":
-            server.config.disabled = True
-            server.config.msg_list.append(
-                "backup_compression_format plain is not compatible with "
-                "backup_compression_location %s" % self.location
+
+class TarCompressor(object):
```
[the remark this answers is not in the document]
mikewallace1979: If we're going to handle compressed tarballs with external invocations of tar then we can probably just remove TarCompressor entirely.

barman/compression.py
mikewallace1979: Considering Barman only supports Gnu/Linux systems and not BSD I think I'm ok with using `--occurrence` here for the sake of avoiding a complete scan of the compressed tarball.

barman/compression.py
```diff
@@ -472,7 +473,8 @@ def get_pg_basebackup_compression(server):
         base_backup_compression_option = GZipPgBaseBackupCompressionOption(
             pg_base_backup_cfg
         )
-        compression = InternalTarGZipCompression()
+        # Use GZipCompression instead of InternalTarGZipCompression to avoid using external python libraries
+        compression = GZipCompression(unix_command_factory())
         return PgBaseBackupCompression(
```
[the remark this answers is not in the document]
mikewallace1979: Having tested this both manually using a few GB of pseudorandom data and via the integration tests I am confident this works as required so I'm :+1: on removing `InternalTarGZipCompression`.

mikewallace1979 (changes requested):
Looks pretty close to mergeable now and much cleaner than the pre-PR code 😃.

Left a few minor comments regarding code which I think can just be removed now (unless I missed something).

mikewallace1979 (approved):

PR #616, opened 2022-06-30
Improve docs for backing up standby servers

### LINKED ISSUE #615: Fix broken docs around the "Current limitations on backup from standby" section
The documentation here is outdated in several places and refers to a pgsql-bugs post which only affects PG10 and earlier. As a minimum that statement needs to be clarified (we should also respond to the psql-bugs post to close the loop there).

There are also other places this section needs improving:

> Barman currently requires that backup data (base backups and WAL files) come from one server only. Therefore, in case of backup from a standby, you should point to the standby server:

This statement confuses a lot of people into thinking backups can only come from that standby and that, once the standby is promoted, the backups and WALs will not be valid for the new primary. This is not the case - the statement is really just saying that base backups and WALs should come from the same server (be that the standby or the primary).

The references to PostgreSQL versions earlier than 10 can be removed altogether since they are no longer supported.

The parent section, `Concurrent Backup and backup from a standby` also has similar issues with references to unsupported PostgreSQL versions so we should resolve those (and make any other possible improvements) while we're there.

mikewallace1979:
Attempts to improve the documentation regarding the concurrent backup
of standby servers by:

  1. Removing references to unsupported PostgreSQL versions (and
     consequently PgEspresso).
  2. Clarifying that backups and WALs must come from the same server but
     this doesn't mean backups taken against a standby are no longer
     valid once it has been promoted to primary.
  3. Avoid recommending WAL streaming over archiving and instead refer
     to the relevant doc sections.
  4. Rewrite the warning about not streaming and archiving WALs from
     a standby - this only applied to PostgreSQL 10 and earlier.

Closes #615.

### DIFF
--- a/doc/manual/50-feature-details.en.md
+++ b/doc/manual/50-feature-details.en.md
@@ -198,18 +198,9 @@ As well as being the recommended backup approach, concurrent backup
 also allows the following architecture scenario with Barman: **backup from
 a standby server**, using `rsync`.
 
-> **IMPORTANT:** **Concurrent backup** requires users of PostgreSQL
-> 9.2, 9.3, 9.4, and 9.5 to install the `pgespresso` open source
-> extension on every PostgreSQL server of the cluster. For more
-> detailed information and the source code, please visit the
-> [pgespresso extension website][9].  Barman supports the new API
-> introduced in PostgreSQL 9.6. This removes the requirement of the
-> `pgespresso` extension to perform concurrent backups from this
-> version of PostgreSQL.
-
-By default, `backup_options` is transparently set to `concurrent_backup`.
-If exclusive backup is required for PostgreSQL servers older than version
-15 then users should set `backup_options` to `exclusive_backup`.
+By default, `backup_options` is set to `concurrent_backup`. If exclusive
+backup is required for PostgreSQL servers older than version 15 then users
+should set `backup_options` to `exclusive_backup`.
 
 When `backup_options` is set to `concurrent_backup`, Barman activates
 the _concurrent backup mode_ for a server and follows these two simple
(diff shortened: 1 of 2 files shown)


doc/manual/50-feature-details.en.md
```diff
@@ -237,33 +224,32 @@ rules:
...
+
+WALs can be obtained from the standby using either WAL streaming or WAL
+archiving. To use WAL streaming follow the instructions in the
+[WAL streaming](#wal-streaming) section.
+
+To use WAL archiving from the standby follow the instructions in the [WAL archiving via archive_command](#wal-archiving-via-archive_command) section
+*and additionally* set `archive_mode = always` in the PostgreSQL config on
+the standby server.
+
+> **NOTE:** With PostgreSQL 10 and earlier Barman cannot handle WAL streaming
```
mikewallace1979, line 248: It's worth adding the full context to the commit message in the unlikely event that someone wants more information: Basically the issue was fixed as a side-effect of https://github.com/postgres/postgres/commit/4a33bb59dfc33566f04e18ab5e1f90b8e7461052 which landed in PG11 - more details can be found at https://gist.github.com/mikewallace1979/3b52cd87faf9a302fefc72377e35fafb.

mxey:
> Clarifying that backups and WALs must come from the same server but
this doesn't mean backups taken against a standby are no longer
valid once it has been promoted to primary.

Can you add an explanation why that is necessary?

doc/manual/50-feature-details.en.md
```diff
@@ -237,33 +224,32 @@ rules:
 
 #### Current limitations on backup from standby
 
-Barman currently requires that backup data (base backups and WAL files)
-come from one server only. Therefore, in case of backup from a
-standby, you should point to the standby server:
+Barman currently requires that WAL files and backup data come from the
+same PostgreSQL server. When taking backups from a standby it is therefore
+important to point the following Barman [configuration options][config-options]
```
didiermichel, line 229: Is that a way to crate an link?

didiermichel (approved):

mikewallace1979:
Filed #631 for investigating whether Barman will allow backups and WALs which come from different servers (e.g. WALs  from the primary, backups from the standby) - current indications are that it could but it doesn't, so we may need to update some of the barman check logic.

PR #628, opened 2022-07-26
Fix epoch timestamps in json output

### LINKED ISSUE #618: list-backups JSON output has incorrect end_time_timestamp
The field `end_time_timestamp` is incorrect by an hour offset.

```
# barman -f json list-backups provid_patroni_test | jq '.provid_patroni_test[0]'
{
  "backup_id": "20220627T000011",
  "end_time": "Mon Jun 27 00:00:18 2022",
  "end_time_timestamp": "1656284418",
  "retention_status": "-",
  "size": "462.8 MiB",
  "size_bytes": 485329527,
  "status": "DONE",
  "tablespaces": [],
  "wal_size": "0 B",
  "wal_size_bytes": 0
}
```

Render the timestamp:

```
# date -d '@1656284418'
Mon 27 Jun 2022 01:00:18 AM CEST
```

Which clearly is not the same time as `Mon Jun 27 00:00:18 2022` in the `end_time` JSON field, or `end_time=2022-06-27 00:00:18.286025+02:00` in backup.info

The culrpit is using `.strftime("%s")` in output.py to create the timestamp. The output is correct once you replace the callsites with `.timestamp()`, see https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp.

mikewallace1979:
Closes #618 by fixing the epoch timestamps in the JSON output to account for timezones.

### DIFF
--- a/barman/output.py
+++ b/barman/output.py
@@ -35,6 +35,7 @@ from barman.utils import (
     human_readable_timedelta,
     pretty_size,
     redact_passwords,
+    timestamp,
 )
 from barman.xlog import diff_lsn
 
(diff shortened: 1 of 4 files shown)


mikewallace1979:
The test coverage metric is lower than usual here because sonarqube counts the python2-compatibility lines as uncovered whereas they are in fact covered when tox runs the python2 tests.

didiermichel (approved):
All seems fine to me

Issue #393, opened 2021-09-09
Recovery does not include partial wal file unless --get-wal is specified which implies that SSH is enabled


allanharry:
Hi
Due to security we dont want that database team can gain acces to backups in BARMAN using SSH.
So we are using streaming protocol for wal files and pgbasebackup where ssh is not needed.

**For the recover part we observe that transactions in partial wal file not being applied !**
barman recover \
--remote-ssh-command "sshpass -p xxxxxx ssh postgres@pg1" \
pg1 \
20210124T232937 \
/pg_database/pg12

**Doing the recover with --get-wal and allowing ssh from pg1 to barman server will include partial wal file and RPO=zero:**
barman recover \
--get-wal \
--remote-ssh-command "sshpass -p xxxxxx ssh postgres@pg1" \
pg1 \
20210124T232937 \
/pg_database/pg12

If backup can be done using streaming protocol then this partial wal file should also somehow be handled during the recover to ensure RPO=zero.
Are there any plans to include this in BARMAN ?

Regards Allan

nlgtuankiet:
I can confirm that the issue exists.
@mikewallace1979 is there any update on this?

mikewallace1979:
@nlgtuankiet Thanks for confirming - I have also reproduced this. We'll get a fix into one of the upcoming releases.

rprovodenko:
This issue has been open for 3 years now. Any workaround at all please?

martinmarques:
The workaround is to manually copy the partial WAL into the `barman_wal` directory. I am triaging this ticket, but the ask from the OP was to provide streaming from the barman server to the postgres server. That seems like something to implement in postgres itself.

martinmarques:
IMO the solution would be for barman to recognize that there is a partial WAL (maybe we should also have a flag in `barman recover` asking for the partial file, in case someone would prefer it not to be used) and copy it together with the WAL files during the `barman recover` execution.

DariuszGarbarz:
Hi @martinmarques, I have a questions regarding "manual" workaround. Is it enough to remove .partial suffix and just copy the file to the wal folder? How safe is that? What happen if partial wal file is somehow corrupted or transaction at the end of partial wal file is uncompleted or was not properly saved due to postgresql crash etc?

martinmarques:
We won't be doing any development regarding this, as we see more dangers than anything helpful. We will be documenting the workaround described earlier here

PR #629, opened 2022-07-27
[wip] Proof-of-concept fix for missing .partial WALs on recovery

mikewallace1979:
Updated with a different approach where `.partial` files are handled after copying the WALs in the archive.

There are a few upsides to doing this:

* We no longer need special cases in the code branch for compressed WALs.
* We can use `Rsync.from_file_list` to copy the .partial files when WALs are uncompressed, since all `.partial` files now get copied and renamed in a temporary staging directory.

Including the `.partial` WAL files in the `get_required_xlog_files` response *seems* a bit odd since one of the first things we do in `_xlog_copy` is put them in a separate list so they can be handled after the WALs in the archive are copied, however it feels like the right thing to do given that the `.partial` file *is* a required xlog file (unless the requested recovery timeline or time dictate otherwise).

### DIFF
--- a/barman/recovery_executor.py
+++ b/barman/recovery_executor.py
@@ -785,9 +785,27 @@ class RecoveryExecutor(object):
         compression_manager = self.backup_manager.compression_manager
         # Fill xlogs and compressors maps from required_xlog_files
         for wal_info in required_xlog_files:
-            if xlog.is_partial_file(wal_info.name):
+            if (
+                xlog.is_partial_file(wal_info.name)
+                and hasattr(wal_info, "orig_filename")
+                and (
+                    os.path.dirname(wal_info.orig_filename)
+                    == self.config.streaming_wals_directory
+                )
+            ):
+                # If a .partial file is in the `streaming` directory then add it to a
+                # separate list so it can be renamed and transferred in a separate
+                # Rsync operation.
                 partial_xlogs.append(wal_info)
             else:
+                if xlog.is_partial_file(wal_info.name):
+                    # If a .partial file is in the main WAL archive then either a
+                    # standby was promoted or a primary was recovered from a backup.
+                    # In either case we do not attempt to handle it and instead warn
+                    # the user.
+                    output.warning(
+                        ".partial WAL file '%s' found in WAL archive", wal_info.name
+                    )
                 hashdir = xlog.hash_dir(wal_info.name)
                 xlogs[hashdir].append(wal_info)
                 # If a compressor is required, make sure it exists in the cache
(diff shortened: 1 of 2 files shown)


martinmarques:
@mikewallace1979 do you think this is something we would like to keep pursuing?

barman/recovery_executor.py
```diff
@@ -782,16 +785,39 @@ def _xlog_copy(self, required_xlog_files, wal_dest, remote_command):
...
+                # separate list so it can be renamed and transferred in a separate
+                # Rsync operation.
+                partial_xlogs.append(wal_info)
+            else:
+                if xlog.is_partial_file(wal_info.name):
+                    # If a .partial file is in the main WAL archive then either a
+                    # standby was promoted or a primary was recovered from a backup.
+                    # In either case we do not attempt to handle it and instead warn
+                    # the user.
+                    output.warning(
```
gcalacoci, line 806: I think we should make this warning more informative, @barthisrael @martinmarques .
What I mean is, we can give more info to the user on what to do with the partial, and why is there.
The comment from Mike right above can be translated into the warning imho

barman/recovery_executor.py
```diff
@@ -882,6 +891,30 @@ def _xlog_copy(self, required_xlog_files, wal_dest, remote_command):
                     )
                     raise DataTransferFailure.from_command_error("rsync", e, msg)
 
+        # Now copy any .partial files we need to care about
+        if partial_xlogs:
+            _logger.info("Copying .partial WAL files from streaming directory.")
```
gcalacoci, line 896: looks like i'm the logging police today but... i would change this to clearly state where the partials are going to be put, something like: 
```
Copying .partial files from streaming directory into %s
```
where %s is the destination path

barman/recovery_executor.py
```diff
@@ -882,6 +891,30 @@ def _xlog_copy(self, required_xlog_files, wal_dest, remote_command):
...
+            partial_staging_dir = tempfile.mkdtemp(prefix="barman_wal-partial-")
+            # Stage a copy of the .partial file with its final name so that Rsync can
+            # copy the file to its final destination.
+            for segment in partial_xlogs:
```
gcalacoci, line 900: also, let's add a debug message (at least) about renaming

barthisrael:
This draft PR was an attempt to address the issue https://github.com/EnterpriseDB/barman/issues/393. However, it's old at this point and needs further analysis.

This branch has been pushed to our internal repository for future analysis.
We are closing this PR in the main repo for now.

PR #630, opened 2022-07-27
Rename backup_manifest files on recovery

### LINKED ISSUE #623: Restored backup_mainfest files in recovered PGDATA prevents generation of backup_manifest for subsequent backups
`barman recover` will copy the `backup_manifest` file into the PGDATA recovery location during the restore process and leave the file in place. When the next backup runs it will include that `backup_manifest` file however the file will not be consistent with the new backup and therefore attempts to verify the new backup against that manifest will fail. The presence of the file also makes manual attempts to generate the manifest fail since the file already exists.

Possible solutions here are:

1. Exclude the `backup_manifest` during the recovery process so that it is never copied into the target PGDATA directory.
2. Rename the file, perhaps with a timestamp, so that the manifest filename does not collide with the backup manifest file for future backups.

The advantage of option 2 is that the backup manifest for the recovered backup would still be available so if a user wanted to verify the recovered backup they could do so.

Note: When we recover from compressed backups we do not recover the backup_manifest file. This is through an implementation detail (the backup_manifest is not included in the compressed tarball it is a separate file) rather than by design. We should therefore make make the recovery of the backup_manifest consistent with recovery from non-compressed backups as part of this work.

### LINKED ISSUE #626: Fix `barman generate-manifest` logs and exit status when backup manifest already exists
If `barman generate-manifest` runs and the backup manifest file already exists it will log at `INFO` level that:

1. The backup manifest exists.
2. The backup is valid.

However the fact that the backup manifest exists does not necessarily mean everything is ok - it is only ok of the backup_manifest is the correct manifest file for that backup. This may not be the case (for example in #623 it is possible to end up with backup manifests for old backups in the backup directory) so this should either an error or a warning condition and the exit status should be non zero.

The message about the backup being valid is misleading because the existence of the backup_manifest file does not guarantee the backup is valid. This should probably just be removed.

mikewallace1979:
Closes #623.
Closes #626.

See https://github.com/EnterpriseDB/barman-testing/pull/60 for related integration tests.

### DIFF
--- a/barman/cli.py
+++ b/barman/cli.py
@@ -1524,7 +1524,10 @@ def generate_manifest(args):
     )
     backup_manifest.create_backup_manifest()
 
-    output.info("Backup %s is valid on server %s" % (args.backup_id, args.server_name))
+    output.info(
+        "Backup manifest for backup %s successfully generated for server %s"
+        % (args.backup_id, args.server_name)
+    )
     output.close_and_exit()
 
 
--- a/barman/backup_manifest.py
+++ b/barman/backup_manifest.py
@@ -16,11 +16,10 @@
 # You should have received a copy of the GNU General Public License
 # along with Barman.  If not, see <http://www.gnu.org/licenses/>.
 
-import logging
 import os
 import json
 
-from barman import output
+from barman.exceptions import BackupManifestException
 
 
 class BackupManifest:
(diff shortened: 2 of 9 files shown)


barman/cli.py
```diff
@@ -1524,7 +1524,10 @@ def generate_manifest(args):
     )
     backup_manifest.create_backup_manifest()
 
-    output.info("Backup %s is valid on server %s" % (args.backup_id, args.server_name))
+    output.info(
+        "Backup manifest for backup %s successfully generated for server %s"
```
didiermichel, line 1528: +1 User friendly message

didiermichel (approved):

PR #634, opened 2022-08-02
Allow receive-wal options against inactive servers

### LINKED ISSUE #574: Allow `barman receive-wal` to run against inactive servers when using the `--stop`, `--reset`, `--drop-slot` and `--create-slot` options
Certain DBA procedures can require WAL streaming to be temporarily disabled and re-enabled at a later date. In order to achieve this it is necessary to:

1. Stop the pg_receivewal process with `barman receive-wal <server> --stop`.
2. Drop the replication slot with `barman receive-wal <server> --drop-slot`.
3. Set the server to inactive in the barman config.
4. At a later date, set the server to active in the barman config.
5. Recreate the replication slot with `barman receive-wal <server> --create-slot`
6. Reset WAL streaming, if the WAL with the position Barman was last streaming from is no longer available: `barman receive-wal <server> --reset`.

The problem is that a Barman installation has a cron job which runs `barman cron` every 60 seconds - for an active server it will always try and start the pg_receivewal process. This means that:

* Between steps 1 and 2 there is a chance the cron job can restart the pg_receivewal process.
* Between steps 4 and 6 the cron job will keep trying to start the pg_receivewal process and spam the logs.

The process *could* be improved by setting the server to inactive before running `barman receive-wal <sever> --stop` and not setting it to active until after the `barman receive-wal <server> --reset` step *except* the command will fail if the server is set to inactive in the configuration.

So Barman could be enhanced here such that it allows `barman receive-wal <server>` commands to be run with the `--stop`, `--reset`, `--drop-slot` and `--create-slot` options.

If we're not comfortable always allowing it against an inactive server we could add a `--force` option which allows the user to override the inactive server check.

mikewallace1979:
Allows the receive-wal options --create-slot, --drop-slot, --reset and
--stop to be used against inactive servers. If `barman receive-wal` is
called with no options in order to start WAL streaming then inactive
servers will continue to be skipped.

The rationale for this is that certain DBA procedures can require that
WAL streaming is temporarily disabled and re-enabled at a later date.
Without this patch the server must be active in order to stop WAL
streaming, reset the stream or create/drop the replication slot. While
the server is active the `barman cron` job will restart WAL streaming
if it runs. Such procedures are therefore inherently racy unless the
`barman cron` job is disabled.

By allowing the receive-wal options to run against inactive servers,
users are now able to mark the server as `active = false` in the Barman
configuration and then carry out the necessary `receive-wal` actions
without needing to worry about `barman cron` restarting WAL streaming.

Closes #574.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -38,6 +38,7 @@ from barman.cli import (
     get_server_list,
     manage_server_command,
     OrderedHelpFormatter,
+    receive_wal,
     recover,
     keep,
     show_servers,
@@ -119,6 +120,27 @@ class TestCli(object):
         out, err = capsys.readouterr()
         assert err
         assert "ERROR: Conflicting path:" in err
+
+    def test_get_server_inactive(self, monkeypatch):
+        """
+        Test that get_server correctly handles inactive servers.
+        """
+        # GIVEN an inactive server
+        args = Mock()
+        monkeypatch.setattr(
+            barman, "__config__", build_config_from_dicts(main_conf={"active": "false"})
+        )
+
+        # WHEN get_server is called with skip_inactive=True
+        # THEN a SystemExit is raised
+        args.server_name = "main"
+        with pytest.raises(SystemExit):
+            get_server(args, skip_inactive=True)
+
+        # AND WHEN get_server is called with skip_inactive=False
+        # THEN a server is returned
+        args.server_name = "main"
+        assert get_server(args, skip_inactive=False) is not None
 
     def test_manage_server_command(self, monkeypatch, capsys):
         """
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -1030,3 +1052,61 @@ def test_show_servers_json(
         json_output = json.loads(out)
         assert [self.test_server_name] == list(json_output.keys())
         assert json_output[self.test_server_name]["description"] == expected_description
+
+
+class TestReceiveWalCli(object):
```
didiermichel, line 1057: I am wondering about the need to have a dedicated class for this test. Isn't `TestCli` suitable?

mikewallace1979: It does look a bit unnecessary doesn't it - I'll move it.

didiermichel (approved):

PR #643, opened 2022-08-05
Dev/636 add lz4 to compressed backup with pg basebackup

didiermichel:


barman/compression.py
```diff
@@ -565,6 +573,36 @@ def validate(self, pg_server_version, remote_status):
...
+        )
+        # "backup_location = server" requires pg_basebackup >= 15
+        if remote_status["pg_basebackup_version"] < Version("15"):
+            issues.append(
+                "backup_compression = %s requires "
+                "pg_basebackup 15 or greater" % self.config.type
+            )
+
+        if self.config.level is not None and (
+            self.config.level < 0 or self.config.level > 9
```
mikewallace1979, line 597: I think the range for the `lz4` cli is 1 to 12.

didiermichel: I have found several limits according to where I search.

lz4 -H on ubuntu sais:
```shell
Compression levels : 
There are technically 2 accessible compression levels.
-0 ... -2 => Fast compression
-3 ... -9 => High compression
```

another help [page](https://www.mankier.com/1/lz4) does say 1 to 12

Maybe I did not instal the proper binary or there is missing information on the help ?

mikewallace1979: Ah interesting. I tried the liblz4-tool package on ubuntu 18.04 which lists levels 0 to 9 though if I use a higher value such as `-12` it still works (though the file was not compressed any more than with `-9`).

With ubuntu 20.04 the lz4 command installed by the `lz4` package implies `-12`.

Maybe we should just validate it's non-negative and leave it at that?

didiermichel: I tried with a high value and got  following mesage:
```
pg_basebackup: error: invalid compression specification: compression algorithm "lz4" expects a compression level between 1 and 12
```
I guess that settles it :)

barman/command_wrappers.py
```diff
@@ -1200,6 +1200,8 @@ def shell_quote(arg):
     # a backslash, and then start another string using a quote character.
 
     assert arg is not None
+    if arg == "|":
+        return arg
```
mikewallace1979, line 1204: I am probably missing something but is this change related to something else which changed in this PR?

didiermichel: Well the change is actually not used. 
At first instead of using `tar --use-compress-program lz4 ...` I tried to use `lz4 -d file.tar.lz4 | tar -xf` 
I realized that `|` as argument was `"|"`  and it created an error at execution time.

I left it there, just in case we would need it at some point.
Do you think it should be removed?

mikewallace1979: I think it should stay - I just felt incomplete for not knowing the story behind it :)

mikewallace1979 (approved):
This looks good to me providing the integration tests are happy.

Local testing with 1GB of not-very-compressible data yields backup times around 5x faster with lz4 vs gzip, and recovery times around 2x faster which is pretty awesome I think.

Issue #636, opened 2022-08-03
Server-side base backups (lz4 and zstd) with pg_basebackup (backup_method = postgres)


didiermichel:
Now that barman can leverage compress base backups with pg_basebackup, it is time to add lz4 and zstd support.
Those compression algorithm will be available for PG >= 15 

Details for lz4:
 Allow --compress server-lz4:level=LEVEL
 with   `1<= LEVEL <=12` 

Details for zstd: 
Allow --compress server-zstd:level=LEVEL,workers=WORKERS
 with `0 <= LEVEL <= 22`
 and `WORKERS >= 0`

PR #644, opened 2022-08-10
add zstd compression algorithm support (#636)

didiermichel:


barman/compression.py
```diff
@@ -603,6 +611,36 @@ def validate(self, pg_server_version, remote_status):
...
+        )
+        # "backup_location = server" requires pg_basebackup >= 15
+        if remote_status["pg_basebackup_version"] < Version("15"):
+            issues.append(
+                "backup_compression = %s requires "
+                "pg_basebackup 15 or greater" % self.config.type
+            )
+
+        if self.config.level is not None and (
+            self.config.level < 1 or self.config.level > 22
```
mikewallace1979, line 635: Negative values are allowed here since the zstd library supports [negative compression levels down to -99](https://github.com/facebook/zstd/blob/03cc84fddb9747ddc63f76b6f3171cd8f66da202/programs/zstdcli.c#L796) and pg_basebackup will [pass whatever level is provided to the zstd library](https://github.com/postgres/postgres/blob/REL_15_BETA3/src/backend/replication/basebackup_zstd.c#L99-L106) (unless it's `0`).

barman/compression.py
```diff
@@ -638,6 +641,11 @@ def validate(self, pg_server_version, remote_status):
                 "backup_compression_level %d unsupported by "
                 "pg_basebackup compression %s" % (self.config.level, self.config.type)
             )
+        if self.config.workers is not None and (type(self.config.workers) is not int or self.config.workers < 0):
```
mikewallace1979: Should we also validate that workers is not set in the validation for gzip and lz4?

didiermichel: That was the idea. but I was counting in actually testing it before going further.

didiermichel: This check now exists in both lz4 and gzip contexts

barman/compression.py
mikewallace1979: Ok - that makes sense.

tests/test_compressor.py
```diff
@@ -421,6 +423,13 @@ class TestPgBaseBackupCompression(object):
                 LZ4PgBaseBackupCompressionOption,
                 LZ4Compression,
             ),
+            # Test lz4 scenario
```
mikewallace1979, line 426: s/lz4/zstd/

tests/test_compressor.py
```diff
@@ -1004,3 +1114,144 @@ def test_get_file_content_file_not_found(self):
...
+        # THEN a CommandFailedException is raised
+        with pytest.raises(CommandFailedException) as exc:
+            zstd_compression.uncompress("/path/to/src", "/path/to/dst")
+
+        # AND the exception message contains the command stderr
+        assert "some error" in str(exc.value)
+
+    def test_get_file_content(self):
+        """
+        Cannot create a tar.zst to actually test this without installing ddependency. So we fake it
```
mikewallace1979, line 1185: typo "ddependency"

barman/compression.py
```diff
@@ -600,6 +616,49 @@ def validate(self, pg_server_version, remote_status):
...
+
+        :param pg_server_version int: the server for which the
+          compression options should be validated.
+        :param dict remote_status: the status of the pg_basebackup command
+        :return List: List of Issues (str) or empty list
+        """
+        issues = super(ZSTDPgBaseBackupCompressionOption, self).validate(
+            pg_server_version, remote_status
+        )
+        # "backup_location = server" requires pg_basebackup >= 15
```
mikewallace1979, line 640: This comment seems out of place here given the code is not validating anything related to `backup_location`.

mikewallace1979 (changes requested):
Just a few very minor changes from me.

tests/test_compressor.py
didiermichel: Nice sed command

barman/compression.py
didiermichel: Indeed, the same thing exist with lz4. I will replace it with:
zstd compression requires pg_basebackup >= 15
same with lz4

mikewallace1979 (approved):

PR #647, opened 2022-08-15
Adjusting documentation of `barman backup` command

barthisrael:
The documentation of `barman backup` command both in `man` and
in the website mention that `barman backup` command can be used
to take backups from a given server or from them all sequentially
by using `all` as server name.

While taking a look at the code of barman we can see that the
server name option is filled with `nargs='+'`, which means a call
like the following:

```
barman backup server-1 server-2
```

Would sequentially backup `server-1` and `server-2`.

This commit amends the docs for `man` and for the website to
reflect that behavior.

Signed-off-by: Israel Barth <israel.barth@laptop428-ma-us.local>


mikewallace1979 (approved):
@barthisrael Thanks for verifying the behavior and fixing the gap in the docs 🙌

PR #661, opened 2022-09-07
Dev/601 clean pg prev 10

didiermichel:


barman/postgres.py
```diff
@@ -279,6 +279,10 @@ def server_major_version(self):
             return simplify_version(result)
         return None
 
+    def is_minimal_postgres_version(self):
+        """Checks if postgres version has at least minimal version"""
+        return self.server_version >= 90600
```
mikewallace1979, line 284: Support for 9.6 was dropped in the 3.0.0 release so this should probable be `>= 100000`.

barman/backup_executor.py
```diff
@@ -1825,21 +1825,19 @@ def check(self, check_strategy):
         :param CheckStrategy check_strategy: the strategy for the management
              of the results of the various checks
         """
-        check_strategy.init_check("pgespresso extension")
+        check_strategy.init_check("postgres minimal version")
         try:
             # We execute this check only if the postgres connection is non None
```
mikewallace1979, line 1830: We should rewrite this comment and the docstring above since the meaning of this check has changed from "check whether postgres supports concurrent backup" to "check whether Barman supports this version of postgres.

barman/backup_executor.py
```diff
@@ -1825,21 +1825,19 @@ def check(self, check_strategy):
...
-                        self.server_name,
-                        False,
-                        hint="required for concurrent "
-                        "backups on PostgreSQL %s" % self.postgres.server_major_version,
-                    )
+            if self.postgres and not self.postgres.is_minimal_postgres_version():
+                check_strategy.result(
+                    self.server_name,
+                    False,
+                    hint="unsupported PostgresSQL version %s"
```
mikewallace1979, line 1837: It might be helpful to the user to also include the minimum supported version in this hint.

barman/backup_executor.py
```diff
@@ -1859,15 +1857,12 @@ def start_backup(self, backup_info):
...
-        pg_version = self.postgres.server_version
-        if pg_version >= 90600:
-            # On 9.6+ execute native concurrent start backup
-            _logger.debug("Start of native concurrent backup")
-            self._concurrent_start_backup(backup_info, label)
-        else:
-            # On older Postgres use pgespresso
-            _logger.debug("Start of concurrent backup with pgespresso")
-            self._pgespresso_start_backup(backup_info, label)
+        if not self.postgres.is_minimal_postgres_version():
```
mikewallace1979, line 1860: I wonder whether we can drop this test entirely, given we will have already checked the PostgreSQL version is supported during the check which runs before a backup is executed?

I can see the rationale for having the extra checks when it couldn't be guaranteed that a supported PostgreSQL would support concurrent backup, but now that all PostgreSQL which do not support concurrent backup natively are EOL perhaps it's overkill.

barman/cloud.py
```diff
@@ -1500,7 +1500,7 @@ def backup(self):
         )
         backup_info.set_attribute("systemid", self.postgres.get_systemid())
         controller = self._create_upload_controller(backup_info.backup_id)
-        if self.postgres.server_version >= 90600 or self.postgres.has_pgespresso:
+        if self.postgres.is_minimal_postgres_version():
```
mikewallace1979, line 1503: This change makes sense within the context of the PR however the resulting code reads a little odd now. Rather than use `ExclusiveBackupStrategy` if the PostgreSQL version is unsupported we should instead exit with an error. We should probably do this check earlier in the code execution for all Barman Cloud commands.

mikewallace1979 (commented):
Nice to see the code relating to unsupported PG versions being cleaned up 🙌.

Is the plan to tackle all references to unsupported postgres versions in this PR or just the pgespresso related bits?

barman/postgres.py
didiermichel: I wanted to do it step by step to be sure I will not miss anything.

barman/backup_executor.py
didiermichel: I am not sure check is called before. 
I am looking for an early place in the code to check pg_version and exit if unsupported. but I am not sure to find a unique place to perform this check.

barman/cloud.py
didiermichel: This is a really good point. I will adjust that.

barman/backup_executor.py
didiermichel: Included

barman/backup_executor.py
didiermichel: Changed

barman/postgres.py
```diff
@@ -507,36 +528,16 @@ def server_txt_version(self):
...
     def is_in_recovery(self):
         """
         Returns true if PostgreSQL server is in recovery mode (hot standby)
         """
         try:
-            # pg_is_in_recovery is only available from Postgres 9.0+
-            if self.server_version < 90000:
-                return False
+            # # pg_is_in_recovery is only available from Postgres 9.0+
+            # # Todo: remove
```
mikewallace1979, line 538: Looks like this can just be removed now?

barman/postgres.py
```diff
@@ -893,25 +886,27 @@ def fetch_remote_status(self):
...
+                ]
+            )
+            # Retrieve wal_keep_segments from version 9.0 onwards, until
+            # version 13.0, where it was renamed to wal_keep_size
+            if self.server_version < 130000:
+                pg_settings.append("wal_keep_segments")
+            else:
+                pg_settings.append("wal_keep_size")
+
+            pg_settings.extend(
```
mikewallace1979, line 903: These elements could be added in the first `pg_settings.extend` call so that all new elements which are no longer version dependent are added in one place.

tests/test_postgres.py
```diff
@@ -625,13 +569,13 @@ def test_get_archiver_stats(self, conn):
         # Build a server
         server = build_real_server()
         cursor_mock = conn.return_value.cursor.return_value
-        # expect None as result for server version <9.4
-        conn.return_value.server_version = 80300
-        assert server.postgres.get_archiver_stats() is None
+        # # expect None as result for server version <9.4
```
mikewallace1979, line 572: Can these three lines be removed now?

tests/test_postgres.py
```diff
@@ -962,14 +863,15 @@ def test_get_remote_status(
             "timestamp": datetime.datetime(2016, 3, 30, 17, 4, 20, 271376),
         }
         current_size_mock.return_value = 497354072
-        has_pgespresso_mock.return_value = True
         server_txt_version_mock.return_value = "9.5.0"
+        # server_txt_version_mock.return_value = PostgreSQL.int_version_to_string_version(PostgreSQL.MINIMAL_VERSION)
```
mikewallace1979, line 867: Should this line and L874 also be removed?

mikewallace1979 (changes requested):
Just a few minor things - overall looks great.

We'll need to follow up with a docs PR since much of the docs refers to older PostgreSQL versions in a way that implies they are still supported - I'll file a new issue for that.

barman/postgres.py
didiermichel: indeed

barman/postgres.py
didiermichel: I was not sure if order had meaning somehow and wanted to minimize the risk.  I will do that

tests/test_postgres.py
didiermichel: Done.

tests/test_postgres.py
didiermichel: They can and they will.

mikewallace1979 (approved):

PR #687, opened 2022-10-31
Use a new connection to primary when switching WALs

### LINKED ISSUE #685: SSL error on primary_conninfo
Barman 3.2.0

I have backup configured to be executed on standby server

```
[sqlv2.dc.lan]
active = "True"
description = "Backup from standby"
ssh_command = ssh postgres@sqlv2.dc.lan
conninfo = "host=sqlv2.dc.lan user=barman dbname=postgres"
primary_conninfo = "host=sqlv1.dc.lan user=barman dbname=postgres"
backup_method = "rsync"
reuse_backup = "link"
backup_options = "concurrent_backup"
archiver = on
minimum_redundancy = "1"
retention_policy = "REDUNDANCY 7"`
```

sqlv2 config

```
archive_command = 'barman-wal-archive backup.dc.lan sqlv2.dc.lan %p'
archive_mode = always
wal_level = replica
hot_standby = on
```

starting the very first backup

```
barman backup sqlv2.dc.lan --reuse-backup=off
Starting backup using rsync-concurrent method for server sqlv2.dc.lan in /var/lib/barman/sqlv2.dc.lan/base/20221030T215301
Backup start at LSN: 32A/14006310 (000000010000032A00000014, 00006310)
Starting backup copy via rsync/SSH for 20221030T215301
WARNING: The usage of include directives is not supported for files that reside outside PGDATA.
Please manually backup the following files:
	/etc/postgresql/11/main/conf.d/25ansible_postgresql.conf

Copy done (time: 19 seconds)
Asking PostgreSQL server to finalize the backup.
```

and here it stops and waits endlessly

on standby, I can see

```
2022-10-30 21:55:20.879 CET [4764] barman@postgres WARNING:  pg_stop_backup still waiting for all required WAL segments to be archived (120 seconds elapsed)
2022-10-30 21:55:20.879 CET [4764] barman@postgres HINT:  Check that your archive_command is executing properly.  pg_stop_backup can be canceled safely, but the database backup will not be usable without all the WAL segments.
```

I have tested connectivity via

```
barman-wal-archive --test backup.dc.lan sqlv2.dc.lan DUMMY
b'Ready to accept WAL files for the server sqlv2.dc.lan\n'
```

while on primary I see
```
2022-10-30 1:52:43.399 CET [2081] barman@postgres LOG:  SSL error: decryption failed or bad record mac
2022-10-30 1:52:43.399 CET [2081] barman@postgres LOG:  could not receive data from client: Connection reset by peer
```

barman check command says it's all green and its log doesn't show anything helpful (even with DEBUG level)

### LINKED ISSUE #686: Force a new psycopg2 connection when switching WALs on the primary in a forked process
When `primary_conninfo` is set, Barman will attempt to switch WALs on the primary while the stop backup call is in progress. This is in order to allow backups against standbys to complete even when there is not sufficient DB traffic to cause a natural WAL switch.

Because the WAL switch happens in a forked process it should be creating a new psycopg2 connection because it is unsafe to reuse psycopg2 connections (which are libpq connections under the hood) across multiple processes. Re-using the connection results in the following log messages:

```
2022-10-31 12:54:18.737 CET [4140] LOG:  SSL error: decryption failed or bad record mac
2022-10-31 12:54:18.737 CET [4140] LOG:  could not receive data from client: Connection reset by peer
```

This connection failure is handled by Barman's existing logic which automatically creates a new connection when an existing connection fails, however relying on this is bad practice which could lead to undefined behavior. We should therefore always force a new connection to be created in the forked process.

This issue may be a direct (or indirect) cause of #685.

mikewallace1979:
Creates a new PostgreSQL connection to the primary in the child process which switches WALs on the primary during the call to pg_backup_stop on the standby.

This replaces the old buggy behaviour of reusing the connection from the parent process which is specifically not supported by psycopg2 or the underlying libpq library. This would lead to undefined behaviour including the closure of connections with the message `SSL error: decryption failed or bad record mac`.

We also close the connection once the child process is finished with it which avoids additional `could not receive data from client: Connection reset by peer` messages on the primary due to dangling connections.

Closes #686.
May also close #685.

### DIFF
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -1563,7 +1563,7 @@ class StandbyPostgreSQLConnection(PostgreSQLConnection):
     def __init__(
         self,
         conninfo,
-        primary_conn,
+        primary_conninfo,
         immediate_checkpoint=False,
         slot_name=None,
         application_name="barman",
@@ -1572,7 +1572,7 @@ class StandbyPostgreSQLConnection(PostgreSQLConnection):
         Standby PostgreSQL connection constructor.
 
         :param str conninfo: Connection information (aka DSN) for the standby.
-        :param barman.postgres.PostgreSQLConnection primary_conn: Connection to the
+        :param str primary_conninfo: Connection information (aka DSN) for the
             primary.
         :param bool immediate_checkpoint: Whether to do an immediate checkpoint
             when a backup is started.
@@ -1587,7 +1587,10 @@ class StandbyPostgreSQLConnection(PostgreSQLConnection):
         )
         # The standby connection has its own connection object used to talk to the
         # primary when switching WALs.
-        self.primary = primary_conn
+        self.primary_conninfo = primary_conninfo
+        # The standby needs a connection to the primary so that it can
+        # perform WAL switches itself when calling pg_backup_stop.
+        self.primary = PostgreSQLConnection(self.primary_conninfo)
 
     def close(self):
         """Close the connection to PostgreSQL."""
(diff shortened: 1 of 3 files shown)


didiermichel (approved):
All seems fine to me

PR #702, opened 2022-11-08
Friendly backup names

mikewallace1979:


barman/utils.py
```diff
@@ -775,3 +777,47 @@ def check_size(value):
...
+    if backup_name in (RESERVED_BACKUP_IDS):
+        raise ArgumentTypeError(
+            "Backup name '%s' is not allowed: reserved word" % backup_name
+        )
+    return backup_name
+
+
+def is_backup_id(backup_id):
+    # TODO check if backup_id matches pattern (is this logic somewhere already?)
+    return bool(re.match(r"(\d{8})T\d{6}$", backup_id))
```
mikewallace1979, line 802: I have a couple of questions regarding this check:
1. It's pretty weak. Should we instead be verifying it is a parseable short-form ISO 8601 timestamp?
2. Has the backup ID historically always been short-form ISO 8601? Is it possible a backup ID could be valid but not pass this check?

mikewallace1979:
Closing in favour of #705

PR #707, opened 2022-11-20
Being able to change recovery config file path

mhkarimi1383:
sometimes we don't want to store generated configs related to recovery into the default file (e.g. `postgresql.auto.conf`)
so I added `--recovery-conf-filename` option to specify the filename

for example in our environment we are using https://github.com/basalam/kaastolon which links  `postgresql.auto.conf` to `/dev/null` in order to avoid `SET` queries on postgresql

credits to @mojtabash78 (for testing and using it in our production)

### DIFF
--- a/barman/recovery_executor.py
+++ b/barman/recovery_executor.py
@@ -375,7 +375,7 @@ class RecoveryExecutor(object):
             if backup_info.version < 120000:
                 results["recovery_configuration_file"] = "recovery.conf"
         else:
-            recovery_info["configuration_files"].append(recovery_conf_filename)
+            recovery_info["configuration_files"].append(recovery_conf_filename) if (recovery_conf_filename in recovery_info["configuration_files"]) else 0
             results["recovery_configuration_file"] = recovery_conf_filename
         # Handle remote recovery options
         if remote_command:
--- a/barman/recovery_executor.py
+++ b/barman/recovery_executor.py
@@ -314,8 +314,8 @@ class RecoveryExecutor(object):
         # 3) copy
         output.info("Identify dangerous settings in destination directory.")
 
-        self._map_temporary_config_files(recovery_info, backup_info, remote_command, recovery_conf_filename)
-        self._analyse_temporary_config_files(recovery_info, recovery_conf_filename)
+        self._map_temporary_config_files(recovery_info, backup_info, remote_command)
+        self._analyse_temporary_config_files(recovery_info)
         self._copy_temporary_config_files(dest, remote_command, recovery_info)
 
         return recovery_info
@@ -365,8 +365,8 @@ class RecoveryExecutor(object):
         }
         recovery_info["results"] = results
         # Set up a list of configuration files
+        recovery_info["configuration_files"].append("postgresql.conf")
         if not recovery_conf_filename:
-            recovery_info["configuration_files"].append("postgresql.conf")
             if backup_info.version >= 90400:
                 recovery_info["configuration_files"].append("postgresql.auto.conf")
 
@@ -376,6 +376,7 @@ class RecoveryExecutor(object):
                 results["recovery_configuration_file"] = "recovery.conf"
         else:
             recovery_info["configuration_files"].append(recovery_conf_filename)
+            results["recovery_configuration_file"] = recovery_conf_filename
         # Handle remote recovery options
         if remote_command:
             recovery_info["recovery_dest"] = "remote"
@@ -1112,7 +1113,7 @@ class RecoveryExecutor(object):
             conf_file_paths.append(conf_file_path)
         return conf_file_paths
 
-    def _map_temporary_config_files(self, recovery_info, backup_info, remote_command, recovery_conf_filename):
+    def _map_temporary_config_files(self, recovery_info, backup_info, remote_command):
         """
         Map configuration files, by filling the 'temporary_configuration_files'
         array, depending on remote or local recovery. This array will be used
@@ -1123,7 +1124,6 @@ class RecoveryExecutor(object):
         :param barman.infofile.LocalBackupInfo backup_info: a backup
             representation
         :param str remote_command: ssh command for remote recovery
-        :param str|None recovery_conf_filename: filename for storing recovery configurations
         """
 
         # Cycle over postgres configuration files which my be missing.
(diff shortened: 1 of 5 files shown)


barman/recovery_executor.py
```diff
@@ -361,17 +364,18 @@ def _setup(self, backup_info, remote_command, dest):
...
+            recovery_info["configuration_files"].append("postgresql.conf")
+            if backup_info.version >= 90400:
+                recovery_info["configuration_files"].append("postgresql.auto.conf")
+
+            # Identify the file holding the recovery configuration
+            results["recovery_configuration_file"] = "postgresql.auto.conf"
+            if backup_info.version < 120000:
+                results["recovery_configuration_file"] = "recovery.conf"
+        else:
+            recovery_info["configuration_files"].append(recovery_conf_filename)
```
mikewallace1979, line 378: When `recovery_conf_filename` is set it looks like `postgresql.conf` is not going to be added to `recovery_info["configuration_files"]`. This means the config mangling which happens in [_analyse_temporary_config_files](https://github.com/EnterpriseDB/barman/blob/5cdd66eab0aad41d0e1e0b0eac9f4660e6bb85d4/barman/recovery_executor.py#L1196-L1212) will skip the `postgresql.conf` file whenever a custom recovery config file path is used. It feels like we'd be better off always including `postgresql.conf` in `recovery_info["configuration_files"]`, since the recover config file doesn't change the fact that Barman needs to scan `postgresql.conf`.

Also, you should set `results["recovery_configuration_file"]` to `recovery_conf_filename` when it is set - this will make the filename appear in the `barman recover` output. You could also reference it via the `recovery_info["results"]["recovery_configuration_file"]` in the functions which currently have a dedicated `recovery_conf_filename` parameter.

mikewallace1979 (changes requested):
Thanks for making this PR - I can see it being a useful feature in environments like stolon.

The changes look reasonable - I did leave a comment in the changes to `RecoveryExecutor._setup` as I think there are some small changes that need to be made there.

There are a few unit tests failing as a result of this change so we need to fix those before merging. It would also be nice if there was a new test which verified that the `recovery_conf_filename` would override the default values.

The only other thing missing I think is to add the new option to the [recover docs](https://github.com/basalam/barman/blob/5cdd66eab0aad41d0e1e0b0eac9f4660e6bb85d4/doc/barman.1.d/50-recover.md) so that the option is also included in the man page and online docs.

Thanks again for the PR :)

barman/recovery_executor.py
mhkarimi1383: I tried to fixed it please check it out

Thanks.

mhkarimi1383:
Please check my changes in https://github.com/EnterpriseDB/barman/pull/707/commits/7cd4099c2e15b8faefa4df708fe2cd0f0ac14848 and verify my changes in unit test.

mikewallace1979:
> Please check my changes in [7cd4099](https://github.com/EnterpriseDB/barman/commit/7cd4099c2e15b8faefa4df708fe2cd0f0ac14848) and verify my changes in unit test.

Thanks for making the changes - I will take a closer look later this week.

mhkarimi1383:
@mikewallace1979 
Hi
any updates on this?

mikewallace1979:
Hi @mhkarimi1383 - apologies, I had hoped to get this into the 3.3.0 release but ran out of time. We should be able to get this into the January release.

mhkarimi1383:
> Hi @mhkarimi1383 - apologies, I had hoped to get this into the 3.3.0 release but ran out of time. We should be able to get this into the January release.

Olay, I'm waiting for that 🤓

mikewallace1979:
> Olay, I'm waiting for that 🤓

I had some time today to take a proper look at this and ended up with a few additional commits which I've pushed (along with these) to a branch [here](https://github.com/EnterpriseDB/barman/pull/718).

Mostly it's formatting, unittests and docs however there is [one commit](https://github.com/EnterpriseDB/barman/pull/718/commits/2b36511e0a06cb4a0a7c89f2ea75f1ed84d2a983) which changes the behaviour:

    1. `postgresql.auto.conf` is now always added to configuration_files.
        The reason being that there are many reasons why someone might use
        a different file for the recovery options, so there may still be a
        real `postgresql.auto.conf`. By always adding it to
        `recovery_info["configuration_files"]` we ensure that Barman either
        checks and mangles the options in the file or logs that it cannot
        find the file and proceeds with the rest of the recovery.
     2. If recovery_conf_filename is set then it is not added to
        `recovery_info["configuration_files"]` because it is a custom
        destination for recovery options which does not contain existing
        options.

For the particular case you want to address, where `postgresql,auto.conf` is a symlink to `/dev/null`, the only change in behaviour is that the following message is now displayed during recovery:

```
WARNING
The following configuration files have not been saved during backup, hence they have not been restored.
You need to manually restore them in order to start the recovered PostgreSQL instance:

    postgresql.auto.conf
```

This seems reasonable to me since anyone using stolon will know this is a non-issue and the recovery options can then be picked up from the filename specified by `--recovery-conf-filename`.

What do you think? If this sounds good to you I can tidy up the commit history on that branch and then get another Barman developer to review the complete set of changes.

mhkarimi1383:
Thanks,
Sounds good to me too.
But I want to make sure that when you put in some default postgresql config file like `postgresql.conf` everything will be Okay

mikewallace1979:
Ok we can add a test case for using `--recovery-conf-filename=postgresql.conf`, though I would recommend instead adding either an [include or include_dir directive](https://www.postgresql.org/docs/current/config-setting.html) in `postgresql.conf` and then having `--recovery-conf-filename` point to a file which will be included by either `include` or `include_dir`.

mhkarimi1383:
> Ok we can add a test case for using `--recovery-conf-filename=postgresql.conf`, though I would recommend instead adding either an [include or include_dir directive](https://www.postgresql.org/docs/current/config-setting.html) in `postgresql.conf` and then having `--recovery-conf-filename` point to a file which will be included by either `include` or `include_dir`.


Adding a test case for that is a great idea

But when using Stolon/Kaastolon inside K8s using helm is not a good idea to have something like `include_dir` or `include` since we are mounting configs using `configmap`s and having multiple `configmap`s is not a good solution there (as my experience)

mikewallace1979:
Hi @mhkarimi1383 - before I merge https://github.com/EnterpriseDB/barman/pull/718 can you confirm you're happy for your contribution to go in? I've squashed the original commits in this PR into a [single commit](https://github.com/EnterpriseDB/barman/pull/718/commits/ccb6b794eb1829aa10c8f3f5f8c444471990fd0d) and then added my changes as separate commits so as to keep a clear record of authorship in the commit history. If you're happy with that I can merge it and the change will (finally) go out in the March release.

mhkarimi1383:
Hi @mikewallace1979
Of course I'm happy with the changes 🤠

mikewallace1979:
Merged in #718 - thank you for your contribution.

Issue #298, opened 2020-08-19
barman check streaming get faild


Roarain:
# check
```shell
# barman --log-level NOTSET check streaming
Server streaming:
	WAL archive: FAILED (please make sure WAL shipping is setup)
	PostgreSQL: OK
	superuser or standard user with backup privileges: OK
	PostgreSQL streaming: OK
	wal_level: OK
	replication slot: FAILED (slot 'barman' not initialised: is 'receive-wal' running?)
	directories: OK
	retention policy settings: OK
	backup maximum age: OK (no last_backup_maximum_age provided)
	compression settings: OK
	failed backups: OK (there are 0 failed backups)
	minimum redundancy requirements: OK (have 0 backups, expected at least 0)
	pg_basebackup: OK
	pg_basebackup compatible: OK
	pg_basebackup supports tablespaces mapping: OK
	systemid coherence: OK (no system Id stored on disk)
	pg_receivexlog: OK
	pg_receivexlog compatible: OK
	receive-wal running: FAILED (See the Barman log file for more details)
	archiver errors: OK
```

# backup
```shell
# barman backup streaming
ERROR: Impossible to start the backup. Check the log for more details, or run 'barman check streaming'
```

# pg10 instance config
```sql
postgres=# select name, setting from pg_settings where name in ('archive_mode', 'max_replication_slots', 'wal_level', 'max_wal_senders', 'hot_standby', 'synchronous_standby_names');
           name            | setting
---------------------------+---------
 archive_mode              | on
 hot_standby               | on
 max_replication_slots     | 10
 max_wal_senders           | 16
 synchronous_standby_names |
 wal_level                 | replica
```

# How to solve these errors 

1. WAL archive: FAILED (please make sure WAL shipping is setup)
2. replication slot: FAILED (slot 'barman' not initialised: is 'receive-wal' running?)
3. receive-wal running: FAILED (See the Barman log file for more details)

Roarain:
2 and 3 has resolved by exec 
```shell 
barman receive-wal streaming
```

but 1 has no idea .
Help !!!!

amenonsen:
As part of an effort to pick up maintenance of Barman, I'm closing this and many other old open issues in this repository.

Issue #88, opened 2017-03-06
Streaming-only backups/WAL archiving?


danbst:
I try to do streaming-only Barman backup scenario. My `barman.conf` is
```
      [barman]
      barman_user = barman
      configuration_files_directory = /etc/barman.d
      barman_home = ${barmanHome}
      log_level = INFO
      compression = gzip

      [live]
      description =  "Streaming copy of live"
      conninfo = host=${liveHostname} user=barman dbname=postgres
      streaming_conninfo = host=${liveHostname} user=streaming_barman
      backup_method = postgres
      archiver = off
      streaming_archiver = on
      slot_name = barman
```
My `postgresql.conf` is
```
synchronous_standby_names = 'barman_receive_wal'
max_wal_senders = 10
max_replication_slots = 10
wal_level = 'hot_standby'
```

I do start wal receive process:
```
# barman receive-wal live
Starting receive-wal for server live
2017-03-06 15:45:59,311 [4456] barman.server INFO: Starting receive-wal for server live
2017-03-06 15:45:59,324 [4456] barman.wal_archiver INFO: Synchronous WAL streaming for barman_receive_wal: True
2017-03-06 15:45:59,325 [4456] barman.wal_archiver INFO: Activating WAL archiving through streaming protocol
live: pg_receivexlog: starting log streaming at 0/2000000 (timeline 1)
2017-03-06 15:45:59,330 [4456] barman.command_wrappers INFO: live: pg_receivexlog: starting log streaming at 0/2000000 (timeline 1)
```

But `barman check live` complains:
```
# barman check live
Server live:
2017-03-06 21:41:54,942 [6044] barman.server ERROR: Check 'WAL archive' failed for server 'live'
        WAL archive: FAILED (please make sure WAL shipping is setup)
2017-03-06 21:41:54,957 [6044] barman.wal_archiver INFO: Synchronous WAL streaming for barman_receive_wal: True
        PostgreSQL: OK
        superuser: OK
        PostgreSQL streaming: OK
        wal_level: OK
        replication slot: OK
        directories: OK
        retention policy settings: OK
        backup maximum age: OK (no last_backup_maximum_age provided)
        compression settings: OK
        failed backups: OK (there are 0 failed backups)
        minimum redundancy requirements: OK (have 0 backups, expected at least 0)
        pg_basebackup: OK
        pg_basebackup compatible: OK
        pg_basebackup supports tablespaces mapping: OK
        pg_receivexlog: OK
[...]

PR #720, opened 2022-12-20
Check WAL shipping setup only in archiver mode

andrey-utkin:
Setups with only streaming_archiver enabled will fail this check. This blocks "backup" operation, while receive-wal works.

This appears to be previously reported issue, for example see https://github.com/EnterpriseDB/barman/issues/298

This bug may be explained by the fact previously enabling both streaming_archiver and archiver (WAL shipping) was the recommended configuration, per documentation. In such configuration this check would pass.

### DIFF
--- a/barman/server.py
+++ b/barman/server.py
@@ -641,27 +641,28 @@ class Server(RemoteStatusMixin):
              of the results of the various checks
         """
         check_strategy.init_check("WAL archive")
-        # Make sure that WAL archiving has been setup
-        # XLOG_DB needs to exist and its size must be > 0
-        # NOTE: we do not need to acquire a lock in this phase
-        xlogdb_empty = True
-        if os.path.exists(self.xlogdb_file_name):
-            with open(self.xlogdb_file_name, "rb") as fxlogdb:
-                if os.fstat(fxlogdb.fileno()).st_size > 0:
-                    xlogdb_empty = False
+        if self.config.archiver:
+            # Make sure that WAL archiving has been setup
+            # XLOG_DB needs to exist and its size must be > 0
+            # NOTE: we do not need to acquire a lock in this phase
+            xlogdb_empty = True
+            if os.path.exists(self.xlogdb_file_name):
+                with open(self.xlogdb_file_name, "rb") as fxlogdb:
+                    if os.fstat(fxlogdb.fileno()).st_size > 0:
+                        xlogdb_empty = False
 
-        # NOTE: This check needs to be only visible if it fails
-        if xlogdb_empty:
-            # Skip the error if we have a terminated backup
-            # with status WAITING_FOR_WALS.
-            # TODO: Improve this check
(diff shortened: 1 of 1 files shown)


andrey-utkin:
Hey, I'm glad to offer this tiny patch, but please take it very critically as it's my first time ever using barman (although not the first time setting up Postgres replication), so I might not know what I've done here.

mikewallace1979:
Hi @andrey-utkin - thanks for reporting the issue and providing a patch. The patch itself looks fine however skipping this particular check if `archiver = off` isn't quite the full solution.

The short explanation is that the `WAL archive` check is checking whether Barman knows about any WALs at all, be they received either via the PostgreSQL `archive_command` or `barman receive-wal`. When this check fails it is because the file used to store WAL metadata (`xlog.db`) is either not present or empty; this means that Barman has not yet received any complete WALs.

Whether `archiver` or `streaming_archiver` (or both) is used, the `WAL archive` check is still relevant, because it is verifying whether any WALs at all exist in Barman's WAL archive. The usual way to deal with this failing check when setting up a new server in Barman is to force a WAL switch on the PostgreSQL server using `barman switch-wal` with the `--force` and `--archive` options.

However when `streaming_archiver` is used and `archiver` is not used it is not necessary to wait for a complete WAL to be copied into Barman's WAL archive - it might be enough to verify that a `.partial` file is present in the path specified by `streaming_wals_directory`. While this would tell us whether or not WAL streaming is working it would not verify that the full process of writing WALs into Barman's WAL archive works end-to-end (for example, it would be possible for the end-to-end process to fail if `wals_directory` pointed to a path which was not writable by the barman user).

So there is still an opportunity here to improve the check when `archiver = off` and `streaming_archiver = on` but in addition to skipping the `xlog.db` check we need to do the following:

1. Check for the presence of a `.partial` file in `streaming_wals_directory`.
2. Check that `wals_directory` is writable by the barman user.

andrey-utkin:
Hi, thanks a lot for the guidance!
Sorry for the long delay.

I've added the two checks you suggested. I have tested it on a new streaming-only setup, so it works at least on my machine.
How do they look to you?
I am not sure should I have used f-string. Do you support Python older than 3.6?

andrey-utkin:
Kindly reminding to check the updated patch.

mikewallace1979:
Hi @andrey-utkin - thanks for the reminder, we'll re-review this shortly.

mikewallace1979 (commented):
Thanks for making the changes and sorry for making the suggestion about checking for `.partial` files.

We still need to come up with a better way of verifying WAL streaming is working as intended in order to skip the `xlog.db` check however we'll need to think about it a bit more to figure out the right approach.

vnghia:
Hello, I would like to ask the status of this PR. We have an issue https://github.com/EnterpriseDB/barman/issues/88 that received a lot of attentions and the root cause is still not solved.

Thank you very much!

PR #726, opened 2023-01-09
Fix broken assertions in compression tests

mikewallace1979:
Replaces `assert command.cmd.called_once()` with a call to the corresponding `assert_called_once` function. This fixes the assertion which was previously only passing due to the mock returning a truthy value.

Thanks to version 5.0.1 of the mock library for detecting this.


mikewallace1979:
The relevant commit in the mock library is this one: https://github.com/testing-cabal/mock/commit/21787a948cf1186f3bd38462d5b912784fa81218

didiermichel (approved):

Issue #620, opened 2022-07-12
Snapshot-based backups in GCS


mikewallace1979:
Updating the ticket description since this issue went well beyond proof-of-concept and ended up being the final product.

PR #728, opened 2023-01-12
Add support for GCP persistent disk snapshots

mikewallace1979:
This is the PR #710 branch ([#620-gce-snapshots-barman-cloud-refactor](https://github.com/EnterpriseDB/barman/tree/620-gce-snapshots-barman-cloud-refactor)) with the commit history rewritten into a more easily consumable form, PR feedback from the draft PR addressed, and rebased onto the latest barman trunk.

The diff between this PR branch and #710 can be found [here](https://github.com/EnterpriseDB/barman/compare/620-gce-snapshots-barman-cloud-refactor..620-gcp-snapshots).

Notes for reviewers:

- The new code is broken down into four separate commits:
  - Support for snapshot backups in Barman.
  - Support for recovery of snapshot backups in Barman.
  - Support for snapshot backups in Barman Cloud.
  - Support for restore of snapshot backups in Barman Cloud.
 - Each commit is preceded by one or more refactoring commits which prepare the existing code for the snapshot-related additions.
 - Unit tests for each part of the feature (or refactoring) are included in the relevant commit.
 - It is suggested that the code is reviewed on a commit-by-commit basis.
 - No documentation changes are included. When the documentation changes are ready they will be a separate PR as this is already a +++ridiculous LOC PR.
 - There is no corresponding integration tests PR. The integration tests still need to run to validate that no existing functionality is broken however snapshot functionality will be tested using a specific test suite for the snapshot feature (not yet ready for PR but a functioning branch is available [here](https://github.com/EnterpriseDB/barman-av/compare/snapshot-integration-testing?expand=1)).

### DIFF
--- a/barman/backup_executor.py
+++ b/barman/backup_executor.py
@@ -40,6 +40,7 @@ import dateutil.parser
 from distutils.version import LooseVersion as Version
 
 from barman import output, xlog
+from barman.cloud_providers import get_snapshot_interface_from_server_config
 from barman.command_wrappers import PgBaseBackup
 from barman.compression import get_pg_basebackup_compression
 from barman.config import BackupOptions
@@ -54,7 +55,7 @@ from barman.exceptions import (
     SshCommandException,
     FileNotFoundException,
 )
-from barman.fs import UnixLocalCommand, UnixRemoteCommand
+from barman.fs import UnixLocalCommand, UnixRemoteCommand, unix_command_factory
 from barman.infofile import BackupInfo
 from barman.postgres_plumbing import EXCLUDE_LIST, PGDATA_EXCLUDE_LIST
 from barman.remote_status import RemoteStatusMixin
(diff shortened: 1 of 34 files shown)


barman/cloud_providers/google_cloud_storage.py
```diff
@@ -333,3 +337,282 @@ def _delete_objects_batch(self, paths):
...
+                        disk_name,
+                    ),
+                },
+            }
+        )
+
+        _logger.info("Waiting for snapshot '%s' completion", snapshot_name)
+        resp.result()
+
+        if resp.done():
```
mikewallace1979, line 458: Checking `.done()` to determine success is ok but the [examples given in the GCP docs](https://cloud.google.com/compute/docs/disks/create-snapshots#create_zonal_snapshot) use the presence of the `error_code` field to determine success status.

Additionally, the examples check the `.warning` field and log anything present, which seems like a practice worth following here.

barman/cloud_providers/google_cloud_storage.py
```diff
@@ -333,3 +337,282 @@ def _delete_objects_batch(self, paths):
...
+                {
+                    "project": self.project,
+                    "snapshot": snapshot_name,
+                }
+            )
+        except NotFound:
+            # If the snapshot cannot be found then deletion is considered successful
+            return
+        resp.result()
+        if resp.done():
```
mikewallace1979, line 542: As with the comment on L458, we should be checking `resp.error_code` to determine if the request was successful and also checking and logging any `resp.warnings`.

barman/cloud_providers/google_cloud_storage.py
```diff
@@ -333,3 +337,282 @@ def _delete_objects_batch(self, paths):
...
+        """
+        Take a snapshot of a persistent disk in GCP.
+
+        :param barman.infofile.LocalBackupInfo backup_info: Backup information.
+        :param str disk_zone: The zone in which the disk resides.
+        :param str disk_name: The name of the source disk for the snapshot.
+        :rtype: str
+        :return: The name used to reference the snapshot with GCP.
+        """
+        snapshot_name = "%s-%s-%s" % (
```
mikewallace1979, line 434: If the server and disk names are reasonably long then there is a risk of exceeding the maximum allowed length of a snapshot name in GCP (66 characters I think?).

Is it worth omitting the server name to give a little more breathing space?

barman/recovery_executor.py
```diff
@@ -1447,13 +1462,293 @@ def _backup_copy(
...
+        """
+        snapshot_interface = get_snapshot_interface_from_backup_info(backup_info)
+        attached_snapshots = self.get_attached_snapshots_for_backup(
+            snapshot_interface, backup_info, recovery_instance, recovery_zone
+        )
+        cmd = fs.unix_command_factory(remote_command, self.server.path)
+        SnapshotRecoveryExecutor.check_mount_points(
+            backup_info, attached_snapshots, cmd
+        )
+        self.check_recovery_dir_exists(dest, cmd)
```
mikewallace1979, line 1670: I wonder if we should also be checking ownership and permissions of the recovery directory (and any tablespace directories). The reason being that, unless the recovery instance is provisioned exactly the same way as the original backup instance, the postgres user and group may have different uid and gids on the recovery instance. If that happens then the files on the disks cloned from the snapshots will not belong to the postgres user on the recovery instance.

In such cases we won't notice until we try and interact with the config files during Barman's part of the recovery process. At that point we will fail noisily and it *should* be clear what the problem is, but it would be nicer if Barman detected it here and failed with a nice human-readable warning.

barman/cloud_providers/google_cloud_storage.py
```diff
@@ -333,3 +337,282 @@ def _delete_objects_batch(self, paths):
...
+                raise SnapshotBackupException(
+                    "Disk %s not attached to instance %s" % (disk_name, instance_name)
+                )
+            # We should always have exactly one attached disk matching the name
+            assert len(attached_disks) == 1
+
+            metadata["device_name"] = attached_disks[0].device_name
+            metadata["device_path"] = self.DEVICE_PREFIX + attached_disks[0].device_name
+            metadata["name"] = self.take_snapshot(backup_info, zone, disk_name)
+            snapshots.append(metadata)
```
mikewallace1979, line 516: @didiermichel I'm adding this comment so we remember to update the snapshots_info format as discussed on slack

mikewallace1979: Resolving this as the snapshots_info metadata changes are done.

barman/cloud_providers/google_cloud_storage.py
didiermichel: "The chain name must be 1-63 characters long and comply with RFC1035".
It is probably worth ensuring it does not exceed 63 characters. Dropping vm instance name is an acceptable solution.

barman/backup_executor.py
```diff
@@ -1388,6 +1384,286 @@ def _format_src(self, path):
...
+            self.snapshot_instance,
+            self.snapshot_zone,
+            self.snapshot_disks,
+        )
+
+        if len(missing_disks) > 0:
+            check_strategy.result(
+                self.config.name,
+                False,
+                hint="cannot find snapshot disks attached to instance %s: %s"
```
mikewallace1979, line 1629: This reads a bit weird because "snapshots disks" sounds like the disks should already be cloned from snapshots, but what we really mean is that they are disks for which snapshots should be taken at backup time.

barman/cloud_providers/google_cloud_storage.py
mikewallace1979: `backup_info.server_name` is going to be the name of the server as known by Barman rather than the instance name, but I am still thinking it's ok to drop it in return for more available characters in the disk name.

The only really important thing is that the snapshot name is unique for a given backup/disk combination and that's going to be true unless someone has attached the same disk to multiple servers (which hopefully isn't even possible?).

barman/backup_executor.py
mikewallace1979: On reflection I think this is ok since `snapshot_disks` is language already used in the Barman config / barman-cloud args.

amenonsen (approved):
Looks good.

barman/cloud_providers/google_cloud_storage.py
mikewallace1979: This ended up being disk name + backup ID because snapshot names cannot start with numbers.

barman/cloud_providers/google_cloud_storage.py
mikewallace1979: This was updated so we use error_code to check for success and log any warnings.

barman/recovery_executor.py
mikewallace1979: This would be a good future enhancement but I don't think it is necessary for the first iteration of snapshot backups since we already fail noisily in this scenario.

PR #732, opened 2023-01-18
include new line so that enumeration is properly formatted

patrickbucher:


mikewallace1979 (approved):
Great catch @patrickbucher - thank you.

PR #741, opened 2023-01-27
Alter workflows for foundation-security

OliverRiggs:
Add a trigger for blackduck on tag

### DIFF
--- a/.github/workflows/blackduck-scan.yml
+++ b/.github/workflows/blackduck-scan.yml
@@ -1,45 +1,61 @@
-name: BlackDuck-Scan-Workflow
+###
+##  Foundation-security BlackDuck workflow 
+#
+name: Foundation-Security/Black Duck Scan
 
 on:
-  # Triggers the workflow on push request events but only for the master branch
-  pull_request:
-    branches: [ master ]
-
-  # Allows you to run this workflow manually from the Actions tab
+  push:
+    tags:        
+      - '**'
   workflow_dispatch:
 
 jobs:
-  scan:
-    name: BlackDuck-scan
-    # The type of runner that the job will run on
-    runs-on: ubuntu-latest
-    if: github.event.pull_request.head.repo.full_name == 'EnterpriseDB/barman'
-
+  Blackduck-Scan:
+    runs-on: ubuntu-20.04
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/blackduck-scan.yml
```diff
@@ -1,45 +1,61 @@
...
+      uses: actions/checkout@v3
+
+    - name: Checkout foundation repository
+      id: checkout-foundation
+      uses: actions/checkout@v3
+      with:
+          repository: EnterpriseDB/foundation-packaging
+          ref: main
+          path: foundation
+          token: ${{secrets.GH_SLONIK}}
```
didiermichel, line 27: This secret does not exist and must be added manually.

.github/workflows/blackduck-scan.yml
```diff
@@ -1,45 +1,61 @@
...
-          args: '--blackduck.url="${{ secrets.BLACKDUCK_URL}}" --blackduck.api.token="${{ secrets.BLACKDUCK_API_TOKEN}}" --detect.project.name="Barman" --detect.project.version.name="CI/CD" --detect.risk.report.pdf=true --detect.excluded.detector.types="npm,cpan,nuget,git,maven,lerna" --detect.blackduck.signature.scanner.license.search'
+    - name: Checkout source repository
+      id: checkout-source
+      uses: actions/checkout@v3
+
+    - name: Checkout foundation repository
+      id: checkout-foundation
+      uses: actions/checkout@v3
+      with:
+          repository: EnterpriseDB/foundation-packaging
```
didiermichel, line 24: Not sure foundation-packaging can be checkout from barman repository as it is a public repository and it does not inherit enterprise secrets.

mikewallace1979: This is a very good point. I wonder if it would be more appropriate to add this scanning to https://github.com/EnterpriseDB/barman-packaging instead of the OSS repo?

OliverRiggs: Alright, that is good to know about the secrets. In any case, Foundation Security changes need to be made to Sonarqube and Blackduck. Please make me aware of the desired repo, with write permissions.

OliverRiggs: I can see that barman-packaging already uses foundation with its branch. No scanning is present in that repo though.

PR #751, opened 2023-02-20
Disable `idle_session_timeout` for rsync-concurrent backups

barthisrael:
Previous to this commit rsync-concurrent backups could fail with log messages like this:

```
2023-02-20 20:18:07,846 [31148] barman.backup_executor INFO: Asking PostgreSQL server to finalize the backup.
2023-02-20 20:18:07,846 [31148] barman.backup ERROR: Backup failed issuing stop backup command (native concurrent).
DETAILS: Connection lost, reconnection not allowed
```

That issue is triggered whenever Barman copy takes more than `idle_session_timeout` to complete because `pg_backup_start` and `pg_backup_stop` need to be executed in the same session, and the session is kept `idle` for as long as the copy takes.

This commit fixes that issue by disabling `idle_session_timeout` in the PostgreSQL session where `pg_backup_start`/`pg_backup_stop` are run.

### DIFF
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -1148,6 +1148,14 @@ class PostgreSQLConnection(PostgreSQL):
                 # PostgreSQLs below 15 have a boolean parameter to specify
                 # not to use exclusive backup
                 pg_backup_args = "%s, %s, FALSE"
+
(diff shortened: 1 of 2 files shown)


mikewallace1979 (approved):
Looks good - thanks for updating the unit tests.

PR #773, opened 2023-03-15
allow use script set-version.sh on 3.4.X branch

didiermichel:


mikewallace1979:
Looks good, but shouldn't the base branch for this PR be `3.4.x`?

didiermichel:
Indeed that's the kind of issues we will face from now on...

PR #781, opened 2023-03-28
Prepare release 3.5.0

mikewallace1979:


didiermichel:
I think this change should be included https://github.com/EnterpriseDB/barman/pull/779

mikewallace1979:
> I think this change should be included #779

You're right - let's merge it, rebase this one and also add the pg10 note to the release notes.

mikewallace1979:
> #779 merged but didn't generate the doc from md files.

No problem - the doc changes affect the manual only, so the generated docs are no different.

amenonsen (approved):

PR #789, opened 2023-04-24
Add links to referenced manual sections

ewie:
Improve the manual by linking to referenced sections instead of just referencing section names and have readers search the text.

The links work in HTML and PDF when building with `doc/build/build`. But I had to change `doc/Dockerfile` to use base image `debian:buster` as Stretch is EOL, causing `apt-get update` to fail. Also need to install package `texlive-fonts-recommended` and remove `latex-xcolor` and `texlive-math-extra`. The Dockerfile change is not included in this PR because I don't know if there are already plans to fix the doc build.

Unfortunately, Pandoc does not warn or fail on unknown anchors. This is prone to typos, as are references to section names.

### DIFF
--- a/doc/manual/00-head.en.md
+++ b/doc/manual/00-head.en.md
@@ -9,4 +9,5 @@
 > **IMPORTANT:** \newline
 > This manual assumes that you are familiar with theoretical disaster
 > recovery concepts, and that you have a grasp of PostgreSQL fundamentals in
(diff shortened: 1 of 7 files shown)


martinmarques:
This doesn't apply anymore with the new doc structure. We'd be happy to get any new contributions over the new docs.

PR #792, opened 2023-05-06
✨ feat: cluster awareness added

mhkarimi1383:
On clustered PostgreSQL instances when master changes we have to run

```bash
barman receive-wal --create-slot
barman receive-wal --reset
```

So I added a new check in `cron` command and if slot is not active run the above commands (or functions) automatically

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,6 +1,5 @@
 Barman TODO - Backlog of features for Barman
 
-* Cluster-awareness
 * TAR copy method
 * TAR storage strategy for backups
 * Centralised WAL hub capability (streaming server)
--- a/barman/server.py
+++ b/barman/server.py
@@ -186,10 +186,12 @@ class CheckStrategy(object):
             else:
                 # otherwise simply log the error (as info)
                 _logger.info(
-                    "Ignoring failed check '%s' for server '%s'" % (check, server_name)
+                    "Ignoring failed check '%s' for server '%s'" % (
+                        check, server_name)
                 )
         else:
-            _logger.debug("Check '%s' succeeded for server '%s'" % (check, server_name))
+            _logger.debug("Check '%s' succeeded for server '%s'" %
+                          (check, server_name))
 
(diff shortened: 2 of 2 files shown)


mhkarimi1383:
@mikewallace1979 
Hi, can you review this?
we are having a bad experience with our clusters using barman

mikewallace1979:
Hi @mhkarimi1383 - it looks like you should be able to achieve part of what this patch does by setting [`create_slot = auto`](https://docs.pgbarman.org/release/3.5.0/#how-to-configure-the-wal-streaming) in the Barman config.

The automatic usage of the reset option is going to be problematic in some scenarios since it causes the removal of any existing `.partial` files - while that may be the right action in the particular case you are considering, there are going to be other times where streaming should still resume from an existing `.partial` file.

There are definitely improvements we need to make for Barman so that it can work better in clustered environments and we are planning on doing some work on this later in the year.

Can you share any more details of the specific problem you are trying to solve with this PR?

mhkarimi1383:
when our clusters (using https://github.com/basalam/kaastolon) fail-over the master node we should run both commands to make backup working, and WAL files are not found from the server and we have to do `reset-wal`

martinmarques:
Are you using some failover VIP or updating DNS records to make barman failover to the current primary node?

mhkarimi1383:
> Are you using some failover VIP or updating DNS records to make barman failover to the current primary node?

We are using connecting Barman to kaastolon Proxy, it will proxy connections to master nodes, when failover happens connections will drop, and upstream node changes

martinmarques:
We will not be moving forward with this PR. I would suggest looking at the models feature we added to the last release which we are using for integration with Patroni

mhkarimi1383:
> We will not be moving forward with this PR. I would suggest looking at the models feature we added to the last release which we are using for integration with Patroni

What about CloudNativePG (It should work) or Stolon (also Kaastolon)

PR #796, opened 2023-05-19
BAR-20 Azure snapshot backups

mikewallace1979:
This PR adds support for the *backup* part of Azure snapshot backups with recovery and delete support to follow in subsequent PRs in tickets BAR-21 and BAR-22.

Most of the commits in the PR are groundwork which prepare the existing code and resolve a few existing issues. The final three commits add the functionality, the unit tests and the documentation.

Once approved the final three commits can be squashed into one, however the groundwork commits are intended to remain separate commits because they address distinct (though related) issues.


barman/cloud_providers/__init__.py
```diff
@@ -69,6 +67,25 @@ def _make_s3_cloud_interface(config, cloud_interface_kwargs):
     return S3CloudInterface(**cloud_interface_kwargs)
 
 
+def _get_azure_credential(credential_type):
```
mikewallace1979, line 70: This will cause `azure.identity` to be imported even if the credential type is `None` which will break barman cloud with Azure Blob Storage in deployments where `azure-identity` is not available - that would be a change in behaviour so we should avoid it.

barman/cloud_providers/google_cloud_storage.py
```diff
@@ -374,12 +374,11 @@ class GcpCloudSnapshotInterface(CloudSnapshotInterface):
...
 
     _required_config_for_backup = CloudSnapshotInterface._required_config_for_backup + (
-        "snapshot_zone",
+        "gcp_zone",
     )
 
     _required_config_for_restore = (
-        CloudSnapshotInterface._required_config_for_restore
-        + ("snapshot_recovery_zone",)
+        CloudSnapshotInterface._required_config_for_restore + ("gcp_zone",)
```
gcalacoci, line 381: we were using 2 different values before, now a single one, is this correct?

gcalacoci (approved):
LGTM

PR #805, opened 2023-06-05
Delete backup_label for barman cloud snapshot backups

mikewallace1979:
Deletes the backup_label from the object store when snapshot backups are deleted. This is necessary because the existing deletion logic only deleted the `backup.info` files and the tarballs containing the backup data (which included the backup label). With snapshot backups, tarballs are not used and the backup label is stored as a unique object. It must therefore be deleted as a special case.

Closes BAR-13.

### DIFF
--- a/barman/clients/cloud_backup_delete.py
+++ b/barman/clients/cloud_backup_delete.py
@@ -185,6 +185,14 @@ def _delete_backup(
                 backup_info, config
             )
             snapshot_interface.delete_snapshot_backup(backup_info)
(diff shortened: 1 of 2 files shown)


mikewallace1979:
I'm going to rewrite this branch since I have made a bit of a mess with fixup commits and no one has started looking at it yet.

mikewallace1979:
> I'm going to rewrite this branch since I have made a bit of a mess with fixup commits and no one has started looking at it yet.

All done - apologies for the extra noise on this PR.

gcalacoci (approved):
LGTM

PR #806, opened 2023-06-07
BAR-9: Fix wal deletion in barman cloud

mikewallace1979:
Related integration tests PR can be found at EnterpriseDB/barman-testing#118

### DIFF
--- a/barman/cloud_providers/aws_s3.py
+++ b/barman/cloud_providers/aws_s3.py
@@ -409,3 +409,37 @@ class S3CloudInterface(CloudInterface):
                     % (error_dict["Key"], error_dict["Code"], error_dict["Message"])
                 )
             raise CloudProviderError()
+
+    def get_prefixes(self, prefix):
+        """
+        Return only the common prefixes under the supplied prefix.
+
+        :param str prefix: The object key prefix under which the common prefixes
+            will be found.
+        :rtype: Iterator[str]
+        :return: A list of unique prefixes immediately under the supplied prefix.
+        """
+        for wal_prefix in self.list_bucket(prefix + "/", delimiter="/"):
+            if wal_prefix.endswith("/"):
+                yield wal_prefix
+
+    def delete_under_prefix(self, prefix):
+        """
+        Delete all objects under the specified prefix.
+
+        :param str prefix: The object key prefix under which all objects should be
+            deleted.
+        """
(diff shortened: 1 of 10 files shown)


tests/test_cloud.py
```diff
@@ -3075,6 +3320,18 @@ def test_parse_backup_id_no_match(self, catalog_with_named_backup):
             exc.value
         )
 
+    def test_get_wal_prefixes(self):
+        """Test we can remove a WAL from the cached list"""
```
mikewallace1979, line 3324: This comment refers to a different test - copy/paste?

tests/test_barman_cloud_backup_delete.py
```diff
@@ -239,32 +245,56 @@ def parse_backup_id(backup_id):
...
         assert len(cloud_interface_mock.delete_objects.call_args_list) == len(
-            expected_calls
+            expected_delete_object_calls
         )
-        for i, call_args in enumerate(expected_calls):
+        for i, call_args in enumerate(expected_delete_object_calls):
             assert (
                 call_args == cloud_interface_mock.delete_objects.call_args_list[i][0][0]
             )
+        # Also verify that delete_under_prefix was never called.
```
mikewallace1979, line 273: This comment should be updated since the tests do now verify the expected call args.

barman/cloud_providers/aws_s3.py
```diff
@@ -407,3 +409,37 @@ def _delete_objects_batch(self, paths):
...
+                yield wal_prefix
+
+    def delete_under_prefix(self, prefix):
+        """
+        Delete all objects under the specified prefix.
+
+        :param str prefix: The object key prefix under which all objects should be
+            deleted.
+        """
+        if len(prefix) == 0 or prefix == "/" or not prefix.endswith("/"):
```
gcalacoci, line 433: I understand the paranoia, so let's keep this. but how can this happen in the real world?
I had to read the error message twice to really understand it, honestly the problem is not that you are unable to safely remove all the files, the problem is that (rightly) removing a full folder is not allowed.
Maybe we can reword the error?

gcalacoci (approved):

barman/cloud_providers/aws_s3.py
mikewallace1979: Rewording makes sense - maybe `Deleting all objects under prefix %s is not allowed`?

PR #812, opened 2023-06-14
Add runbook for recovering an Azure snapshot backup

mikewallace1979:
Closes BAR-31.


doc/runbooks/snapshot_recovery_azure.md
```diff
@@ -0,0 +1,207 @@
+# Recovering snapshot backups on Microsoft Azure
```
mikewallace1979, line 1: Does anything actually link to this runbook?

The google cloud script can be reached from docs.pgbarman.org so I think we should add a link to this runbook in a similar way.

gcalacoci (approved):

PR #822, opened 2023-07-11
Add support for deletion of AWS snapshot backups

mikewallace1979:
Adds support for deletion of AWS snapshot backups by implementing the delete_snapshot_backup function in AwsCloudSnapshotInterface.

Closes BAR-25.

### DIFF
--- a/barman/cloud_providers/aws_s3.py
+++ b/barman/cloud_providers/aws_s3.py
@@ -786,7 +786,9 @@ class AwsCloudSnapshotInterface(CloudSnapshotInterface):
             error_code = exc.response["Error"]["Code"]
             # If the snapshot could not be found then deletion is considered successful
             # otherwise we raise a CloudProviderError
-            if error_code != "InvalidSnapshot.NotFound":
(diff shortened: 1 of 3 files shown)


barman/cloud_providers/aws_s3.py
```diff
@@ -774,13 +774,38 @@ def take_snapshot_backup(self, backup_info, instance_identifier, volumes):
...
+        """
+        Delete the specified snapshot.
+
+        :param str snapshot_id: The ID of the snapshot to be deleted.
+        """
+        try:
+            self.ec2_client.delete_snapshot(SnapshotId=snapshot_id)
+        except ClientError as exc:
+            error_code = exc.response["Error"]["Code"]
+            # If the snapshot could not be found then deletion is considered successful
```
gcalacoci, line 787: Tbh, if not found, we should add a log line at least at debug level, if not info, where we say "snapshot not found/doesn't exists", keeping track of events like this one helps debugging messy situations

gcalacoci (approved):

PR #842, opened 2023-09-12
allow set-version to build doc with docker image

didiermichel:
Changed set-version.sh behavior.

Instead of positional argument it uses flags.
before:
`set-version.sh 3.5.0 20231206`
now:
* `set-version.sh -r 3.5.0 -d 20231206` (identical)
* `set-version.sh -r 3.5.0 -d 20231206 -D` (using docker image)
* `set-version.sh -r 3.5.0 -D` (uses docker image and current date)

Updated release process documentation in [Source code changes](https://enterprisedb.atlassian.net/wiki/spaces/PD/pages/3407282177/release+process+with+foundation#2--Source-code-changes) section.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,3 +1,7 @@
+ifndef DIMAGE
+override DIMAGE =  barman-pandoc
+endif
+
 MANPAGES=barman.1 barman.5 \
 		barman-wal-archive.1 barman-wal-restore.1 \
     barman-cloud-backup.1 \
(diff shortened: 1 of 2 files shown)


doc/Makefile
```diff
@@ -69,6 +69,12 @@ clean:
 		$(MAKE) -C $$dir clean; \
 	done
 
+build-image:
+	docker build . -t ghcr.io/enterprisedb/barman/barman-pandoc:master
```
gcalacoci, line 73: maybe i'm wrong, but ghcr.io/enterprisedb images are not public, not sure this should be on an open-source project... as this will not work for someone that is not part of EDB

doc/Makefile
```diff
@@ -69,6 +69,12 @@ clean:
...
+
+create-all:
+	docker run --rm --volume "`pwd`:/data" -w="/data" --user `id -u`:`id -g` ghcr.io/enterprisedb/barman/barman-pandoc:master make clean all
```
gcalacoci, line 76: same as above

gcalacoci (changes requested):

doc/Makefile
gcalacoci: I don't remember what is inside the internal image, maybe we can make the image public? I think so but not sure yet.

PR #843, opened 2023-09-12
Avoid using internal functions/object with multiprocessing

gcalacoci:
A Multiprocessing start method is the technique used by the Multiprocessing library to start child processes in Python.

There are three start methods:

* spawn: start a new Python process.
* fork: copy a Python process from an existing process.
* forkserver: new process from which future forked processes will be copied.

Each OS has a list of supported start methods:
* Windows (win32): spawn
* MacOS (darwin): spawn (default), fork, forkserver.
* Linux (unix): spawn, fork (default), forkserver.

Spawn not only is the most supported, but as per recent investigations of the devs of the Multiprocessing library, is the most secure on all the OS.

Spawn method makes use of the Pickle library for starting the new subprocesses and Pickle is unable to process internal objects and functions.

Get rid an internal function (handler inside the make_logging_handler method of the Command class) and use a callable object instead.

Consider substituting internal objects and functions with Pickle compliant options all around the code and eventually enforce the usage of the spawn method.

### DIFF
--- a/barman/command_wrappers.py
+++ b/barman/command_wrappers.py
@@ -41,6 +41,22 @@ from barman.exceptions import CommandFailedException, CommandMaxRetryExceeded
 _logger = logging.getLogger(__name__)
 
 
+class Handler:
+    def __init__(self, logger, level, prefix=None):
+        self.class_logger = logger
+        self.level = level
+        self.prefix = prefix
+
+    def run(self, line):
+        if line:
+            if self.prefix:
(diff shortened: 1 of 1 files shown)


mikewallace1979 (approved):
Huge 🙌 for doing the research and experimentation behind this patch.

> Consider substituting internal objects and functions with Pickle compliant options all around the code and eventually enforce the usage of the spawn method.

This sounds like the most sensible way forward and something we need to do in a separate ticket, maybe under the general title of Python 3.12 compatibility.

Issue #844, opened 2023-09-18
A checkpoint can be forced without superuser-privileges since PostgreSQL 14 introduced the `pg_checkpoint` role


toydarian:
Since the predefined role `pg_checkpoint` was introduced in PostgreSQL version 14, superuser-privileges aren't required to run `CHECKPOINT` anymore, if the user has the `pg_checkpoint`-role. (See the [PostgreSQL docs](https://www.postgresql.org/docs/15/predefined-roles.html) for more information.)

To reflect this, barman should not fail to run `switch-wal --force` without superuser-privileges anymore.

toydarian:
implemented in #845

PR #845, opened 2023-09-18
allows to run 'barman switch-wal --force' when the user has the 'pg_checkpoint' role

toydarian:
Since `CHECKPOINT;` can be run by regular users that have the `pg_checkpoint`-role, we can use this to run `switch-wal --force` without superuser-privileges.

I implemented a method that will check if the current user has that role (or is a superuser) for PostgreSQL 14 and above and just checks for superuser-privileges for older versions of PostgreSQL.

Related issue: #844

### DIFF
--- a/tests/test_postgres.py
+++ b/tests/test_postgres.py
@@ -37,7 +37,7 @@ from barman.exceptions import (
     PostgresIsInRecovery,
     BackupFunctionsAccessRequired,
     PostgresObsoleteFeature,
-    PostgresSuperuserRequired,
+    PostgresCheckpointPrivilegesRequired,
     PostgresUnsupportedFeature,
 )
 from barman.postgres import (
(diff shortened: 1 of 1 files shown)


mikewallace1979:
Thanks for filing #844 and making this PR.

I tested locally and this appears to work as advertised - here using PostgreSQL 15 and a non-superuser:

    $ barman switch-wal main --force
    ERROR: Barman switch-wal --force requires superuser rights or the 'pg_checkpoint' role

After granting the `pg_checkpoint` role:

    $ barman switch-wal main --force
    The WAL file 000000010000000100000098 has been closed on server 'main'

The patch itself looks solid however the unit tests will need updating before we can merge it - the test case [`tests.test_postgres.TestPostgres.test_checkpoint`](https://github.com/EnterpriseDB/barman/blob/cb42b10e04afcd7ac727d61704e28078d04fbde8/tests/test_postgres.py#L1067-L1087) is currently failing because it is testing for the old `PostgresSuperuserRequired` exception instead of the new `PostgresCheckpointPrivilegesRequired` exception, and the test case will also need to set `server_version` on the `server.postgres` mock so that the code under test gets an int instead of a mock.

Ideally that test case could be expanded so that it covers both the PG < 14 and the PG >= 14 behaviour.

Are you happy to update this test case to cover the patch?

Thanks again for the contribution - it'll be great to get this merged.

toydarian:
Sorry for breaking unit-tests. I ran `pytest` on `tests`, but it seems it didn't pick up all tests, my bad.  
I fixed the failing test and I wrote an additional one for the `has_checkpoint_privileges` property.

I also ran `tox` this time and it seems there are additional tests failing (in `test_cli.py` and `test_fs.py`), but I don't think those are related to my changes.

mikewallace1979:
> Sorry for breaking unit-tests. I ran `pytest` on `tests`, but it seems it didn't pick up all tests, my bad. I fixed the failing test and I wrote an additional one for the `has_checkpoint_privileges` property.

No worries and thank you for following up with them so quickly.

PR #847, opened 2023-09-19
Update backup compression options that call `pg_basebackup`.

didiermichel:
add `backup_compression=none` to create an uncompressed archive that only supports `backup_compression_level=0`.
for pg >= 15:
update `zstd` compression level range from `-131072` to `22` update `lz4` compression level range from `0` to `12`

For PG < 15:
Allow using `backup_compression_level=0` with `backup_compression=gzip` algorithm.

In configuration 'none' can be used as value and not identified as None value.

### DIFF
--- a/barman/compression.py
+++ b/barman/compression.py
@@ -467,36 +467,41 @@ def get_pg_basebackup_compression(server):
         server.config.backup_compression_location,
         server.config.backup_compression_workers,
     )
-
+    base_backup_compression_option = None
+    compression = None
     if server.config.backup_compression == GZipCompression.name:
         # Create PgBaseBackupCompressionOption
         base_backup_compression_option = GZipPgBaseBackupCompressionOption(
             pg_base_backup_cfg
         )
         compression = GZipCompression(unix_command_factory())
-        return PgBaseBackupCompression(
-            pg_base_backup_cfg, base_backup_compression_option, compression
-        )
+
     if server.config.backup_compression == LZ4Compression.name:
         base_backup_compression_option = LZ4PgBaseBackupCompressionOption(
             pg_base_backup_cfg
         )
         compression = LZ4Compression(unix_command_factory())
-        return PgBaseBackupCompression(
-            pg_base_backup_cfg, base_backup_compression_option, compression
-        )
+
     if server.config.backup_compression == ZSTDCompression.name:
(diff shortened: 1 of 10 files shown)


barman/compression.py
```diff
@@ -574,13 +579,32 @@ def validate(self, pg_server_version, remote_status):
...
-            issues.append(
-                "backup_compression_level %d unsupported by "
-                "pg_basebackup compression %s" % (self.config.level, self.config.type)
+            msg = (
+                "backup_compression_level %d unsupported by compression algorithm."
+                " %s expects a compression level between 1 and 9 (-1 will use default compression level)."
+                % (self.config.level, self.config.type)
             )
+            if self.config.level == 0:
+                msg += "\nIf you need to create an uncompress archive, you should set `backup_compression = none`."
```
gcalacoci, line 606: I would change "uncompress" with "not compressed"

barman/compression.py
```diff
@@ -972,3 +1026,70 @@ def get_file_content(self, filename, archive):
...
+    file_extension = "tar"
+
+    def __init__(self, command):
+        """
+
+        :param command: barman.fs.UnixLocalCommand
+        """
+        self.command = command
+
+    def uncompress(self, src, dst, exclude=None, include_args=None):
```
gcalacoci, line 1042: not sure about this method's name, looks like something that decompress something not that leave stuff not compressed

gcalacoci (changes requested):

barman/compression.py
didiermichel: Agreed: will substitute `an uncompress archive` for `an archive not compressed`

barman/compression.py
didiermichel: That is the implementation of the abstract method inherited from `Compression.uncompress`. It's purpose is to uncompress and extract files from an archive.

There is two way of seeing this, and both make sense to me:
- Maybe a more accurate definition of the function is `extract`(because it represents the final state wanted) . it may or may not imply decompression depending on the compression algorithm -> this implies to rename all uncompress methods with extract 
- Or we can see uncompress for none compression as a special king of compression that does not compress at all. -> in this case nothing to change.

gcalacoci: sorry my fault i got confused following the various methods names around 😄

gcalacoci (approved):
LGTM

PR #857, opened 2023-10-17
Drain queues before joining worker processes

gcalacoci:


mikewallace1979:
Closing because we solved this another way.

PR #861, opened 2023-10-25
barman lockfiles cleanup

gcalacoci:


barman/config.py
```diff
@@ -876,6 +878,9 @@ def _parse_global_config(self):
         self.barman_lock_directory = (
             self.get("barman", "barman_lock_directory") or self.barman_home
         )
+        self.lock_directory_cleanup = self.get(
```
mikewallace1979, line 881: `parse_boolean` is only called on the default and not on the value of `self.get("barman", "lock_directory_cleanup")` which means any value entered in the config is going to be a string rather than a boolean so a config of `lock_directory_cleanup = false` will not currently cause the cleanup to be skipped.

barman/utils.py
```diff
@@ -866,3 +868,37 @@ def get_backup_id_using_shortcut(server, shortcut, BackupInfo):
...
+    This method is part of cron and should help
+    keeping clean the lockfile directory.
+    """
+    if not lock_directory_cleanup:
+        # Auto cleanup of lockfile directory disabled.
+        # Log for debug only and return
+        _logger.debug("Auto-cleanup of '%s' directory disabled" % lock_dir)
+        return
+    _logger.info("Cleaning up lockfiles directory.")
+    for filename in glob("%s/*.lock" % lock_dir):
```
mikewallace1979, line 888: This glob expression doesn't match any dot-prefixed lock files - I think `"%s/.*.lock" % lock_dir` would work.

mikewallace1979 (changes requested):

barman/utils.py
gcalacoci: right, fixed it and also used `os.path.join` to create the path instead than the plain substitution

barman/config.py
gcalacoci: right, should be applied to both `lock_directory_cleanup` and `DEFAULT_CLEANUP`

mikewallace1979 (approved):

PR #864, opened 2023-11-01
Capability of tracking the config file for each configuration option

barthisrael:
This PR adds a way of tracking, for each configuration option, the file that provides the effective configuration value. It is particularly useful if a user happens to define the same configuration option in more than one Barman configuration file.

In order to expose that information to the user, we added a `--show-config-source` flag to `barman diagnose` command, which when specified will change the structure of all `config` keys in the `barman diagnose` command output.

If the flag is omitted, the behavior is the same as previously: it will write a JSON object where each key is the name of the configuration option, and the value is the value of the configuration option.

On the other hand, if the flag is passed, the output format changes a bit: it will write a JSON object where each key is still the name of the configuration option, but the value is not a string but a JSON object with a couple keys -- `source`, pointing to the config file that provides the effective value, and `value`, the effective value.

Previous to this PR Barman used to use `ConfigParser` to handle parsing of the INI configuration files. In order to implement the mapping, we basically wrote a wrapper around that class: a new class named `ConfigMapping` which subclasses `ConfigParser`. That way we keep all the interfaces provided by `ConfigParser` still available to the Barman code. The only restriction is that when using `ConfigMapping` we should use `read_config` new method to parse INI files, which will take care of mapping each configuration option to the corresponding source file when parsing them.

References: BAR-127.

### DIFF
--- a/barman/cli.py
+++ b/barman/cli.py
@@ -1175,7 +1175,16 @@ def check(args):
     output.close_and_exit()
 
 
-@command()
+@command(
+    [
+        argument(
+            "--show-config-source",
+            help="Include the source file which provides the effective value "
+            "for each configuration option",
+            action="store_true",
+        )
+    ],
+)
 def diagnose(args=None):
     """
     Diagnostic command (for support and problems detection purpose)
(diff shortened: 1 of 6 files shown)


barthisrael:
Just squashed the "black format" commits.

PR #867, opened 2023-11-06
[WIP] Allow WAL streaming from a different host to the backup source

mikewallace1979:
This is a work-in-progress for BAR-120 where we add the ability for users to define conninfo strings to be used for streaming WALs. This allows WALs to be streamed from a different host than the host from which backups are streamed.

The implementation is a lot more complicated than I'd like and I feel like there are opportunities for simplifying things which I'm not seeing, so suggestions along these lines are particularly welcome.


mikewallace1979:
Closing in favour of https://github.com/EnterpriseDB/barman/pull/868

PR #868, opened 2023-11-07
Dev/bar 120 attempt 2

mikewallace1979:
This is functionally-equivalent to #867 however the different conninfo strings for WAL streaming are managed before creating the server.

The general idea is that:
1. `barman backup` and most other barman commands create `barman.server.Server` objects using the `conninfo` and `streaming_conninfo` in the configuration as usual.
2. `barman receive-wal` and `barman cron` create `barman.server.Server` objects using `wal_conninfo` and `wal_streaming_conninfo` if they are set but use`conninfo` and `streaming_conninfo` otherwise.
3. `barman replication-status` will use either, depending on the value of its `--source` option (which defaults to the backup conninfo, to preserve existing behaviour).

This works quite nicely however:
1. We need to be sure this is a safe thing to do. Might `barman cron` need to use the backup-host conninfo strings for some reason? Are there other barman commands which should have the WAL conninfo?
2. `barman check` still needs work because it is checking the complete configuration of the Barman server.

Point 2 is addressed by the second commit in this PR where we make `barman check` get the WAL streaming status and carry out the relevant checks using that.

For point 1 we can do some careful review.

Overall this seems cleaner than #867 and does not require the scattergun changes to `barman.server.Server` which are present in that PR.

Related: [Integration tests](https://github.com/EnterpriseDB/barman-testing/pull/142).

### DIFF
--- a/barman/config.py
+++ b/barman/config.py
@@ -572,7 +572,9 @@ class ServerConfig(BaseConfig):
         "streaming_conninfo",
         "streaming_wals_directory",
         "tablespace_bandwidth_limit",
+        "wal_conninfo",
         "wal_retention_policy",
+        "wal_streaming_conninfo",
         "wals_directory",
     ]
 
@@ -920,6 +922,34 @@ class ServerConfig(BaseConfig):
 
         self.msg_list.extend(msg_list)
         self.disabled = True
+
+    def get_wal_conninfo(self):
+        """
+        Return WAL-specific conninfo strings for this server.
+
+        Returns the value of ``wal_streaming_conninfo`` and ``wal_conninfo`` if they
+        are set in the configuration. If ``wal_conninfo`` is unset then it will
+        be given the value of ``wal_streaming_conninfo``. If ``wal_streaming_conninfo``
+        is unset then fall back to ``streaming_conninfo`` and ``conninfo``.
+
+        :rtype: (str,str)
+        :return: Tuple consisting of the ``wal_streaming_conninfo`` and
+            ``wal_conninfo`` defined in the configuration if ``wal_streaming_conninfo``
+            is set, a tuple of ``streaming_conninfo`` and ``conninfo`` otherwise.
+        """
+        wal_streaming_conninfo, wal_conninfo = None, None
+        if self.wal_streaming_conninfo is not None:
+            wal_streaming_conninfo = self.wal_streaming_conninfo
+            if self.wal_conninfo is not None:
+                wal_conninfo = self.wal_conninfo
+            else:
+                wal_conninfo = self.wal_streaming_conninfo
+        else:
+            # If wal_streaming_conninfo is not set then return the original
+            # streaming_conninfo and conninfo parameters
+            wal_streaming_conninfo = self.streaming_conninfo
+            wal_conninfo = self.conninfo
+        return wal_streaming_conninfo, wal_conninfo
 
 
 class ModelConfig(BaseConfig):
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -666,6 +666,42 @@ class PostgreSQLConnection(PostgreSQL):
                 return None
 
     @property
+    def has_monitoring_privileges(self):
+        """
+        Check whether the current user can access monitoring information.
+
+        Returns ``True`` if the current user is a superuser or if the user has the
+        necessary privileges to monitor system status.
+
+        :rtype: bool
+        :return: ``True`` if the current user can access monitoring information.
+        """
+        if self.is_superuser:
+            return True
+        else:
+            monitoring_check_query = """
+            SELECT
+            (
+                pg_has_role(CURRENT_USER, 'pg_monitor', 'MEMBER')
+                OR
+                (
+                    pg_has_role(CURRENT_USER, 'pg_read_all_settings', 'MEMBER')
+                    AND pg_has_role(CURRENT_USER, 'pg_read_all_stats', 'MEMBER')
+                )
+            )
+            """
+            try:
+                cur = self._cursor()
+                cur.execute(monitoring_check_query)
+                return cur.fetchone()[0]
+            except (PostgresConnectionError, psycopg2.Error) as e:
+                _logger.debug(
+                    "Error checking privileges for functions needed for monitoring: %s",
+                    force_str(e).strip(),
+                )
+                return None
+
+    @property
     def current_xlog_info(self):
         """
         Get detailed information about the current WAL position in PostgreSQL.
@@ -938,6 +974,7 @@ class PostgreSQLConnection(PostgreSQL):
 
             result["is_superuser"] = self.is_superuser
             result["has_backup_privileges"] = self.has_backup_privileges
+            result["has_monitoring_privileges"] = self.has_monitoring_privileges
             result["is_in_recovery"] = self.is_in_recovery
             result["server_txt_version"] = self.server_txt_version
             result["version_supported"] = self.is_minimal_postgres_version()
@@ -1393,7 +1430,7 @@ class PostgreSQLConnection(PostgreSQL):
         try:
             cur = self._cursor(cursor_factory=NamedTupleCursor)
 
-            if not self.has_backup_privileges:
+            if not self.has_monitoring_privileges:
                 raise BackupFunctionsAccessRequired(
                     "Postgres user '%s' is missing required privileges "
                     '(see "Preliminary steps" in the Barman manual)'
--- a/barman/server.py
+++ b/barman/server.py
@@ -598,6 +598,7 @@ class Server(RemoteStatusMixin):
                 # Postgres configuration is not available on passive nodes
                 if not self.passive_node:
                     self.check_postgres(check_strategy)
+                    self.check_wal_streaming(check_strategy)
                 # Check barman directories from barman configuration
                 self.check_directories(check_strategy)
                 # Check retention policies
@@ -773,8 +774,29 @@ class Server(RemoteStatusMixin):
                     check="no access to backup functions",
                 )
 
+        self._check_streaming_supported(check_strategy, remote_status)
+
+        self._check_wal_level(check_strategy, remote_status)
+
+        if self.config.primary_conninfo is not None:
+            self._check_standby(check_strategy)
+
+    def _check_streaming_supported(self, check_strategy, remote_status, suffix=None):
+        """
+        Check whether the remote status indicates streaming is possible.
+
+        :param CheckStrategy check_strategy: The strategy for the management
+            of the result of this check
+        :param dict[str, None|str] remote_status: Remote status information used
+            by this check
+        :param str|None suffix: A suffix to be appended to the check name
+        """
         if "streaming_supported" in remote_status:
-            check_strategy.init_check("PostgreSQL streaming")
+            check_name = "PostgreSQL streaming" + (
+                "" if suffix is None else f" ({suffix})"
+            )
+
+            check_strategy.init_check(check_name)
             hint = None
 
             # If a streaming connection is available,
@@ -784,10 +806,22 @@ class Server(RemoteStatusMixin):
             check_strategy.result(
                 self.config.name, remote_status.get("streaming"), hint=hint
             )
+
+    def _check_wal_level(self, check_strategy, remote_status, suffix=None):
+        """
+        Check whether the remote status indicates ``wal_level`` is correct.
+
+        :param CheckStrategy check_strategy: The strategy for the management
+            of the result of this check
+        :param dict[str, None|str] remote_status: Remote status information used
+            by this check
+        :param str|None suffix: A suffix to be appended to the check name
+        """
         # Check wal_level parameter: must be different from 'minimal'
         # the parameter has been introduced in postgres >= 9.0
         if "wal_level" in remote_status:
-            check_strategy.init_check("wal_level")
+            check_name = "wal_level" + ("" if suffix is None else f" ({suffix})")
+            check_strategy.init_check(check_name)
             if remote_status["wal_level"] != "minimal":
                 check_strategy.result(self.config.name, True)
             else:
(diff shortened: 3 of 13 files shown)


barman/postgres.py
```diff
@@ -665,6 +665,43 @@ def has_checkpoint_privileges(self):
                 )
                 return None
 
+    @property
+    def has_monitoring_privileges(self):
+        """
+        Returns true if the current user is a superuser or if the user has the
```
barthisrael, line 671:
```suggestion
        Returns ``True`` if the current user is a superuser or if the user has the
```

(Sphinx recommendation for literal values).

barman/postgres.py
```diff
@@ -665,6 +665,43 @@ def has_checkpoint_privileges(self):
...
+                    AND pg_has_role(CURRENT_USER, 'pg_read_all_stats', 'MEMBER')
+                )
+            )
+            FROM
+            pg_user
+            WHERE
+            usename = CURRENT_USER
+            """.format(
+                **self.name_map
+            )
```
barthisrael, line 693:
```suggestion
            SELECT
            (
                pg_has_role(CURRENT_USER, 'pg_monitor', 'MEMBER')
                OR
                (
                    pg_has_role(CURRENT_USER, 'pg_read_all_settings', 'MEMBER')
                    AND pg_has_role(CURRENT_USER, 'pg_read_all_stats', 'MEMBER')
                )
            )
```

We are not using anything from the `pg_catalog.pg_user` view, so I guess we can remove that part of the query.
Similarly, we don't have anything to interpolate in the string.

barman/server.py
```diff
@@ -518,6 +518,14 @@ def _init_retention_policies(self):
...
                 )
 
+    def init_wal_streaming_config(self):
+        if self.config.wal_streaming_conninfo is not None:
+            self.config.streaming_conninfo = self.config.wal_streaming_conninfo
+            if self.config.wal_conninfo is not None:
+                self.config.conninfo = self.config.wal_conninfo
+            else:
+                self.config.conninfo = self.config.wal_streaming_conninfo
+
```
barthisrael, line 528:
```suggestion
```

Not used anywhere -- at least so far 😆 .

barman/server.py
```diff
@@ -773,8 +782,20 @@ def check_postgres(self, check_strategy):
...
                 )
 
+        self._check_streaming_supported(check_strategy, remote_status)
+
+        self._check_wal_level(check_strategy, remote_status)
+
+        if self.config.primary_conninfo is not None:
+            self._check_standby(check_strategy)
+
+    def _check_streaming_supported(self, check_strategy, remote_status, suffix=None):
```
barthisrael, line 792: Add docstring.

barman/server.py
```diff
@@ -773,8 +782,20 @@ def check_postgres(self, check_strategy):
...
         if "streaming_supported" in remote_status:
-            check_strategy.init_check("PostgreSQL streaming")
+            check_name = "PostgreSQL streaming"
+            if suffix is not None:
+                check_name = f"{check_name} ({suffix})"
```
barthisrael, line 796:
```suggestion
            check_name = "PostgreSQL streaming" + ("" if suffix is None else f" ({suffix})")
```
Maybe we could have a single attribution here?

barman/server.py
```diff
@@ -784,10 +805,15 @@ def check_postgres(self, check_strategy):
             check_strategy.result(
                 self.config.name, remote_status.get("streaming"), hint=hint
             )
+
+    def _check_wal_level(self, check_strategy, remote_status, suffix=None):
```
barthisrael, line 809: Add docstring.

barman/server.py
```diff
@@ -784,10 +805,15 @@ def check_postgres(self, check_strategy):
...
         # Check wal_level parameter: must be different from 'minimal'
         # the parameter has been introduced in postgres >= 9.0
         if "wal_level" in remote_status:
-            check_strategy.init_check("wal_level")
+            check_name = "wal_level"
+            if suffix is not None:
+                check_name = f"{check_name} ({suffix})"
```
barthisrael, line 815:
```suggestion
            check_name = "wal_level" + ("" if suffix is None else f" ({suffix})")
```

Same comment in regards to a single attribution.

barman/server.py
```diff
@@ -797,10 +823,64 @@ def check_postgres(self, check_strategy):
                     hint="please set it to a higher level than 'minimal'",
                 )
 
+    def _check_has_monitoring_privileges(
```
barthisrael, line 826: Add docstring.

barman/server.py
```diff
@@ -797,10 +823,64 @@ def check_postgres(self, check_strategy):
...
+        self, check_strategy, remote_status, suffix=None
+    ):
+        check_name = "has monitoring privileges"
+        if suffix is not None:
+            check_name = f"{check_name} ({suffix})"
```
barthisrael, line 831:
```suggestion
        check_name = "has monitoring privileges" + ("" if suffix is None else f" ({suffix})")
```

Same comment in regards to single attribution.

barman/server.py
```diff
@@ -797,10 +823,64 @@ def check_postgres(self, check_strategy):
...
+        check_strategy.init_check(check_name)
+        if remote_status.get("has_monitoring_privileges"):
+            check_strategy.result(self.config.name, True)
+        else:
+            check_strategy.result(
+                self.config.name,
+                False,
+                hint="privileges for PostgreSQL monitoring functions are "
+                "required (see documentation)",
+                check="no access to monitoring functions",
```
barthisrael, line 841: We need to change the documentation. Currently it only explains about a whole set of privileges which were required by "has_backup_privileges". The docs should be amended to make a separation of the privileges, and explain which ones should be granted because of what reasons / settings.

barman/server.py
```diff
@@ -797,10 +823,64 @@ def check_postgres(self, check_strategy):
...
+            )
+
+    def check_wal_streaming(self, check_strategy):
```
barthisrael, line 844: Add docstring.

barman/server.py
```diff
@@ -797,10 +823,64 @@ def check_postgres(self, check_strategy):
...
+            self.check_replication(check_strategy, remote_status)
+
+    def check_replication(self, check_strategy, remote_status=None, suffix=None):
         # Check the presence and the status of the configured replication slot
         # This check will be skipped if `slot_name` is undefined
         if self.config.slot_name:
-            check_strategy.init_check("replication slot")
+            check_name = "replication slot"
+            if suffix is not None:
+                check_name = f"{check_name} ({suffix})"
```
barthisrael, line 880:
```suggestion
            check_name = "replication slot" + ("" if suffix is None else f" ({suffix})")
```

Same comment in regards to single attribution.

barman/server.py
```diff
@@ -797,10 +823,64 @@ def check_postgres(self, check_strategy):
...
+            )
+            self._check_wal_level(check_strategy, remote_status, "WAL streaming")
+            self.check_identity(check_strategy, remote_status, "WAL streaming")
+            self.check_replication(check_strategy, remote_status, "WAL streaming")
+        else:
+            # Use the status for the existing postgres connections
+            remote_status = self.get_remote_status()
+            self.check_replication(check_strategy, remote_status)
+
+    def check_replication(self, check_strategy, remote_status=None, suffix=None):
```
barthisrael, line 874: Add docstring.

barman/server.py
```diff
@@ -1117,7 +1194,7 @@ def check_archiver_errors(self, check_strategy):
             hint=WalArchiver.summarise_error_files(errors),
         )
 
-    def check_identity(self, check_strategy):
+    def check_identity(self, check_strategy, remote_status=None, suffix=None):
```
barthisrael, line 1197: Add new arguments to docstring.

barman/server.py
```diff
@@ -1126,9 +1203,13 @@ def check_identity(self, check_strategy):
         :param CheckStrategy check_strategy: the strategy for the management
              of the results of the various checks
         """
-        check_strategy.init_check("systemid coherence")
+        check_name = "systemid coherence"
+        if suffix is not None:
+            check_name = f"{check_name} ({suffix})"
```
barthisrael, line 1208:
```suggestion
        check_name = "systemid coherence" = ("" if suffix is None else f" ({suffix})")
```

Single attribution.

barthisrael (commented):
This PR definitely looks better than the previous one.
I'm posting a few thoughts.

barman/server.py
mikewallace1979: Yes - thanks for raising this, we'll need to make sure permissions get covered in the documentation for this change. I'll add a commit with the docs before making this ready for review.

mikewallace1979:
> This PR definitely looks better than the previous one. I'm posting a few thoughts.

Thanks for reviewing the draft PR @barthisrael - I implemented all of your suggestions so far 😄 

There will be a few follow additional commits with unit tests and docs, plus any other changes we determine we need as we continue to think about this patch.

barthisrael (commented):
I really like this PR.
I posted a few suggestions, mostly related with docstring syntax, but also some related with docs, tests, and a possibly alternative approach on the `get_wal_conninfo` function, if you think that's relevant.

doc/manual/23-wal_streaming.en.md
```diff
@@ -104,6 +104,45 @@ create the replication slot by setting:
...
+If an alternative configuration is required, such as backups being sourced from
+a standby with WALs being streamed from the primary, then this can be achieved
+using the following options:
+
+- `wal_streaming_conninfo`: A connection string which Barman will use instead
+   of `streaming_conninfo` when receiving WAL segments via the streaming
+  replication protocol and when checking the status of the replication slot
+  used for receiving WALs.
+- `wal_conninfo`: A connection string which Barman will use instead of
+  `wal_streaming_conninfo` when checking the status of the replication slot
```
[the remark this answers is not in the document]
mikewallace1979: No it is supposed to be `wal_streaming_conninfo` - or at least, that was my intention however maybe we need to rethink it...

The idea is that `wal_streaming_conninfo` is expected to be a conninfo string which can make streaming replication connections and also make SQL connections with monitoring privileges, so if only `conninfo`, `streaming_conninfo` and `wal_streaming_conninfo` are defined, then Barman uses `wal_streaming_conninfo` for receiving WALs via the replication and Barman also uses `wal_streaming_conninfo` for querying the status of the replication slot and other streaming related checks.

For users who do not want one conninfo string having permissions to make streaming replication connections and check monitoring statistics, they can set `wal_conninfo` and this will be used instead of `wal_streaming_conninfo` for checking replication slots and related queries.

This feels over complicated, but if we were to have `wal_conninfo` override `conninfo` then Barman would use `conninfo` with `wal_streaming_conninfo` and the likelihood is these would point to different hosts and therefore give misleading information (and probably fail for reasons like not being able to find the replication slot).

I don't know if we can make it clearer in the docs about how these are intended to be used, or whether perhaps there is a simpler scheme that I am not seeing which would be more intuitive - can you think of a better approach here?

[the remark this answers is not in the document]
mikewallace1979: Maybe this is a bit clearer?

```
- `wal_conninfo`: An optional connection string specifically for monitoring
  WAL streaming status and performing related checks. If set, Barman will use
  this instead of `wal_streaming_conninfo` when checking the status of the
  replication slot.
```

doc/manual/23-wal_streaming.en.md
```diff
@@ -104,6 +104,45 @@ create the replication slot by setting:
...
+  used for receiving WALs.
+
+The following restrictions apply and are enforced by Barman during checks:
+
+- Connections defined by `wal_streaming_conninfo` and `wal_conninfo` must reach
+  a PostgreSQL instance which belongs to the same cluster reached by the
+  `streaming_conninfo` and `conninfo` connections.
+- The `wal_streaming_conninfo` connection string must be able to create
+  streaming replication connections. If `wal_conninfo` is unset it must also
```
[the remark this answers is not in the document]
mikewallace1979: No it's genuinely ambiguous and sloppy writing on my part. I'll update it.

doc/manual/23-wal_streaming.en.md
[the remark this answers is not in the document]
barthisrael: Oh, that was an oversight from my side. IAC I guess your new descriptions is indeed better. Thanks for expanding!

doc/manual/23-wal_streaming.en.md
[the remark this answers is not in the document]
mikewallace1979: Ok I've tried to address this and the point below with the following rewrite:

```
- The `wal_streaming_conninfo` connection string must be able to create
  streaming replication connections.
- Either `wal_streaming_conninfo` *or* `wal_conninfo` (if it is set) must have
  sufficient permissions to read settings and check replication slot status.
  The required permissions are one of:
    - The `pg_monitor` role.
    - Both the `pg_read_all_settings` and `pg_read_all_stats` roles.
    - The `superuser` role.
```

barman/server.py
```diff
@@ -797,10 +831,109 @@ def check_postgres(self, check_strategy):
...
+        :param str|None suffix: A suffix to be appended to the check name
+        """
         # Check the presence and the status of the configured replication slot
         # This check will be skipped if `slot_name` is undefined
         if self.config.slot_name:
-            check_strategy.init_check("replication slot")
+            check_name = "replication slot" + ("" if suffix is None else f" ({suffix})")
+            check_strategy.init_check(check_name)
+            if remote_status is None:
+                remote_status = self.get_remote_status()
```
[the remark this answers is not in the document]
mikewallace1979: Ok I checked the history and this logic wasn't even present in the original so we can indeed remove this. I'll also remove the `None` default since omitting `remote_status` is not a valid way of calling this function.

barman/server.py
barthisrael, line 1: I was wondering: shouldn't we check the WAL streaming connection on `cron_receive_wal` before spawning `pg_receivewal`?

I mean, something similar to what we do in `backup` method when it calls the `check` method. Maybe `check_wal_streaming` could be enough for `cron_receive_wal`.

What do you think?

barthisrael (commented):
Mostly docstring suggestions. Although I have a question which might be relevant regarding "check" on WAL streaming.

barthisrael (changes requested):
Small suggestions based on the work performed through this commit in the docs PR: https://github.com/EnterpriseDB/barman/pull/880/commits/c84dbb42e7248f0e20644e208dece5a92ca96420.

gcalacoci (approved):

barthisrael (approved):
Posted only a small side comment about a docstring, but the PR looks great to me :)

barman/config.py
```diff
@@ -921,6 +923,34 @@ def update_msg_list_and_disable_server(self, msg_list):
...
         self.disabled = True
 
+    def get_wal_conninfo(self):
+        """
+        Return WAL-specific conninfo strings for this server.
+
+        Returns the value of ``wal_streaming_conninfo`` and ``wal_conninfo`` if they
+        are set in the configuration. If ``wal_conninfo`` is unset then it will
+        be given the value of ``wal_streaming_conninfo``. If ``wal_streaming_conninfo``
+        is unset then ``None`` will be returned for both values.
```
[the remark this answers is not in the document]
mikewallace1979: Oops - good catch!

PR #871, opened 2023-11-24
Remove full-depth traversal - only list items in the requested 'directory'

sjuls:
Currently the azure container storage implementation of list_bucket does a depth-first traversal of the blob directory while the aws s3 implementation only lists the items in the target directory. Since doing the full traversal can be _VERY_ time-consuming when a lot of wal segments are archived in the bucket this degrades recovery operations.

We are currently experiencing 10 minute download times for a 300 byte .history file because the tree walk lists all wal segments on all timelines.

This PR removes the recursive calls and only returns first-level content similar to aws s3 implementation.

### DIFF
--- a/barman/cloud_providers/azure_blob_storage.py
+++ b/barman/cloud_providers/azure_blob_storage.py
@@ -44,7 +44,6 @@ except ImportError:
 
 try:
     from azure.storage.blob import (
-        BlobPrefix,
         ContainerClient,
         PartialBatchErrorException,
     )
--- a/barman/cloud_providers/azure_blob_storage.py
+++ b/barman/cloud_providers/azure_blob_storage.py
@@ -290,26 +290,6 @@ class AzureCloudInterface(CloudInterface):
         # the storage account level in Azure)
         self.container_client.create_container()
 
-    def _walk_blob_tree(self, obj, ignore=None):
-        """
-        Walk a blob tree in a directory manner and return a list of directories
-        and files.
-
-        :param ItemPaged[BlobProperties] obj: Iterable response of BlobProperties
-          obtained from ContainerClient.walk_blobs
-        :param str|None ignore: An entry to be excluded from the returned list,
-          typically the top level prefix
-        :return: List of objects and directories in the tree
-        :rtype: List[str]
-        """
-        if obj.name != ignore:
-            yield obj.name
-        if isinstance(obj, BlobPrefix):
-            # We are a prefix and not a leaf so iterate children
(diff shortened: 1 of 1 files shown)


sjuls:
Hi @mikewallace1979,

Any chance I can get a review of this PR? This should bring list_bucket implementation of azure blob storage in-line with the aws s3 implementation so it should be safe.

Happy to make any change you deem necessary.

mikewallace1979:
Hi @sjuls - thanks for doing the analysis and proposing a fix, we will take a closer look this week.

mikewallace1979 (changes requested):
@sjuls I think your proposed fix is correct.

Recursing through the "directories" when fetching the `.history` file occurs during the [`download_wal`](https://github.com/sjuls/barman/blob/8ee2ab25c6f38bba057cd837f58fc530a3498802/barman/clients/cloud_walrestore.py#L142) function where the bucket prefix will be the `wals` directory of the server and the `.history` files will be directly under it. Recursing through the common prefixes containing the WAL segments themselves is indeed completely unnecessary here, since when a WAL *is* required from a prefix Barman will have already determined the correct path under which to look.

I checked the other calls to `list_bucket` to verify nothing depends on recursing through the directories. The following places do need to iterate through all the objects under the prefix, however they call `list_bucket` with no delimiter and therefore will get all the object keys:

* https://github.com/sjuls/barman/blob/fix-az-list/barman/annotations.py#L159-L161
* https://github.com/sjuls/barman/blob/8ee2ab25c6f38bba057cd837f58fc530a3498802/barman/cloud.py#L2162-L2164

The remaining calls to `list_bucket` (in `get_backup_list` and `get_backup_files`) are where the default delimiter of `/` is used but the code has already determined the correct prefix under which the target objects will be:

* https://github.com/sjuls/barman/blob/8ee2ab25c6f38bba057cd837f58fc530a3498802/barman/cloud.py#L2257
* https://github.com/sjuls/barman/blob/8ee2ab25c6f38bba057cd837f58fc530a3498802/barman/cloud.py#L2122

If you could add a commit removing the now unused import of `BlobPrefix` on [L47](https://github.com/sjuls/barman/blob/fix-az-list/barman/cloud_providers/azure_blob_storage.py#L47) then I think we should be able to merge this PR.

sjuls:
@mikewallace1979 Ah right, good catch. I've added a commit to remove the import.

Thanks for the review 🙏

mikewallace1979 (approved):

PR #874, opened 2023-11-29
Dev/bar 128

gcalacoci:


barman/utils.py
```diff
@@ -902,3 +902,39 @@ def lock_files_cleanup(lock_dir, lock_directory_cleanup):
...
+    :param option: the config key to edit or add
+    :param value: the value for the config key to update or add
+    """
+    conf_section = False
+    idx = 0
+    with open(file, "r") as config:
+        lines = config.readlines()
+        eof = len(lines) - 1
+        for idx, line in enumerate(lines):
+            if conf_section and line.startswith("["):  # next section
```
barthisrael, line 925:
```suggestion
            if conf_section and line.strip().startswith("["):  # next section
```

`ConfigParser` doesn't care about spaces around the brackets:

```python
>>> from io import StringIO
>>> from configparser import ConfigParser
>>> parser = ConfigParser(strict=False)
>>> string = StringIO("""
...         [barman]
... some_option = some_value
...
... """)
>>> parser.read_file(string)
>>> parser.sections()
['barman']
```

barman/utils.py
```diff
@@ -902,3 +902,39 @@ def lock_files_cleanup(lock_dir, lock_directory_cleanup):
...
+                lines.insert(idx - 1, option + " = " + value)
+                break
+            elif conf_section and line.startswith(option + " = "):
```
barthisrael, line 928:
```suggestion
            elif conf_section and line.replace(" ", "").startswith(option + "="):
```

Remove all spaces, just in case something weird like this comes up:

```python
line = "   some_option          =some_value"
```

barman/utils.py
```diff
@@ -902,3 +902,39 @@ def lock_files_cleanup(lock_dir, lock_directory_cleanup):
                 _logger.debug(
                     "%s file lock already acquired, skipping removal" % filename
                 )
+
+
+def edit_config(file, section, option, value):
```
barthisrael, line 907: I'd say we could change this method to receive a dict or a list of tuples, so we can update several options in a batch, and only after that call the open function and write method.

barthisrael (commented):
Looks very promising. Just saving a few comments.

barman/config.py
```diff
@@ -1314,6 +1320,228 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+
+        Generates configuration change objects starting from a dictionary with
+        the same fields.
+
+        """
+        if set(obj.keys()) == set(cls._fields):
+            return cls(**obj)
+        if set(obj.keys()) == set(ConfigChange._fields):
+            return ConfigChange(**obj)
+        raise ValueError("Malformed configuration change serialization: %r" % obj)
```
barthisrael, line 1402:
```suggestion
class BaseChange:

    _fields = []

    def __eq__(self, other):
        """Equality support."""
        if isinstance(other, self.__class__):
            return self.as_tuple() == other.as_tuple()
        return False

    def __hash__(self):
        """Hash/set support."""
        return hash(self.as_tuple())

    def as_tuple(self) -> tuple:
        """Convert to a tuple, ordered as self._fields."""
        return tuple(vars(self)[k] for k in self._fields)

    def as_dict(self):
        return {k: vars(self)[k] for k in self._fields}


class ConfigChange(BaseChange):
    """Identifies a configuration change action received"""

    _fields = ["key", "value", "config_file"]

    def __init__(self, key, value, config_file=None):
        """Representation of a configuration change received"""
        self.key = key
        self.value = value
        self.config_file = config_file

    @classmethod
    def from_dict(cls, obj):
        """
        Factory for configuration change objects.

        Generates configuration change objects starting from a dictionary with
        the same fields.

        """
        if set(obj.keys()) == set(cls._fields):
            return cls(**obj)
        raise ValueError("Malformed configuration change serialization: %r" % obj)


class ConfigChangeSet(BaseChange):

    _fields = ["section", "changes_set"]

    def __init__(self, section, changes_set=[]):
        self.section = section
        self.changes_set = changes_set

    @classmethod
    def from_dict(cls, obj):
        """
        Factory for configuration change objects.

        Generates configuration change objects starting from a dictionary with
        the same fields.

        """
        if set(obj.keys()) == set(cls._fields):
            return cls(**obj)
        if set(obj.keys()) == set(ConfigChange._fields):
            return ConfigChange(**obj)
        raise ValueError("Malformed configuration change serialization: %r" % obj)
```

Maybe we could put the common stuff in a base class and inherit from it on both `ConfigChange` and `ConfigChangeSet`. What do you think?

I was also wondering that as `ConfigChangeSet.from_dict` can return either a `ConfigChange` or a `ConfigChangeSet`, maybe we could remove the `from_dict` methods and create a factory class with a `__new__` method instead? We did something like that on Patroni when implementing some stuff for EPAS support, e.g.: https://github.com/zalando/patroni/blob/efdedc7049527117b35849dc5c01525d7f1133e8/patroni/postgresql/validator.py#L169-L232

barthisrael (commented):
I'm putting a few comments I had so far.

As Giulio mentioned he intends to perform some changes next week, I'll wait for the changes before going deeper in the review.

barman/cli.py
```diff
@@ -1810,6 +1814,32 @@ def check_wal_archive(args):
             output.close_and_exit()
 
 
+@command(
+    [
+        argument(
+            "server_name",
```
mikewallace1979, line 1820: It definitely feels weird specifying the `server_name` here but then also specifying the server name as a key in `json_changes` and my instinct is to remove this parameter altogether and just rely on the keys supplied in the json - if they map to a section in the config file then changes are applied there, otherwise it's an error.

This would allow things like `["barman": {...}}` to update the global config, which would help Barman play nicely with external automation beyond the current HA scope.

mikewallace1979: I was thinking about alternative formats for `json_data` which might make bulk changes easier - something like:

```
[
  {
    "scope": "global",
    "log_level": "DEBUG"
  },
  {
    "scope": "server",
    "server_name": "main",
    "conninfo": "db=main.local"
  },
  {
    "scope": "model",
    "server_name": "main",
    "model_name": "main-failover",
    "conninfo": "db=failover.local"
  },
  {
    "scope": "model",
    "server_name": "main",
    "model_name": "all",
    "primary_conninfo": "main-failover"
  },
]
```

The proposal being:
- The JSON body is a list where each item is a set of config changes.
- `scope`, `server_name` and `model_name` are special keys which define the scope (global/server/model) and the server and model (if required by the scope).
- `model_name` supports a special value `all` which causes the change to be applied to all models for a server.

This would need further refinement but as a general idea does it make (any) sense? It's a step away from mapping input JSON directly to sections but would allow more flexibility in how external actors specify changes.

mikewallace1979 (commented):
Overall this PR looks reasonable given the amount of flexibility it gives to the user. I left a few comments and proposed an alternative format for the input JSON.

The most obvious gap I can see is in validation and error handling which makes sense since this is an early iteration of the feature - eventually there'll be some way of handling changes in the queue which can't be applied.

barman/config.py
```diff
@@ -1314,6 +1320,228 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+
+        # If there are no configuration change we've nothing to do here
+        if len(changes_list) == 0:
+            _logger.debug("No valid changes submitted")
+            return
+
+        with ConfigChangesQueue(self.config.config_changes_queue) as changes_queue:
+            changes_queue.queue.extend(changes_list)
+
+    def process_conf_changes_queue(self):
```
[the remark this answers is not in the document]
barthisrael: I think that makes sense, yes.

barman/cli.py
barthisrael: I do like that idea. The only thing that sounded weird to me at a first glance is the models part:

```
  {
    "scope": "model",
    "server_name": "main",
    "model_name": "main-failover",
    "conninfo": "db=failover.local"
  },
  ```
  
  I think we should not "tie" models with servers given both servers and models belong to clusters. IMHO we should use the 3 scopes (either global, server or scope) and do not mix `server_name` and `model_name`. Thoughts?

Comment (barthisrael) [reply barman/cli.py:1820]: Some side thought too: We will likely "get rid" of this command before merging, right? I mean, there are some separate issues in the board to handle the commands -- unless we decide referencing those through this PR and shipping all together.

File: barman/config.py (hunk 4 above)
Comment (barthisrael) [reply barman/config.py:1402]: You mentioned something earlier in the call about the number of classes.
Maybe we can get rid of these `ConfigChange` and `ConfigChangeSet` classes if we just respect some "dict structure" when passing things to the queue. Not sure if that's worth it or not, just thinking out loud.

File: barman/config.py [hunk 7]
```diff
@@ -1314,6 +1320,228 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        """Reads a json file containing a list of configuration changes."""
+        try:
+            with open(path, "r") as queue_file:
+                # Read the queue if exists
+                return json.load(queue_file, object_hook=ConfigChangeSet.from_dict)
+        except FileNotFoundError:
+            return []
+
+    def __enter__(self):
+        return self
```
barthisrael, line 1428: Maybe we should call `self.open()` here instead of on `__init__`?

barman/config.py
```diff
@@ -1314,6 +1320,228 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        # Get all the available configuration change files in order
+        changes_list = []
+        cfg_json = json.loads(changes)
+        for section in cfg_json:
+            set = ConfigChangeSet(section)
+            for json_cng in cfg_json[section]:
+                chg = ConfigChange(
+                    json_cng,
+                    cfg_json[section][json_cng],
+                    self.config.config._config.get_config_source(section, json_cng),
```
barthisrael, line 1484: side comment: `self.config.config._config` looks so weird 😆

barman/utils.py
```diff
@@ -902,3 +902,40 @@ def lock_files_cleanup(lock_dir, lock_directory_cleanup):
...
+        with open(file, "r") as config:
+            lines = config.readlines()
+    eof = len(lines) - 1
+    for idx, line in enumerate(lines):
+        if conf_section and line.startswith("["):  # next section
+            lines.insert(idx - 1, option + " = " + value)
+            break
+        elif conf_section and line.startswith(option + " = "):
+            lines.pop(idx)
+            lines.insert(idx, option + " = " + value + "\n")
```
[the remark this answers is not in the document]
gcalacoci: so, ok about normalizing, added it to the code.
about using a single op to overwrite, a bit of time passed from when I tested it... but basically the pop and add was the consistent method to do this... I'll keep this spil for now, eventually changing with a the on line substitution you suggested

barman/utils.py
```diff
@@ -902,3 +902,40 @@ def lock_files_cleanup(lock_dir, lock_directory_cleanup):
...
+            break
+        elif conf_section and idx == eof:
+            lines.insert(idx + 1, option + " = " + value + "\n")
```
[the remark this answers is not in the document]
gcalacoci: yeah changed in my code, will push soon

barman/config.py
```diff
@@ -1314,6 +1320,228 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+class ConfigChangeSetEncoder(json.JSONEncoder):
+    """Encode a configuration change as a dictionary."""
+
+    def default(self, obj):
+        if isinstance(obj, ConfigChangeSet):
+            return dict(obj.as_dict())
+        # Let the base class default method raise the TypeError
+        if isinstance(obj, ConfigChange):
+            return dict(obj.as_dict())
+        return super().default(obj)
```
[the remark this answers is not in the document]
gcalacoci: yeah, I forgot that isisntance accepts a tuple as a second arg... :)

barman/config.py
gcalacoci: what I mentioned is that I was sort of searching for an elegant way to reduce the amount of classes involved in this. in the end a ConfigChangeSet could do everything if we find a decent way to handle the change without using an obj.
BUT, tbh I'm not sure ATM is worth investing in trying to reduce the amount of classes...
I'm probably going with your baseclass suggestion, at least we get rid of duplicated code

barman/config.py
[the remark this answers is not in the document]
gcalacoci: not sure if we want cron to be involved... as we cannot assume that users are running it every minute as suggested in the doc..
take a scenario where it is executed every 15 mins... I don't think is an acceptable delay between a submitted config and an applied one...

barman/cli.py
gcalacoci: yeah i do agree with this and is something I was thinking myself... I need to check how to correctly handle the server loading without putting much weight in the cli module. cycling on this atm

barman/config.py
gcalacoci: init ensure the loading at creation class, enter execution happens later in the sequence of actions...
tbh the sooner we load it the better...

barman/cli.py
gcalacoci: i'm not sure we want this command to be able to write global/barman-section configurations...
besides this... i was thinking around the scope etc... well if we put aside global for a moment... we have the `server` and `model` scope.. but in the end a `model` is a "normal" config + `model = true` so my question is... does it need a scope? @mikewallace1979 @barthisrael

gcalacoci: keeping the scope around for now... what do you think of this format:
```
{
    "main": { 
        "scope": "server", 
        "conninfo": "db=main.local" }, 
    "main": {
        "scope": "model",
        "model_name": "main-failover", 
        "conninfo": "db=failover.local" }, 
    "main": { 
        "scope": "model", "model_name": "all", 
        "primary_conninfo": "main-failover" 
    }
}
```

In terms of readability and order using the servername as identifier sorta makes sense...
OTOH, the above example seems exceptionally weird as were have 3 "main" labels (and I'm not even sure is ok in terms of json)... open to discussions @barthisrael @mikewallace1979

barman/config.py
barthisrael: I see. Yeah, my initial thought was to get rid of duplicated code. But I also understand your concerns about having several classes to do "one thing" 😆

barman/config.py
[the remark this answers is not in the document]
barthisrael: I see. At a first glance `barman cron` would be the best place to put this. I mean, that's what starts Barman routines. Would you have a better idea of how to detect and retry that "apply" operation other than `barman cron`?

barman/cli.py
barthisrael: I wonder if we should use a list of dicts, instead of a dict of dicts for the payload.
I mean, something like this:

```
[
    {
        "scope": "server",
        "server_name": "main",
        "conninfo": "db=main.local"
    }, 
    {
        "scope": "model",
        "model_name": "main-failover", 
        "conninfo": "db=failover.local"
    }, 
    { 
        "scope": "model",
        "model_name": "all", 
        "primary_conninfo": "main-failover" 
    }
]
```

I mean, it sounds more aligned with what we want to do. Also, my understanding is that models are tied to clusters, and servers are tied to clusters, but models are not tied to servers. So I guess the model specification in your example would sound a bit weird in that case.

What do you think?

barman/config.py
barthisrael: I see. The sooner we can detect issues too in that case.

barman/config.py
barthisrael: Heh, me neither!

barman/config.py
```diff
@@ -1362,30 +1366,13 @@ def from_dict(cls, obj):
...
-
-    def as_tuple(self):
-        """Convert to a tuple, ordered as self._fields."""
-        return tuple(vars(self)[k] for k in self._fields)
-
-    def as_dict(self):
-        return {k: vars(self)[k] for k in self._fields}
-
     @classmethod
     def from_dict(cls, obj):
```
barthisrael, line 1377: I guess we can also move this one to the base class

barman/config.py
gcalacoci: >Maybe ~/.barman.conf could be a good place, as suggested by Giulio.

I did various tests and in the end the simplest solution is: if is default dump everything in ~/.barman.conf

We just need to document it. 
IF we cannot write in the config file (ie config file in /etc/ with just read rights) we want to die horribly until the rights are updated... opinions on this @barthisrael @mikewallace1979 ?

barthisrael: I guess that makes sense, yes.
I mean, files under `/etc` have higher priority than `~/.barman.conf`, so we need to let the user know if we fail to update an `/etc` file as writing to `~/.barman.conf` would not take effect in that case because of the setting being defined in these two places.

barthisrael (changes requested):
I'll leave some comments about thoughts that I had and issues that I faced while testing this draft code:

* `config_changes_queue` seems to fit best as global instead of server setting
* If you send an invalid JSON through `config-update`, it fails to serialize and a `/var/lib/barman/cfg_changes.queue.tmp` is left over, preventing newer executions of the command because of:

```
EXCEPTION: [Errno 17] File exists: '/var/lib/barman/cfg_changes.queue.tmp'
```

* JSON encoding currently fails. We can solve with:

```diff
$ git diff barman/config.py
diff --git a/barman/config.py b/barman/config.py
index 1feaf6b..8c16695 100644
--- a/barman/config.py
+++ b/barman/config.py
@@ -1538,7 +1538,7 @@ class ConfigChangeSetEncoder(json.JSONEncoder):
     """Encode a configuration change as a dictionary."""

     def default(self, obj):
-        if isinstance(obj, ConfigChangeSet):
+        if isinstance(obj, (ConfigChange, ConfigChangeSet)):
             # Let the base class default method raise the TypeError
             return dict(obj.as_dict())
         return super().default(obj)
```

* The following debug message will not work, and will log errors because of `json.dumps` failures -- we need to set the encoder:

```python
                    try:
                        self.apply_change(chgs_set)
                    except Exception as e:
                        # Log that something went horribly wrong and re-raise
                        msg = "Unable to process a set of changes. Exiting."
                        output.error(msg)
                        _logger.debug(
                            "Error while processing %s. \nError: %s"
                            % json.dumps(chgs_set),
                            e,
                        )
                        raise e
```

errors:

```
2024-01-10 19:24:46,781 [5469] barman.config ERROR: Unable to process a set of changes. Exiting.
2024-01-10 19:24:46,782 [5469] barman.config ERROR: Cannot execute <barman.config.ConfigChangeSet object at 0x7f332bd026a0>: Object of type 'ConfigChangeSet' is not JSON serializable
```

* `edit_config` fails if the file doesn't exist, e.g. when we are sending changing "defaults" and attempting to use a non-existing `~/.barman.conf` as target:

```python
    if lines is None:
        with open(file, "r") as config:
            lines = config.readlines()
```

barman/config.py
```diff
@@ -1314,6 +1323,234 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+            section_name = section["server_name"]
+            set = ConfigChangeSet(section_name)
+            for json_cng in section:
+                if json_cng in ("server_name", "scope"):
+                    continue
+                file_name = self.config._config.get_config_source(
+                    section_name, json_cng
+                )
+                if file_name == "default":
+                    file_name = "~/.barman.conf"
```
barthisrael, line 1477:
```suggestion
                    file_name = os.path.expanduser("~/.barman.conf")
```

We need to perform the expansion through the `os` module, otherwise it tries to read/write a bogus file.

barthisrael (changes requested):
Besides the changes that have been suggested, I found the following issue upon executing the command twice:

```bash
[barman@patroni-barman-api-barman ~]$ barman config-update '[{"archiver": "off", "create_slot": "manual", "server_name": "patroni-barman-api-pg-1"}]'
[barman@patroni-barman-api-barman ~]$ barman config-update '[{"archiver": "off", "create_slot": "manual", "server_name": "patroni-barman-api-pg-1"}]'
EXCEPTION: No section: 'barman'
See log file for more details.
No section: 'barman'
See log file for more details.
  File "/usr/lib64/python3.6/configparser.py", line 846, in items
    d.update(self._sections[section])
KeyError: 'barman'

During handling of the above exception, another exception occurred:

  File "/var/lib/barman/barman/barman/cli.py", line 2172, in main
    global_config(args)
  File "/var/lib/barman/barman/barman/cli.py", line 1875, in global_config
    config = barman.config.Config(filename)
  File "/var/lib/barman/barman/barman/config.py", line 976, in __init__
    self._parse_global_config()
  File "/var/lib/barman/barman/barman/config.py", line 1024, in _parse_global_config
    self._global_config = set(self._config.items("barman"))
  File "/usr/lib64/python3.6/configparser.py", line 849, in items
    raise NoSectionError(section)
configparser.NoSectionError: No section: 'barman'
```

When I took a look at the config parse logic in Barman I had the impression it would apply one file on top of another, but in practice it seems to be ignoring my `/etc/barman.conf` once a `~/.barman.conf` is created 🤔

barthisrael:
I didn't notice at a first moment the `break` statement here:

```python
            for path in self.CONFIG_FILES:
                full_path = os.path.expanduser(path)
                if os.path.exists(full_path) and full_path in self._config.read_config(
                    full_path
                ):
                    filename = full_path
                    break
```

That defeats the purpose of using `~/.barman.conf` :(

gcalacoci:
> I didn't notice at a first moment the `break` statement here:
> 
> ```python
>             for path in self.CONFIG_FILES:
>                 full_path = os.path.expanduser(path)
>                 if os.path.exists(full_path) and full_path in self._config.read_config(
>                     full_path
>                 ):
>                     filename = full_path
>                     break
> ```
> 
> That defeats the purpose of using `~/.barman.conf` :(

yeah... and this doesn't happen because of this patch, but has always been like this... at this point maybe we want a new file for this

gcalacoci:
> > I didn't notice at a first moment the `break` statement here:
> > ```python
> >             for path in self.CONFIG_FILES:
> >                 full_path = os.path.expanduser(path)
> >                 if os.path.exists(full_path) and full_path in self._config.read_config(
> >                     full_path
> >                 ):
> >                     filename = full_path
> >                     break
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That defeats the purpose of using `~/.barman.conf` :(
> 
> yeah... and this doesn't happen because of this patch, but has always been like this... at this point maybe we want a new file for this

just to be clear... what I mean is we create a new file with a new name, like `.defaults-barman.conf` or something like this... and we load it like we load the config file into the config dir...

I think I have a patch in mind, I'll use the `.defaults-barman.conf` name for now... but I don't like it so I'm totally open to any suggestion 😄

barman/utils.py
martinmarques: What happens if `some_value` has valid white spaces in it?

barthisrael: Oh, this is only in the `if` statement, and only for the parameter name.
We are interested in checking if the line "starts" with `OPTION_NAME=`, `OPTION_NAME` being the setting that we want to change the value.

barman/cli.py
```diff
@@ -1871,6 +1882,26 @@ def config_switch(args):
...
+            help="specifies the configuration changes to apply, in json format ",
+        ),
+    ]
+)
+def config_update(args):
+    """
+    Receives a set of configuration changes in json format and applies them.
+    """
+    json_changes = json.loads(args.json_changes)
+    # this stops the execution of multiple execitions of the config-update command
```
barthisrael, line 1898:
```suggestion
    # this prevents multiple concurrent executions of the config-update command
```

barman/config.py
```diff
@@ -1326,21 +1333,24 @@ def load_configuration_files_directory(self):
...
-                        "barman] section" % filename
-                    )
-                    _logger.fatal(msg)
-                    raise SystemExit("FATAL: %s" % msg)
-            else:
-                # Add an info that a file has been discarded
-                _logger.warn("Discarding configuration file: %s (not a file)", filename)
+            self.load_config_file(cfile)
+
+    def load_config_file(self, cfile):
```
barthisrael, line 1338:
```suggestion
    def load_config_file(self, cfile):
        """Load and merge configuration from *cfile*.
        
        .. note::
            *cfile* is not expected to contain Barman global settings. If that is the case, then quit with a fatal message.
        
        :param cfile: path to a configuration file to be loaded and merged into the already loaded config.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+    Provides methods for equality comparison, hashing, and conversion
+    to tuple and dictionary.
+    """
+
+    _fields = []
+
+    def __eq__(self, other):
+        """
+        Equality support.
+        """
```
barthisrael, line 1703:
```suggestion
        """
        Equality support.
        
        :param other: other object to compare this one against.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        if isinstance(other, self.__class__):
+            return self.as_tuple() == other.as_tuple()
+        return False
+
+    def __hash__(self):
+        """
+        Hash/set support.
+        """
```
barthisrael, line 1711:
```suggestion
        """
        Hash/set support.
        
        :return: a hash of the tuple created though :meth:`as_tuple`.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        return hash(self.as_tuple())
+
+    def as_tuple(self) -> tuple:
+        """
+        Convert to a tuple, ordered as self._fields.
+        """
```
barthisrael, line 1717:
```suggestion
        """
        Convert to a tuple, ordered as :attr:`_fields`.
        
        :return: tuple of values for :attr:`_fields`.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        return tuple(vars(self)[k] for k in self._fields)
+
+    def as_dict(self):
+        """
+        Convert to a dictionary, using self._fields as keys.
+        """
```
barthisrael, line 1723:
```suggestion
        """
        Convert to a dictionary, using :attr:`_fields` as keys.
        
        :return: a dictionary where keys are taken from :attr:`_fields` and values are the corresponding values for those fields.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+
+
+class ConfigChange(BaseChange):
+    """
+    Represents a configuration change received.
+
+    Attributes:
+        key (str): The key of the configuration change.
+        value (str): The value of the configuration change.
+        config_file (str, optional): The configuration file associated with the change.
```
barthisrael, line 1734:
```suggestion
    :ivar key str: The key of the configuration change.
    :ivar value str: The value of the configuration change.
    :ivar config_file Optional[str]: The configuration file associated with the change, or ``None``.
```

Use reST, which is used through the code, instead of Google style docstrings.

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+    """
+
+    _fields = ["key", "value", "config_file"]
+
+    def __init__(self, key, value, config_file=None):
+        """
+        Initialize a ConfigChange object.
+        """
```
barthisrael, line 1742:
```suggestion
        """
        Initialize a :class:`ConfigChange` object.
        
        :param key str: the configuration setting to be changed.
        :param value str: the new configuration value.
        :param config_file Optional[str]: configuration file associated with the change, if any, or ``None``.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+    @classmethod
+    def from_dict(cls, obj):
+        """
+        Factory method for creating ConfigChange objects from a dictionary.
+
+        :param obj: Dictionary representing the configuration change.
+        :type obj: dict
+        :return: Configuration change object.
+        :rtype: ConfigChange
+        :raises ValueError: If the dictionary is malformed.
```
barthisrael, line 1756:
```suggestion
        Factory method for creating :class:`ConfigChange` objects from a dictionary.

        :param obj: Dictionary representing the configuration change.
        :type obj: :class:`dict`
        :return: Configuration change object.
        :rtype: :class:`ConfigChange`
        :raises:
            :exc:`ValueError`: If the dictionary is malformed.
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        """
+        if set(obj.keys()) == set(cls._fields):
+            return cls(**obj)
+        raise ValueError("Malformed configuration change serialization: %r" % obj)
+
+
+class ConfigChangeSet(BaseChange):
```
barthisrael, line 1763:
```suggestion
class ConfigChangeSet(BaseChange):
    """Represents a set of :class:`ConfigChange` for a given configuration section.
    
    :ivar section str: name of the configuration section related with the changes.
    :ivar changes_set List[:class:`ConfigChange`]: list of configuration changes to be applied to the section.
    """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+    _fields = ["section", "changes_set"]
+
+    def __init__(self, section, changes_set=[]):
```
barthisrael, line 1766:
```suggestion
    def __init__(self, section, changes_set=[]):
        """Initialize a new :class:`ConfigChangeSet` object.
        
        :param section str: name of the configuration section related with the changes.
        :param changes_set List[ConfigChange]: list of configuration changes to be applied to the *section*.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+
+        Generates configuration change objects starting from a dictionary with
+        the same fields.
+
+        :param obj: Dictionary representing the configuration changes set.
+        :type obj: dict
+        :return: Configuration set of changes.
+        :rtype: ConfigChangeSet
+        :raises ValueError: If the dictionary is malformed.
+        """
```
barthisrael, line 1783:
```suggestion
        """
        Factory for configuration change objects.

        Generates configuration change objects starting from a dictionary with
        the same fields.
        
        .. note::
            Handles both :class:`ConfigChange` and :class:`ConfigChangeSet` mapping.

        :param obj: Dictionary representing the configuration changes set.
        :type obj: :class:`dict`
        :return: Configuration set of changes.
        :rtype: :class:`ConfigChangeSet`
        :raises:
            :exc:`ValueError`: If the dictionary is malformed.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+
+class ConfigChangesQueue:
+    """
+    Wraps the management of the config changes queue.
+
+    The `ConfigChangesQueue` class provides methods to read, write, and manipulate
+    a queue of configuration changes. It is designed to be used as a context manager
+    to ensure proper opening and closing of the queue file.
+
+    Once instantiated the queue can be accessed using the `queue` property.
```
barthisrael, line 1805:
```suggestion
    The :class:`ConfigChangesQueue` class provides methods to read, write, and manipulate
    a queue of configuration changes. It is designed to be used as a context manager
    to ensure proper opening and closing of the queue file.

    Once instantiated the queue can be accessed using the :attr:`queue` property.
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+    """
+
+    def __init__(self, queue_file):
+        """
+        Initialize the Config object.
```
barthisrael, line 1810:
```suggestion
        Initialize the :class:`ConfigChangesQueue` object.
        
        :param queue_file str: file where to persist the queue of changes to be processed.
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        """
+        self.queue_file = queue_file
+        self._queue = None
+        self.open()
+
+    @staticmethod
+    def read_file(path) -> List[ConfigChangeSet]:
+        """
+        Reads a json file containing a list of configuration changes.
+        """
```
barthisrael, line 1820:
```suggestion
        """
        Reads a json file containing a list of configuration changes.
        
        :return: the list of :class:`ConfigChangeSet` to be applied to Barman configuration sections.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+
+    @property
+    def queue(self):
+        """
+        Returns the queue object.
+
+        If the queue object is not yet initialized, it will be opened before returning.
+
+        Returns:
+            The queue object.
```
barthisrael, line 1848:
```suggestion
        :return: the queue object.
```

Use reST everywhere instead of mixing reST and Google style.

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        """
+        if self._queue is None:
+            self.open()
+
+        return self._queue
+
+    def open(self):
```
barthisrael, line 1855:
```suggestion
    def open(self):
        """Open and parse the :attr:`queue_file` into :attr:`_queue`."""
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        self._queue = self.read_file(self.queue_file)
+
+    def close(self):
```
barthisrael, line 1858:
```suggestion
    def close(self):
        """Write the new content and close the :attr:`queue_file`."""
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        self._queue = None
+
+
+class ConfigChangesProcessor:
+    """
+    The class is responsible for processing the config changes to
+    apply to the barman config
+    """
+
+    def __init__(self, config):
```
barthisrael, line 1882:
```suggestion
    def __init__(self, config):
        """Initialize a new :class:`ConfigChangesProcessor` object,
        
        :param config Config: the Barman configuration.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+        self.config = config
+        self.applied_changes = []
+
+    def receive_config_changes(self, changes):
+        """
+        Process all the configuration changes
```
barthisrael, line 1888:
```suggestion
        Process all the configuration *changes*.
        
        :param changes Dict[str, str]: each key is the name of a section to be updated, and the value is a dictionary of configuration options along with their values that should be updated in such section.
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+
+    def process_conf_changes_queue(self):
+        """
+        Process the configuration changes in the queue.
+
+        This method iterates over the configuration changes in the queue and applies them one by one.
+        If an error occurs while applying a change, it logs the error and raises an exception.
+
+        Raises:
+            Exception: If an error occurs while applying a change.
```
barthisrael, line 1938:
```suggestion
        :raises:
            :exc:`Exception`: If an error occurs while applying a change.
```

Same comment about reST vs Google.

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+                    changes_queue.queue.pop(0)
+                    self.applied_changes.append(chgs_set)
+
+        except Exception as err:
+            _logger.error("Cannot execute %s: %s", chgs_set, err)
+
+    def apply_change(self, changes):
+        """
+        Apply the given changes to the configuration files.
+        """
```
barthisrael, line 1974:
```suggestion
        """
        Apply the given changes to the configuration files.
        
        :param changes List[ConfigChangeSet]: list of sections and their configuration options to be updated.
        """
```

barman/config.py
```diff
@@ -1676,6 +1687,317 @@ def _validate_with_keys(config_items, allowed_keys, section):
...
+                changed_files.get(chg.config_file),
+            )
+        for file, lines in changed_files.items():
+            with open(file, "w") as cfg_file:
+                cfg_file.writelines(lines)
+
+
+class ConfigChangeSetEncoder(json.JSONEncoder):
+    """
+    JSON encoder for ConfigChangeSet objects.
```
barthisrael, line 1991:
```suggestion
    JSON encoder for :class:`ConfigChange` and :class:`ConfigChangeSet` objects.
```

barthisrael (changes requested):
Some docstring related suggestions.

barthisrael (approved):
I guess this one is good to merge once we apply the docstring changes.
The feature seems to be working well and not failing anymore with any kind of exception.
We need to document the feature later through a separate PR.

PR #882, opened 2024-01-08
Change the minor version on the docs for rsync optionally needed

martinmarques:
There's a requirement in the docs which is optional if you don't use rsync mode. This stated that you needed version 3.0.4 or better, yet this was not true as we need at least version 3.1.0 to use `--ignore-missing-args`.

This commit sets the documentation right so there's no confusion regarding the minimum version of rsync that is required to run barman with rsync.

### DIFF
--- a/doc/manual/15-system_requirements.en.md
+++ b/doc/manual/15-system_requirements.en.md
@@ -10,7 +10,7 @@
     - python-dateutil
     - setuptools
 - PostgreSQL >= 10 (next version will require PostgreSQL >= 11)
-- rsync >= 3.0.4 (optional)
+- rsync >= 3.1.0 (optional)
 
 > **IMPORTANT:**
 > Users of RedHat Enterprise Linux, CentOS and Scientific Linux are


doc/manual/15-system_requirements.en.md
```diff
@@ -10,7 +10,7 @@
     - python-dateutil
     - setuptools
 - PostgreSQL >= 10 (next version will require PostgreSQL >= 11)
-- rsync >= 3.0.4 (optional)
+- rsync >= 3.1.2 (optional)
```
barthisrael, line 13: The PR looks good.

From the [changelog](https://download.samba.org/pub/rsync/NEWS#3.1.0) and PR description/commit message I can see we actually require `3.1.0` for `--ignore-missing-args` to be available.

I can see an internal comment about `3.1.2` being the oldest available in supported operating systems, so I guess that's why we are setting `3.1.2` here. But that seems more of a packaging characteristic than a requirement of Barman.

With that in mind I wonder if we should take some of the following actions:

* Set `3.1.0` as a requirement, and have `>= 3.1.0` in the package requirements; or
* Adjust the commit message/PR description to clarify why we are setting `3.1.2` in the markdown docs file instead of `3.1.0`, just so the history will make sense if one ever attempts to inspect the commit log

_I know this comment seems an overkill, but I felt I should raise it just in case_

Thoughts?

martinmarques: think this makes sense. I'll push another commit to over-write the version dropping to 3.1.0

barthisrael (approved):

PR #886, opened 2024-01-17
Consume queue during cron taking care of config-update interruptions

gcalacoci:


barthisrael (approved):
Looks good to me. Should we add unit tests to cover this?

barthisrael:
I was wondering about something else. The operation should be quite quick. However, it will grab the lock each time `barman cron` runs, even if the file is empty. What if someone or something calls `barman config-update` at the same tiem `barman cron` is already checking the empty  queue?

gcalacoci:
> Looks good to me. Should we add unit tests to cover this?

UT for `ConfigChangesProcessor` are part of the bar-128 pr we merged. in general cli.py has no UT

gcalacoci:
> I was wondering about something else. The operation should be quite quick. However, it will grab the lock each time `barman cron` runs, even if the file is empty. What if someone or something calls `barman config-update` at the same tiem `barman cron` is already checking the empty queue?

well. the config update will fail and tbh I'm fine with that. we don't want to have multiple processes messing with config files at the same time. as the corn will potentially modify them IMHO is worth the risk of a conflict.

barthisrael:
Ok with both points, makes sense. Let's go for it and merge this PR.

PR #888, opened 2024-01-17
Extend `barman config-update` to handle models

barthisrael:
So far `barman config-update` was only able to handle server configuration.

This commit extends the command so it is now also able to handle model configuration.

Besides that we enhance the command in a couple points:

* It now logs the changes applied by `barman config-update` to the Barman configuration sections
* Some basic checks and validations are made when executing the command, e.g.:
  * Basic integrity of the JSON input regarding required keys when creating or update server or model configuration
  * Check if the section is what the user expects it to be, i.e., do not update a server section if the user asked to update that as a model -- and vice-versa

Unit tests updated accordingly.

References: BAR-123.

### DIFF
--- a/barman/config.py
+++ b/barman/config.py
@@ -1930,12 +1930,77 @@ class ConfigChangesProcessor:
         # Get all the available configuration change files in order
         changes_list = []
         for section in changes:
-            section_name = section["server_name"]
+            original_section = deepcopy(section)
+            section_name = None
+            scope = section.pop("scope")
+
+            if scope not in ["server", "model"]:
+                output.warning(
+                    "%r has been ignored because 'scope' is "
+                    "invalid: '%s'. It should be either 'server' "
+                    "or 'model'.",
+                    original_section,
+                    scope,
+                )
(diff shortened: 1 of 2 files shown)


barman/config.py
```diff
@@ -1930,12 +1930,77 @@ def receive_config_changes(self, changes):
         # Get all the available configuration change files in order
         changes_list = []
         for section in changes:
-            section_name = section["server_name"]
+            original_section = deepcopy(section)
+            section_name = None
+            scope = section.pop("scope")
+
+            if scope not in ["server", "model"]:
```
gcalacoci, line 1937: I think we want to move this validation section in a separate method? not now but in the future... for cleaner code. maybe is just me overthinking btw.

gcalacoci (approved):

barman/config.py
barthisrael: I think that makes sense, yes.
I guess we could do that altogether later, when implementing the more advanced validations to the command?

PR #903, opened 2024-02-23
Update instructions for barman user needed privileges

martinmarques:
There was a block of SQL code added some time ago which helps users that do not want to have a superuser running the backups.

These instructions provided SQL commands to GRANT those specific privileges to the barman user.

Since version 15, two of the functions that needed to be GRANTed have not only changed their names, but also their signatures. There is a note regarding such change with appropriate instructions on the new lines to use.

Given there are already two newer versions out, and Postgres 17 coming out in a some months, it makes more sense to have instructions for the latest versions, and a note for anyone running on older Postgres versions.

This patch amends this.

### DIFF
--- a/doc/manual/21-preliminary_steps.en.md
+++ b/doc/manual/21-preliminary_steps.en.md
@@ -29,9 +29,8 @@ postgres@pg$ createuser -P barman
 ```
 
 ``` sql
-GRANT EXECUTE ON FUNCTION pg_start_backup(text, boolean, boolean) to barman;
-GRANT EXECUTE ON FUNCTION pg_stop_backup() to barman;
-GRANT EXECUTE ON FUNCTION pg_stop_backup(boolean, boolean) to barman;
+GRANT EXECUTE ON FUNCTION pg_backup_start(text, boolean) to barman;
+GRANT EXECUTE ON FUNCTION pg_backup_stop(boolean) to barman;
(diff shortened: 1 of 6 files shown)


doc/manual/21-preliminary_steps.en.md
```diff
@@ -29,24 +29,24 @@ postgres@pg$ createuser -P barman
...
 GRANT pg_read_all_settings TO barman;
 GRANT pg_read_all_stats TO barman;
 ```
 
-In the PostgreSQL 15 beta and any subsequent PostgreSQL versions the functions
-`pg_start_backup` and `pg_stop_backup` have been renamed and have different
-signatures. You will therefore need to replace the first three lines in the
+In the case of using PostgreSQL version 14 or a prior version, the functions
+`pg_backup_start` and `pg_backup_stop` had different names and different
+signatures. You will therefore need to replace the first two lines in the
```
gcalacoci, line 43: putting myself in the user shoes, i think I would love to see here a link to the old documentation functions...
I don't know in general I think we should add more links to pg doc when we put code blocks like this one. maybe is not a suggestion for now...

gcalacoci (approved):
LGTM

doc/manual/21-preliminary_steps.en.md
martinmarques: I agree, but I didn't want to make this PR about refactoring the docs. Maybe we should track that separately

PR #909, opened 2024-03-12
Make argcomplete optional

gcalacoci:
Argcomplete is a nice library for bash completion but is not necessary for barman to work properly.
Installation of the lib should be optional and not enforced, and the code should handle this.


amenonsen (approved):

barthisrael (approved):
Looks good.

PR #912, opened 2024-03-14
Load the autoconf file only if it exists

gcalacoci:
Added a guarding statement that loads the barman.auto.conf file only if the file exists.
Improved warning messages in the load_config_file method, adding a specific one in case the file is not found.


barthisrael (approved):
Looks good.

PR #913, opened 2024-03-21
Handle malformed or empty config_changes.queue file

gcalacoci:
If for whatever reason (ie: invalid JSON content or empty queue file)
The JSON parser throws an exception, emits a warning and continues with an
empty list

References: BAR-151

### DIFF
--- a/barman/config.py
+++ b/barman/config.py
@@ -1860,6 +1860,11 @@ class ConfigChangesQueue:
                 return json.load(queue_file, object_hook=ConfigChangeSet.from_dict)
         except FileNotFoundError:
             return []
+        except json.JSONDecodeError:
+            output.warning(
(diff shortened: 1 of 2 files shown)


barthisrael (approved):
Changes look good. They even bring a self-healing mechanism which throws a WARNING just once, and fixes the issue:

```bash
$ barman cron
Starting WAL archiving for server pg

$ echo "" > barman_home/cfg_changes.queue

$ barman cron
WARNING: Malformed or empty configuration change queue: /home/vagrant/barman_home/cfg_changes.queue
Starting WAL archiving for server pg

$ barman cron
Starting WAL archiving for server pg

$ cat barman_home/cfg_changes.queue
[]
```

martinmarques:
I know this is closed, but shouldn't we emit an output message as a warning when the queue file is not found?

PR #931, opened 2024-06-07
add summarize_wal field to backup.info (backup metadata)

andremagui:
With postgres major version >= 17, we have the possibility to do incremental backups with pg_basebackup. 

A crucial element for enabling this functionality is the WAL summarizer process. This is a recent addition, designed to assist incremental backups by identifying the differences in the data directory since the previous backup. 

A new field was added in configuration file ` postgresql.conf` called `summarize_wal`. This field handles the state of the process (on or off).

This commit adds a new field called summarize_wal to the backup.info metadata.

For postgres versions prior to 17, this field will remain as `None`.

Unit tests updated accordingly.

References: BAR-172.

Signed-off-by: Andre Marchesini de Aguiar <andre.marchesini@enterprisedb.com>

### DIFF
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -24,11 +24,12 @@ import os
 import mock
 import pytest
 from dateutil import tz
-from mock import Mock, PropertyMock, patch
+from mock import Mock, PropertyMock, patch, call
 
 from barman.backup_executor import (
     ExclusiveBackupStrategy,
     PostgresBackupExecutor,
+    PostgresBackupStrategy,
     RsyncBackupExecutor,
     SnapshotBackupExecutor,
 )
@@ -589,7 +590,8 @@ class TestStrategy(object):
         """
         # Build a backup_manager using a mocked server
         server = build_mocked_server(
-            main_conf={"backup_options": BackupOptions.EXCLUSIVE_BACKUP}
+            main_conf={"backup_options": BackupOptions.EXCLUSIVE_BACKUP},
+            pg_version=170000,
         )
         backup_manager = build_backup_manager(server=server)
 
@@ -643,7 +645,8 @@ class TestStrategy(object):
         # Test: start concurrent backup
         # Build a backup_manager using a mocked server
         server = build_mocked_server(
-            main_conf={"backup_options": BackupOptions.CONCURRENT_BACKUP}
+            main_conf={"backup_options": BackupOptions.CONCURRENT_BACKUP},
+            pg_version=170000,
         )
         backup_manager = build_backup_manager(server=server)
         # Simulate old Postgres version
@@ -826,6 +829,45 @@ class TestStrategy(object):
         # check has status True
         else:
             assert check_result.status is True
+
+    @pytest.mark.parametrize(
+        ("server_version", "expected_value"),
+        [(160000, None), (170000, "on")],
+    )
+    def test__pg_get_metadata(self, server_version, expected_value):
+        # Given a PostgreSQL connection of the specified version
+        mock_postgres = mock.Mock()
+        mock_postgres.server_version = server_version
+        # Mock postgres server.get_setting() call
+        mock_postgres.get_setting.side_effect = [
+            "data_directory",
+            expected_value,
+        ]
+
+        # Mock postgres server.get_configuration_files() call
+        mock_postgres.get_configuration_files.return_value = dict(
+            config_file="/etc/postgresql.conf",
+            hba_file="/pg/pg_hba.conf",
+            ident_file="/pg/pg_ident.conf",
+        )
+        # Mock postgres server.get_tablespaces() call
+        tablespaces = [Tablespace._make(("test_tbs", 1234, "/tbs/test"))]
+        mock_postgres.get_tablespaces.return_value = tablespaces
+
+        strategy = PostgresBackupStrategy(mock_postgres, "test server")
+        backup_info = build_test_backup_info()
+        strategy._pg_get_metadata(backup_info)
+
+        mock_postgres.get_tablespaces.assert_called_once()
+        mock_postgres.get_configuration_files.assert_called_once()
+        if mock_postgres.server_version < 170000:
+            mock_postgres.get_setting.assert_called_once_with("data_directory")
+            assert backup_info.summarize_wal is None
+        else:
+            calls = [call("data_directory"), call("summarize_wal")]
+            assert mock_postgres.get_setting.call_count == 2
+            mock_postgres.get_setting.assert_has_calls(calls, any_order=False)
+            assert backup_info.summarize_wal == "on"
 
 
 class TestPostgresBackupExecutor(object):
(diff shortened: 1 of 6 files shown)


barthisrael:
I run a test with the code and it seems to do what is required. I can see the `summarize_wal` with the correct value in both of my Postgres 17 servers (one with rsync, another with postgres method), and I can see `None` for the Postgres 16 server:

```console
$ find ~/ -type f -name "backup.info" -print -exec bash -c "grep summarize_wal {}; echo" \;
/home/vagrant/barman_home/pg17-postgres-default/base/20240610T182147/backup.info
summarize_wal=on

/home/vagrant/barman_home/pg17-rsync/base/20240610T182149/backup.info
summarize_wal=on

/home/vagrant/barman_home/pg16-postgres-default/base/20240610T182145/backup.info
summarize_wal=None
```

Let me check the code and commits as well.

tests/testing_helpers.py
```diff
@@ -405,7 +407,9 @@ def build_real_server(global_conf=None, main_conf=None):
     )
 
 
-def build_mocked_server(name=None, config=None, global_conf=None, main_conf=None):
+def build_mocked_server(
+    name=None, config=None, global_conf=None, main_conf=None, pg_version=170000
```
barthisrael, line 411:
```suggestion
    name=None, config=None, global_conf=None, main_conf=None, pg_version=None
```

I'd say to make it default to `None`, then adjust the unit tests so they set the version they expect for a given unit test.

This way we have a more explicit setting in the unit test which requires this information, if you get what I mean?

tests/testing_helpers.py
```diff
@@ -433,6 +438,8 @@ def build_mocked_server(name=None, config=None, global_conf=None, main_conf=None
     server.postgres.xlog_segment_size = DEFAULT_XLOG_SEG_SIZE
     server.path = "/test/bin"
     server.systemid = "6721602258895701769"
+    # Add server_version to comply with major PG server version contraint for incremental backups
```
barthisrael, line 441:
```suggestion
```

I'd leave this comment out, as that's something too specific for a test "fixture".

barthisrael (changes requested):
The changes look good.

I posted a couple of comments regarding the testing helpers.

Besides that, we need to work on cleaning up the commit history here. We have 8 commits -- let's squash them and reduce the amount of commits as we will use the "rebase and merge" approach later. I'd say to do something like this:

* Code changes in one commit;
* Refactoring of the testing helpers, which is kind of unrelated (it improves things like the pg 9.6 things), in another commit;
* Unit tests for the code changes in a last commit.

Note: you may also choose splitting into 2 commits only, e.g. helpers refactoring in one commit, code + unit tests in another commit.

A couple of notes in regards to commits:

* Try to be more descriptive in the commit messages, giving some context regarding why/how the changes were performed. You need to judge how much information you will add to the commit, but we avoid pushing empty commit messages (AKA `git commit -m 'title')
* We add a `References: BAR-XYZ` in the last line of the commit, just so it's easier to track the Jira related with the ticket -- if we ever need to do that when inspecting the commit log in the future.

On the same line, we do something similar for the PR description.

If you want, you can check https://github.com/EnterpriseDB/barman/pull/888 for an example.

tests/testing_helpers.py
gcalacoci: @barthisrael I had a quick call with @andremagui yesterday, and I suggested setting the value to 17000 as seemed to me at a glance, the less impactful way to implement this change.

Looking deeper into the code today, the above suggestion seems sensible and more correct. I recommend it too.

gcalacoci (changes requested):

tests/testing_helpers.py
barthisrael: Thanks for the heads up, got it!
I mentioned about using `None` because it would remain backward compatible with the existing tests, and would "force" us to choose it appropriately in the unit tests where that thing is a constraint.

barthisrael:
@gcalacoci for some reason the SonarQube scan is failing 🤔 (since a few runs already: https://github.com/EnterpriseDB/barman/actions/workflows/sonarqube-scan.yml).
I wonder if we need to report this to Foundation team? I guess so given the nature of the issue:

```console
java.lang.UnsupportedClassVersionError: org/sonar/batch/bootstrapper/EnvironmentInformation has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

tests/testing_helpers.py
```diff
@@ -415,6 +419,7 @@ def build_mocked_server(name=None, config=None, global_conf=None, main_conf=None
         it is possible to override or add new values to the [barman] section
     :param dict[str,str|None]|None main_conf: using this dictionary
         it is possible to override/add new values to the [main] section
+    :param int pg_version: postgres server version, default to 17
```
barthisrael, line 422:
```suggestion
    :param int pg_version: postgres server version, default to ``None``
```

gcalacoci (changes requested):
@andremagui while discussing this with @barthisrael we noticed that we are missing specific tests for the `backup_info` new field `summarize_wal`.

What I mean is that we should add a new test, something like `test_summarize_wal` or something similar in the `test_executor.py` file that tests that if version is 17, the value is correctly set, if is lower than 17, is the value is `none`

tests/test_executor.py
```diff
@@ -831,6 +830,41 @@ def test_exclusive_check(self, server_version, expected_message, capsys):
...
+    @pytest.mark.parametrize(
+        ("server_version", "expected_value"),
+        [(160000, None), (170000, "on")],
+    )
+    def test__pg_get_metadata(self, server_version, expected_value):
+        # Given a PostgreSQL connection of the specified version
+        mock_postgres = mock.Mock()
+        mock_postgres.server_version = server_version
+        # Mock postgres server.get_setting() call
+        mock_postgres.get_setting.return_value = expected_value
```
barthisrael, line 842:
```suggestion
        mock_postgres.get_setting.side_effect = [
            "some_data_directory",
            "on",
        ]
```

tests/test_executor.py
```diff
@@ -831,6 +830,41 @@ def test_exclusive_check(self, server_version, expected_message, capsys):
...
+        strategy._pg_get_metadata(backup_info)
+
+        mock_postgres.get_tablespaces.assert_called_once()
+        mock_postgres.get_configuration_files.assert_called_once()
+        if mock_postgres.server_version < 170000:
+            mock_postgres.get_setting.assert_called_once_with("data_directory")
+            assert backup_info.summarize_wal is None
+        else:
+            calls = [call("data_directory"), call("summarize_wal")]
+            mock_postgres.get_setting.assert_has_calls(calls, any_order=False)
```
barthisrael, line 865:
```suggestion
            assert mock_postgres.get_setting.call_count == 2
            mock_postgres.get_setting.assert_has_calls(calls, any_order=False)
```

barthisrael (changes requested):
Final round of nitpicking 😆

barthisrael (approved):
Looks good!

gcalacoci (approved):
LGTM

PR #932, opened 2024-06-07
Change postgres executor for incremental backups

gustabowill:
Changes the postgres executor and related classes to handle the new `--incremental ` CLI argument. This first part is mostly only concerned about passing the argument correctly to the `pg_basebackup` command. I also added a method to get the manifest of the parent backup, although I'm not so sure if it was the best way to do it.


barman/backup_executor.py
```diff
@@ -733,6 +749,14 @@ def _prepare_backup_destination(self, dest_dirs):
             # chmod 0700 octal
             os.chmod(dest_dir, 448)
 
+    def _get_backup_manifest_from_backup(self, backup_info):
```
barthisrael, line 752: Maybe we could use `BackupManifest._get_manifest_file_path` to handle this instead of defining a new method?

barman/server.py
```diff
@@ -1589,7 +1589,7 @@ def delete_backup(self, backup):
             output.error("Permission denied, unable to access '%s'" % e)
             return
 
-    def backup(self, wait=False, wait_timeout=None, backup_name=None):
+    def backup(self, wait=False, wait_timeout=None, backup_name=None, **kwargs):
```
barthisrael, line 1592: Add the expected `kwparam` entry to the docstring.

tests/test_server.py
```diff
@@ -2210,7 +2210,7 @@ def test_put_wal_fsync(self, fd_mock, ff_mock, tmpdir, capsys, caplog):
         assert (
             "Received file '00000001000000EF000000AB' "
             "with checksum '34743e1e454e967eb76a16c66372b0ef' "
-            "by put-wal for server 'main'\n" in caplog.text
+            "by put-wal for server 'main'" in caplog.text
```
barthisrael, line 2213: This seems an unrelated change.

barthisrael (changes requested):
Put an initial quick review.

barman/backup_executor.py
```diff
@@ -358,7 +358,7 @@ def _validate_compression(self, remote_status):
                 exc,
             )
 
-    def backup(self, backup_info):
+    def backup(self, backup_info, parent_backup_info=None):
```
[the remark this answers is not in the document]
gcalacoci: I do agree with @barthisrael assessment, `**kwargs` in the base class is the cleanest/safest way to handle this

barman/backup_executor.py
```diff
@@ -562,7 +565,7 @@ def fetch_remote_status(self):
 
         return remote_status
 
-    def backup_copy(self, backup_info):
+    def backup_copy(self, backup_info, parent_backup_info=None):
```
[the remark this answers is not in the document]
gcalacoci: All the signature changes have to happen in the base abstract class...  otherwise, we are not correctly implementing the abstract methods (and just overloading on the methods names)

barman/backup_executor.py
gcalacoci: chiming in on this:
we don't need anything else except the path to the parent's manifest, using `BackupManifest._get_manifest_file_path` requires to instantiate a `BackupManifest` obj and the result is not really different from what @gustabowill has done in his implementation. 

STILL I'm not sure about using `_get_backup_manifest_from_backup` as in the end is a shortcut to an oneliner to build a path, that realistically is going to be used once or maybe twice. 

Makes sense to me to add stuff to the backup info possibly. @barthisrael let's discuss this further...

barman/backup_executor.py
```diff
@@ -608,6 +615,14 @@ def backup_copy(self, backup_info):
         # for the whole duration of the copy
         self.server.close()
 
+        # Find the backup_manifest file of the previous backup in case it is
+        # an incremental backup
+        parent_backup_manifest = None
```
gcalacoci, line 620: A little suggestion here, if is just a path and not a full manifest, the name of the var should reflect it.
simply put: reading `parent_backup_manifest` I was expecting to get the full contents of the manifest in there, not just the path to the file. so should be `parent_backup_manifest_path`.
Same for the name of the method `_get_backup_manifest_from_backup`, I would expect this to return the full manifest file contents, not just the path.

barman/backup_executor.py
barthisrael: Yeah, I didn't notice the other 2 parameters in the class.

Although we could hack that as the objects are not used in `__init__` 😆 

```python
>>> from barman.backup_manifest import BackupManifest
>>> BackupManifest("/home/vagrant/barman_home/pg17-postgres-default/base/20240610T115127/data", None, None)._get_manifest_file_path()
'/home/vagrant/barman_home/pg17-postgres-default/base/20240610T115127/data/backup_manifest'
```

But this doesn't seem elegant either.

This is one of those situations where we would like to avoid logic duplication, but doing so seems to be an overkill...

At a first glance I don't think we would like to add that information to the backup info. All the information in the backup info seems to not depend on paths in the Barman server:

```ini
backup_label='START WAL LOCATION: 0/17000028 (file 000000010000000000000017)\nCHECKPOINT LOCATION: 0/17000080\nBACKUP METHOD: streamed\nBACKUP FROM: primary\nSTART TIME: 2024-06-10 11:51:27 UTC\nLABEL: pg_basebackup base backup\nSTART TIMELINE: 1\n'
begin_offset=40
begin_time=2024-06-10 11:51:27.384936+00:00
begin_wal=000000010000000000000017
begin_xlog=0/17000028
compression=None
config_file=/var/lib/pgsql/17/data/postgresql.conf
copy_stats={'copy_time': 1.653982, 'total_time': 1.653982}
deduplicated_size=23403347
end_offset=0
end_time=2024-06-10 11:51:29.057709+00:00
end_wal=000000010000000000000018
end_xlog=0/18000000
error=None
hba_file=/var/lib/pgsql/17/data/pg_hba.conf
ident_file=/var/lib/pgsql/17/data/pg_ident.conf
included_files=['/var/lib/pgsql/17/data/postgresql.auto.conf']
mode=postgres
pgdata=/var/lib/pgsql/17/data
server_name=pg17-postgres-default
size=23403347
status=WAITING_FOR_WALS
summarize_wal=on
systemid=7377136404149220174
tablespaces=None
timeline=1
version=170000
xlog_segment_size=16777216
```

Doing something like that (adding the manifest path to the backup info) could break a setup if someone changes the `base_directory` or `barman_home` or server name or ... in the Barman configuration.

Maybe we should just get rid of the method and do a oneliner with duplicated code after all.

Thoughts?

barman/backup_executor.py
barthisrael: Although the name looks longer, which we usually tend to avoid, it provides a more clear understanding of its purpose.

barman/backup_executor.py
gcalacoci: > Although we could hack that as the objects are not used in __init__ 😆

Yeah, I thought about that too, but discarded the idea as is terrible and, as you noted, not elegant 😛 

> Doing something like that (adding the manifest path to the backup info) could break a setup if someone changes the base_directory or barman_home or server name or ... in the Barman configuration.

Paths can be preserved if relative to barman directories instead of being absolute...
What I mean is: no matter what you change in the configuration, the path to the basebackup directory is always available inside the config object as `config.basebackups_directory`

what I mean is that we should save values not like this:
```
parent_manifest_path=/absolute/path/to/backups/dir/<backupid>/<manifestname>
```
But something like this:
```
parent_manifest_path=<backupid>/<manifestname>
```
whatever is previous to backupid is ALWAYS inside the `basebackups_directory` value in the config, at this point building the path is trivial.

My point here is that keeping track of parent backup stuff inside the backup info makes it easier to access parent data in the code, AND is relevant info that possibly simplifies things even when we need to rebuild the full chain of backups for the recovery. 

Btw let's cycle around this a bit more @barthisrael @gustabowill

gustabowill: I agree with the point that @barthisrael made about the information in the `backup_info` file, most info in there seems mostly related to the postgres server configuration and not a lot related to barman. 

I think that if we can not have the absolute path to the manifest file then the efforts of building it from a relative path would be pretty much the same as what I am doing now, which is building it manually using the data directory + the file name.

I agree with @gcalacoci that keeping information about the parent is important for rebuilding the chain of backups afterwards and think that there is already a separate ticket for that, but i guess the id of the parent is most important thing, we could get the rest of the information once we have that, right?

We could maybe build a method in the `LocalBackupInfo` to return the manifest path? Or, just like @barthisrael  suggested, remove the method and use a one line code in the place where it's needed.

barthisrael: @gustabowill Ends up I never submitted my reply here, sorry.

After discussing with @gcalacoci earlier today, we thought of moving your method to the `LocalBackupInfo` class, as you just suggested.

* Move `PostgresBackupExecutor._get_backup_manifest_from_backup` to `LocalBackupInfo.get_manifest_path`
* Slightly change the logic (because manifest is created by `pg_basebackup`, so it only makes sense for `postgres` mode`:
   * if mode is not postgres -> return None;
   * if mode is postgres, do something similar to what you did here (data dir + "backup_manifest").

gustabowill: @gcalacoci @barthisrael Ok, I just pushed some minor changes related to variables names and docstrings as well as put everything in a single commit.

barthisrael:
The PR looks good.

I'm only missing a unit test for the `PgBaseBackup` class -- check if `incremental` is passed down when there is a parent manifest.

For the executor itself I think we can leave it as it is right now as we might refactor this a bit through BAR-165.

barthisrael (approved):
Looks good to me.

gcalacoci (approved):
LGTM

PR #935, opened 2024-06-12
Check that summarize_wal is enabled on incremental backups

gustabowill:
When performing an incremental backup using postgres mode with the new `--incremental` flag it is required that the user has `summarize_wal` enabled on the postgres server. This PR adds a new validation method which is responsible for checking if the user has the option enabled and issue an error otherwise.

This validation could not be added to the already existing `check` method since that is also run when `barman check <server>` is issued and `--incremental` is a one-time config. I had to add a new method `validate_backup_args` to the backup manager and executors, which is called in the server context before the backup starts.

References: BAR-160

### DIFF
--- a/barman/backup.py
+++ b/barman/backup.py
@@ -44,6 +44,7 @@ from barman.compression import CompressionManager
 from barman.config import BackupOptions
 from barman.exceptions import (
     AbortedRetryHookScript,
+    BackupException,
     CompressionIncompatibility,
     LockFileBusy,
     SshCommandException,
@@ -583,6 +584,32 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
         script.run()
 
         return True
+
+    def validate_backup_args(self, **kwargs):
+        """
+        Validate backup arguments and Postgres configurations. Arguments
+        might be syntactically correct but still be invalid if necessary
+        Postgres configurations are not met.
+
+        :kwparam str parent_backup_id: id of the parent backup when taking a
+            Postgres incremental backup
+        :raises BackupException: if a command argument is considered invalid
+        """
+        if "parent_backup_id" in kwargs:
+            self._validate_incremental_backup_configs()
+
+    def _validate_incremental_backup_configs(self):
+        """
+        Check required configurations for a Postgres incremental backup
+
+        :raises BackupException: if a required configuration is missing
+        """
+        summarize_wal = self.server.postgres.get_setting("summarize_wal")
+        if summarize_wal != "on":
+            raise BackupException(
+                "'summarize_wal' option has to be enabled in the Postgres server "
+                "to perform an incremental backup using the Postgres backup method"
+            )
 
     def backup(self, wait=False, wait_timeout=None, name=None, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


barman/backup_executor.py
```diff
@@ -358,6 +367,26 @@ def _validate_compression(self, remote_status):
...
+        Postgres configurations are not met.
+
+        :kwparam barman.infofile.LocalBackupInfo parent_backup_info:
+            information of the parent backup in case it is an incremental backup
+            using the postgres mode
+        :raises CommandFailedException: if a command argument is considered invalid
+        """
+
+        if "parent_backup_info" in kwargs:
+            summarize_wal = self.server.postgres.get_setting("summarize_wal")
```
gcalacoci, line 383: this information is gathered, and saved in the backup_info, thanks to @andremagui patch.
doesn't make sense to gather it twice.
On the other hand, @andremagui patch gathers that value during the start backup action which is too late in the chain of calls, we should stop the backup earlier.

barman/backup.py
```diff
@@ -584,6 +584,16 @@ def delete_backup(self, backup, skip_wal_cleanup_if_standalone=True):
...
 
+    def validate_backup_args(self, **kwargs):
+        """
+        Validate backup arguments and Postgres configurations. Arguments
+        might be syntactically correct but still be invalid if necessary
+        Postgres configurations are not met.
+
+        :raises CommandFailedException: if a command argument is considered invalid
+        """
+        self.executor.validate_backup_args(**kwargs)
```
gcalacoci, line 595: IMHO the implementation of this method should be part of the `BackupManager` object and not of the executor

barman/backup_executor.py
gcalacoci: OK, so @gustabowill I went back and forth in the code a couple of times and I think we should:
* ignore for now that the same check down in the call chain, during the start backup.
* move the implementation into the `BackupManager` object from the `Executor` object. TBH this can be possibly moved into the `Server` object too.
* change the exception from `CommandFailedException` to `BackupException` that is more appropriate IMHO as `CommandFailedException` is used for sub-commands calls like `rsync` or `pg_basebackup` or barman internal sub-calls, and is usually associated to the `Command` object ( https://github.com/EnterpriseDB/barman/blob/f0911baff5372dae1d28dc3d2d8a68aebb2a3310/barman/command_wrappers.py#L115 )

gcalacoci (changes requested):

barman/backup_executor.py
gustabowill: Thanks @gcalacoci , I decided to put the logic in the `BackupManager`. I will wait for the #936 to be merged to rebase this PR on top of the master and change what is necessary.

barthisrael (commented):
I like the changes proposed here.

They can feel a bit like "spaghetti" as we are defining methods in `BackupManager` and calling them from `Server` to validate things that are related with `postgres` backup mode, but I think that's the best thing we can achieve with the current code.

Maybe in the future we are able to do something better than this, if we study refactoring a bit the code.

I'm only wondering if we should use `BackupException` or create some other class as `BackupException` is used for other things, which usually led to marking the backup as `FAILED` in other places? Maybe I'm overthinking here, but was willing to bring this up to get other people opinion about this IAC.

gcalacoci (approved):
LGTM

barthisrael (approved):
Looks good!

PR #936, opened 2024-06-13
Maintain relationship between parent and children backups in `backup.info`

gcalacoci:
This PR adds a couple of fields to the `backup.info`:

* `parent_backup_id`: stores the ID of the parent backup when taking an incremental backup through `pg_basebackup`;
* `children_backup_ids`: stores a list of IDs of the backups which are incremental of this backup.

That information is updated by the backup manager whenever a backup is executed with `parent_backup_id` parameter set, which indicates this is a `pg_basebackup` incremental backup. When starting the new backup, its parent backup ID is immediately filled. Before finishing the backup, the list of children of its parent backup is also updated. We do that
independently if the current backup succeeded or failed (we need to know who are the children so we can delete them all later if removal of the parent backup is requested, even if the children are failed backups).

Through this PR we also implemented three new methods in the `LocalBackupInfo` class. These are intended to make it easier to navigate through the incremental backup tree, making it easier to take incremental backups, and to later remove the backups in the tree when needed. The methods are:

* `get_backup_manifest_path`: the manifest of the parent backup is required by `pg_basebackup` when taking incremental backups. This method is handy for getting the path to that file;
* `get_parent_backup_info`: gets the backup info object of the parent backup of this backup. Makes it easier to take incremental backups and to update the relationships;
* `get_child_backup_info`: gets the backup info object of a child backup of this backup. Will make it easier to remove backups and apply the retention policies.

With all that in mind, we also applied a small refactoring of the `backup` and `backup_copy` methods. They used to use `kwargs` to avoid breaking the interface of the methods as they support both Postgres and rsync based backups, and in case of Postgres both incremental and full backups. As we now have a way of navigate through backups in the incremental backup tree, we no longer need to keep passing `LocalBackupInfo` objects through all functions and methods -- we can easily retrieve them ay any point by using the helper methods listed above.

References: BAR-165.


barman/backup.py
```diff
@@ -697,6 +701,23 @@ def backup(self, wait=False, wait_timeout=None, name=None, **kwargs):
...
             if backup_info:
+
+                # IF is an incremental backup, we save here child backup info id
+                # inside the parent list of children. no matter if the backup
+                # is successful or not. This is needed to be able to retrieve
+                # also failed incremental backups for removal or other operations
+                # like show-backup.
+                parent_backup_info = backup_info.get_parent_backup_info()
+
+                if parent_backup_info:
```
gcalacoci, line 712: I did some thinking here and dumped most of it in the above comment. i think that doing this save here is the way to go @barthisrael

barthisrael: That makes sense!
Can you believe I had a similar through while I was driving -- but because of that I was not able to tell you 😆 
I'm happy with that!

barman/backup.py
```diff
@@ -697,6 +701,23 @@ def backup(self, wait=False, wait_timeout=None, name=None, **kwargs):
...
+                    if parent_backup_info.children_backup_ids:
+                        parent_backup_info.children_backup_ids.append(  # type: ignore
+                            backup_info.backup_id
+                        )
```
barthisrael, line 716: This is not working because currently the value is stored as text.

barman/infofile.py
```diff
@@ -792,3 +794,25 @@ def save(self, filename=None, file_object=None):
...
+
+    def get_child_backup_info(self, child_backup_id):
+        if self.children_backup_ids:
+            if child_backup_id in self.children_backup_ids:
+                return LocalBackupInfo(
+                    self.server,
+                    backup_id=child_backup_id,
+                )
+
+        return None
```
barthisrael, line 818: I think we need to check the for "empty" backup info.

The implementation seems to be working well, but if the parent or child backup doens't exist anymore, we will get an empty local backup info, e.g.:

```python
LocalBackupInfo(backup_label=None, begin_offset=None, begin_time=None, begin_wal=None, begin_xlog=None, children_backup_ids=None, compression=None, config_file=None, copy_stats=None, deduplicated_size=None, end_offset=None, end_time=None, end_wal=None, end_xlog=None, error=None, hba_file=None, ident_file=None, included_files=None, mode='postgres', parent_backup_id=None, pgdata=None, server_name='pg17-postgres-default', size=None, status='EMPTY', summarize_wal=None, systemid=None, tablespaces=None, timeline=None, version=None, xlog_segment_size=16777216)
```

I'd say to also return `None` here if the `status` is `EMPTY`.

What do you think?

barman/backup.py
barthisrael: I added a couple of functions to handle dump and load of backup ids.

barman/infofile.py
```diff
@@ -792,3 +824,31 @@ def save(self, filename=None, file_object=None):
...
+            if child_backup_id in self.children_backup_ids:
+                backup_info = LocalBackupInfo(
+                    self.server,
+                    backup_id=child_backup_id,
+                )
+
+                if backup_info.status != BackupInfo.EMPTY:
+                    return backup_info
+
+        return None
```
barthisrael, line 854: I added some logic here to treat "empty" `LocalBackupInfo` as `None`.

barthisrael:
I guess this looks to be in a good shape.
I think we can work on adding the missing docstrings and/or comments.
Then we can work on squashing the commits for a cleaner history. We can use the `Co-authored-by: NAME <NAME@EXAMPLE.COM>` syntax to both of us are assigned to the commits.

barthisrael:
(We also need unit tests)

barman/backup.py
gcalacoci: yeah I realized that this was necessary later after I posted this code snippet, something similar to what happens for tablespaces

gcalacoci (commented):
Interestingly enough I cannot approve my pull request...
@martinmarques can you?

Also, I've added some missing docstrings and everything else looks good to me. @barthisrael  we can merge as it is or eventually squash more.

martinmarques (approved):

barthisrael (approved):
Looks good to me.

PR #938, opened 2024-06-17
Add shortcut to get latest full backup

andremagui:
With the imminent introduction of Postgres core incremental backups support in Barman, it will become necessary to retrieve the latest full backup. Currently, Barman can only retrieve the most recent backup regardless of its type.

This adjustment is essential to align with the specifications outlined in the BAR-161 proposal for enhancing the command-line interface that introduces the --incremental option, which facilitates the execution of incremental backups for postgres databases.

Incremental backups establish a sequential series of backups, where the initial backup invariably constitutes a full backup. Subsequently, numerous instances may arise wherein both full and incremental backups are generated.

References: BAR-173

Signed-off-by: Andre <andre.marchesini@enterprisedb.com>

### DIFF
--- a/barman/backup.py
+++ b/barman/backup.py
@@ -361,8 +361,8 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
         Get the id of the latest/last backup in the catalog (if exists)
 
         :param status_filter: The status of the backup to return,
-            default to DEFAULT_STATUS_FILTER.
-        :return string|None: ID of the backup
+            default to :attr:`DEFAULT_STATUS_FILTER`.
+        :return str|None: ID of the backup
         """
         available_backups = self.get_available_backups(status_filter)
         if len(available_backups) == 0:
@@ -370,6 +370,29 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
 
         ids = sorted(available_backups.keys())
         return ids[-1]
+
+    def get_last_full_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
+        """
+        Get the id of the latest/last FULL backup in the catalog (if exists)
+
+        :param status_filter: The status of the backup to return,
+            default to :attr:`DEFAULT_STATUS_FILTER`.
+        :return str|None: ID of the backup
+        """
+        available_full_backups = list(
+            filter(
+                lambda backup: backup.is_full_and_eligible_for_incremental(),
+                self.get_available_backups(status_filter).values(),
+            )
+        )
+
+        if len(available_full_backups) == 0:
+            return None
+
+        backup_infos = sorted(
+            available_full_backups, key=lambda backup_info: backup_info.backup_id
+        )
+        return backup_infos[-1].backup_id
 
     def get_first_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
         """
--- a/barman/infofile.py
+++ b/barman/infofile.py
@@ -877,3 +877,26 @@ class LocalBackupInfo(BackupInfo):
                     return backup_info
 
         return None
+
+    def is_full_and_eligible_for_incremental(self):
+        """
+        Used to filter out backups that have a parent backup id and are not
+        considered as FULL backups.
+
+        .. note::
+            Only consider backups which are eligible for Postgres core
+            incremental backups:
+
+            * backup_method = ``postgres``
+            * summarize_wal = ``on``
+            * parent_backup_id = ``None``
+
+        :return bool: True if it's a full backup or False if not.
+        """
+        if (
+            self.backup_manager.config.backup_method == "postgres"
+            and self.summarize_wal == "on"
+            and not self.parent_backup_id
+        ):
+            return True
+        return False
--- a/barman/server.py
+++ b/barman/server.py
@@ -1691,10 +1691,22 @@ class Server(RemoteStatusMixin):
         Get the id of the latest/last backup in the catalog (if exists)
 
         :param status_filter: The status of the backup to return,
+            default to :attr:`BackupManager.DEFAULT_STATUS_FILTER`.
+        :return str|None: ID of the backup
+        """
+        return self.backup_manager.get_last_backup_id(status_filter)
+
+    def get_last_full_backup_id(
+        self, status_filter=BackupManager.DEFAULT_STATUS_FILTER
+    ):
+        """
+        Get the id of the latest/last FULL backup in the catalog (if exists)
+
+        :param status_filter: The status of the backup to return,
             default to DEFAULT_STATUS_FILTER.
         :return string|None: ID of the backup
         """
-        return self.backup_manager.get_last_backup_id(status_filter)
+        return self.backup_manager.get_last_full_backup_id(status_filter)
 
     def get_first_backup_id(self, status_filter=BackupManager.DEFAULT_STATUS_FILTER):
         """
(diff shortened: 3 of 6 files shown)


barman/backup.py
```diff
@@ -370,6 +370,24 @@ def get_last_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
...
 
+    def get_last_full_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
+        """
+        Get the id of the latest/last FULL backup in the catalog (if exists)
+
+        :param status_filter: The status of the backup to return,
+            default to DEFAULT_STATUS_FILTER.
+        :return string|None: ID of the backup
+        """
+        available_backups = self.get_available_backups(status_filter)
```
gcalacoci, line 381: I wonder if we should just filter out whatever has not parent = none.
something like:
```python
    available_full_backups = sorted(
        list(filter(lambda x: x.parent_backup_id is None,
                    self.get_available_backups(status_filter))))
    if len(available_full_backups) == 0:
        return None

    ids = sorted(available_full_backups.keys())
    return ids[-1]
```
I thought about this because I expect the number of incremental backups for a server to be higher compared to full backups... and is totally possible that we have to discard a non-small amount of incrementals before finding the first full.
The above implementation is technically quicker as we use a primitive chain of calls that has an underlying implementation in C..
But is possibly an overkill and maybe less readable.
@barthisrael @andremagui @bleidens and @gustabowill  thoughts on this?

barthisrael: We don't know how people will play with this in practice, but I guess your code is clean enough to read.

I'd only say to create a couple of methods in `LocalBackupInfo` maybe, let's say:

* `is_full_backup`: check if mode is postgres, summarize is on, parent is None
* `is_incremental_backup`: check if mode is postgres, summarize is on, parent is not None

Or some other name? Would worth it a note in the docstring of the method, something like:

```python
"""
    (...)
    .. note::
        Only consider backups which are eligible for Postgres core incremental backups
    (...)
"""
```

Then we could have something like this:

```python
list(filter(lambda x: x.is_full_backup(),
```

Any thoughts?

andremagui: LTGM. It looks cleaner having this split in methods and applied like that! I do agree with @gcalacoci implementation and @barthisrael solution so we have something more readable.

gcalacoci: I'm not sure that `is_full_backup` as the name of the method is "correct" as theoretically a pg_basebackup executed backup that has the summarize on, is full... but is not eligible for incremental... 
OTOH `is_full_and_incremental_eligible` is a terrible name 😓

andremagui: has_parent_backup_id?

barthisrael: > I'm not sure that `is_full_backup` as the name of the method is "correct" as theoretically a pg_basebackup executed backup that has the summarize on, is full... but is not eligible for incremental... OTOH `is_full_and_incremental_eligible` is a terrible name 😓 @barthisrael @andremagui

Right, that's why I thought of using the docstring note, which will show on hover in the IDE e.g.

I don't have many ideas of better names 😆

gcalacoci: formally is the most correct, but i cannot stand it

barman/infofile.py
```diff
@@ -877,3 +877,24 @@ def get_child_backup_info(self, child_backup_id):
...
+        as FULL backups.
+
+        :note: Only consider backups which are eligible for Postgres core incremental
+        backups
+        backup_method = 'postgres', summarize_wal = 'on', parent_backup_id = None
+
+        :return bool: True if it's a full backup or False if not.
+        :raise barman.exceptions.BackupException: If the backup_method != 'postgres'
+        """
+        if self.backup_manager.config.backup_method != "postgres":
```
gcalacoci, line 893: not sure we want to raise an exception. If the backup method is not Postgres, the backup is not eligible and we just return false and the backup is filtered out of the list. An exception here seems too disruptive IMHO.

gcalacoci (changes requested):

barman/backup.py
```diff
@@ -371,6 +371,27 @@ def get_last_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
         ids = sorted(available_backups.keys())
         return ids[-1]
 
+    def get_last_full_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
+        """
+        Get the id of the latest/last FULL backup in the catalog (if exists)
+
+        :param status_filter: The status of the backup to return,
+            default to DEFAULT_STATUS_FILTER.
+        :return string|None: ID of the backup
```
barthisrael, line 380:
```suggestion
            default to :attr:`DEFAULT_STATUS_FILTER`.
        :return str|None: ID of the backup
```

Adjusting sphinx syntax.

barman/infofile.py
```diff
@@ -877,3 +877,22 @@ def get_child_backup_info(self, child_backup_id):
...
         return None
+
+    def is_full_and_eligible_for_incremental(self):
+        """
+        Used to filter out backups that have a parent backup id and are not considered
+        as FULL backups.
+
+        :note: Only consider backups which are eligible for Postgres core incremental
+        backups
+        backup_method = 'postgres', summarize_wal = 'on', parent_backup_id = None
```
barthisrael, line 888:
```suggestion
        .. note::
            Only consider backups which are eligible for Postgres core incremental
            backups:

            * backup_method = ``postgres``
            * summarize_wal = ``on``
            * parent_backup_id = ``None``
```

Adjusting to use the sphinx syntax for notes (`.. note::`). Also, using double backticks for literals.

The usage of a list is only a side suggestion.

barman/server.py
```diff
@@ -1696,6 +1696,18 @@ def get_last_backup_id(self, status_filter=BackupManager.DEFAULT_STATUS_FILTER):
...
 
+    def get_last_full_backup_id(
+        self, status_filter=BackupManager.DEFAULT_STATUS_FILTER
+    ):
+        """
+        Get the id of the latest/last FULL backup in the catalog (if exists)
+
+        :param status_filter: The status of the backup to return,
+            default to DEFAULT_STATUS_FILTER.
+        :return string|None: ID of the backup
```
barthisrael, line 1707:
```suggestion
        :param status_filter: The status of the backup to return,
            default to :attr:`BackupManager.DEFAULT_STATUS_FILTER`.
        :return str|None: ID of the backup
```

barthisrael (changes requested):
Some sphinx related suggestions.

gcalacoci (approved):
LGTM

barman/backup.py
```diff
@@ -371,6 +371,27 @@ def get_last_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
...
+        Get the id of the latest/last FULL backup in the catalog (if exists)
+
+        :param status_filter: The status of the backup to return,
+            default to :attr:`DEFAULT_STATUS_FILTER`.
+        :return str|None: ID of the backup
+        """
+        available_full_backups = dict(
+            filter(
+                lambda backup: backup[1].is_full_and_eligible_for_incremental(),
+                self.get_available_backups(status_filter).items(),
```
barthisrael, line 385:
```suggestion
                lambda backup: backup.is_full_and_eligible_for_incremental(),
                self.get_available_backups(status_filter).values(),
```

I think this will look cleaner than `items` + `[1]`.

barthisrael (changes requested):

barthisrael (approved):
looks good.

PR #939, opened 2024-06-17
Check that a backup is valid when taking incremental backup

andremagui:
With the imminent introduction of Postgres core incremental backups support in Barman, it will become necessary to validate the possibility of taking an incremental backup.

Incremental backups establish a sequential series of backups, where the initial backup invariably constitutes a full backup. Subsequently, numerous instances may arise wherein both full and incremental backups are generated.

This PR will handle validation of the parent backup info before taking an incremental backup.

References: BAR-162

Signed-off-by: Andre [andre.marchesini@enterprisedb.com](mailto:andre.marchesini@enterprisedb.com)

### DIFF
--- a/barman/backup.py
+++ b/barman/backup.py
@@ -596,9 +596,9 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
         :raises BackupException: if a command argument is considered invalid
         """
         if "parent_backup_id" in kwargs:
-            self._validate_incremental_backup_configs()
+            self._validate_incremental_backup_configs(**kwargs)
 
-    def _validate_incremental_backup_configs(self):
+    def _validate_incremental_backup_configs(self, **kwargs):
         """
         Check required configurations for a Postgres incremental backup
 
@@ -616,6 +616,16 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
             raise BackupException(
                 "'summarize_wal' option has to be enabled in the Postgres server "
                 "to perform an incremental backup using the Postgres backup method"
+            )
+
+        parent_backup_id = kwargs.get("parent_backup_id")
+        parent_backup_info = self.get_backup(parent_backup_id)
+
+        if parent_backup_info and parent_backup_info.summarize_wal != "on":
+            raise BackupException(
+                "The specified backup is not eligible as a parent for an "
+                "incremental backup because WAL summaries were not enabled "
+                "when that backup was taken."
             )
 
     def backup(self, wait=False, wait_timeout=None, name=None, **kwargs):
--- a/tests/test_backup.py
+++ b/tests/test_backup.py
@@ -922,39 +922,38 @@ class TestBackup(object):
             None,
             "off",
             "on",
+            "on",
         ]
+
+        err_msg = "'summarize_wal' option has to be enabled in the Postgres server "
+        "to perform an incremental backup using the Postgres backup method"
+
         # ensure incremental backup with summarize_wal disabled raises exception
-        with pytest.raises(BackupException):
+        with pytest.raises(BackupException, match=err_msg):
             backup_manager._validate_incremental_backup_configs()
-        with pytest.raises(BackupException):
+        with pytest.raises(BackupException, match=err_msg):
             backup_manager._validate_incremental_backup_configs()
         # no exception is raised when summarize_wal is on
         backup_manager._validate_incremental_backup_configs()
 
-    @pytest.mark.parametrize(
-        ("backup_method", "should_raise_ex"),
-        [
-            ("any_non_postgres_method", True),
-            ("postgres", False),
-        ],
-    )
-    def test_validate_incremental_backup_configs_postgres_method(
-        self,
-        backup_method,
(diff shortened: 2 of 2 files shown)


barman/backup_executor.py
```diff
@@ -616,6 +617,8 @@ def backup_copy(self, backup_info):
         if parent_backup_info:
             parent_backup_manifest_path = parent_backup_info.get_backup_manifest_path()
 
+            self._raise_summarize_wal_value_from_parent_backup(parent_backup_info)
```
andremagui, line 620: I thought about doing this here because there was already a parent_backup_info. But now i am starting to think if it makes sense to move this to PostgresBackupStrategy.start_backup which is called first thing in the PostgresBackupExecutor. The only thing is that if exception is raised, the backup_info will not be saved apparently.

barthisrael: I'd say to perform the check here: https://github.com/EnterpriseDB/barman/blob/master/barman/backup.py#L601-L612

That way we avoid creating a leftover backup.info (failed). We can use the backup manager to get the backup info of the parent backup there, and check if summarize WAL was enabled for the parent.

barthisrael (changes requested):

barman/backup_executor.py
andremagui: Agreed

barman/backup_executor.py
```diff
@@ -671,6 +674,28 @@ def backup_copy(self, backup_info):
...
+        if parent_backup_info.summarize_wal:
+            if parent_backup_info.summarize_wal == "off":
+                raise BackupException(
+                    "summarize_wal is 'off' for the parent backup id. "
+                    "Executing incremental backups require for it to be "
+                    "set to 'on' in your PostgreSQL server before taking "
+                    "the first full backup of the chain."
+                )
+        else:
+            raise PostgresUnsupportedFeature(
```
andremagui, line 693: Does it make sense to check both None and "off"? And do the errors make sense as well? @gcalacoci @barthisrael @gustabowill @bleidens

andremagui: Here i thought about splitting because None is a PG_VERSION issue and off would be just a config issue!

barthisrael: Maybe we can simply test `!= "on"` through this PR.
There is another card to verify the version, which will likely be one more `if` statement around the logic provided by this PR.

gcalacoci: not a fan of `PostgresUnsupportedFeature` in this context. I would use  `BackupException` here too... this is a backup "lifecycle" error... not a postgres one. if you get what I mean.

andremagui: Yes, thats what it thought @barthisrael ! The pg_version would be higher in the validation! and yes, i got what you mean @gcalacoci! I will leave just one then!

barman/backup.py
```diff
@@ -611,6 +611,15 @@ def _validate_incremental_backup_configs(self):
                 "to perform an incremental backup using the Postgres backup method"
             )
 
+        parent_backup_id = kwargs.get("parent_backup_id")
+        parent_backup_info = self.get_backup(parent_backup_id)
+
+        if parent_backup_info and parent_backup_info.summarize_wal != "on":
+            raise BackupException(
+                "'summarize_wal' option has to be enabled in the Postgres server "
```
gcalacoci, line 619: The error here is different, the parent backup ID submitted is not valid because summarize_wal was not enabled when the backup has been executed, so the backup is not eligible for incremental.

The exception message should be changed

gcalacoci (changes requested):

barman/backup.py
andremagui: @gcalacoci I rewrote the error message to be like this, what do you think?
"'summarize_wal' needs to be enabled in the Postgres Server for all backups in the backup chain for incremental mode. The summarize_wal field o the parent backup was not enabled in the Postgres server when it was taken."

barthisrael: I'm not sure if it makes sense to mention a chain at this point, as a chain is more of a recovery thing.

Maybe we could have something like "The specified backup is not eligible as a parent for an incremental backup because WAL summaries were not enabled when that backup was taken." -- or something like that. Thoughts?

andremagui: agreed, sounds better!

gcalacoci (approved):
this LGTM

barman/backup.py
```diff
@@ -618,6 +618,16 @@ def _validate_incremental_backup_configs(self):
                 "to perform an incremental backup using the Postgres backup method"
             )
 
+        parent_backup_id = kwargs.get("parent_backup_id")
+        parent_backup_info = self.get_backup(parent_backup_id)
+
+        if parent_backup_info and parent_backup_info.summarize_wal != "on":
```
barthisrael, line 624: The `parent_backup_info` is expected to always be different of `None` at this point because `parse_backup_id` in the CLI already takes care of that.

But keeping the `if parent_backup_if` part of the `if` statement shouldn't hurt either.

Issue #881, opened 2023-12-18
After PITR recovery, Postgres won't start and report missing WAL file


PaganMuffin:
After doing PITR recovery with specific time Postgres won't start. 
Recovery command is reporting that postgres is successfully prepared for recovery, but after starting postgres instance it fail and log that WAL file, which is  after target time according to timeline, is missing.
I can do a restore to the latest version of DB without a problem. 

**Command used to recovery**
```
barman recover --target-time 20231215T162400 --remote-ssh-command 'ssh SERVER' remote 20231215T162003  /root/dev/pg-wal/pg-data/pgdata
```

**Command log**
```
Starting remote restore for server remote using backup 20231215T162003
Destination directory: /root/dev/pg-wal/pg-data/pgdata
Remote command: ssh SERVER
Doing PITR. Recovery target time: '2023-12-15 16:24:00+01:00'
Copying the base backup.
Copying required WAL segments.
Generating recovery configuration
Identify dangerous settings in destination directory.

Recovery completed (start time: 2023-12-18 08:39:26.886501+01:00, elapsed time: 33 seconds)
Your PostgreSQL server has been successfully prepared for recovery!
```

**Postgres Logs** 
```
PostgreSQL Database directory appears to contain a database; Skipping initialization

2023-12-18 07:41:06.770 UTC [1] LOG:  starting PostgreSQL 14.10 on x86_64-pc-linux-musl, compiled by gcc (Alpine 13.2.1_git20231014) 13.2.1 20231014, 64-bit
2023-12-18 07:41:06.771 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2023-12-18 07:41:06.771 UTC [1] LOG:  listening on IPv6 address "::", port 5432
2023-12-18 07:41:06.786 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2023-12-18 07:41:06.807 UTC [21] LOG:  database system was interrupted while in recovery at log time 2023-12-15 15:20:03 UTC
2023-12-18 07:41:06.807 UTC [21] HINT:  If this has occurred more than once some data might be corrupted and you might need to choose an earlier recovery target.
cp: can't stat 'barman_wal/00000002.history': No such file or directory
2023-12-18 07:41:06.887 UTC [21] LOG:  starting point-in-time recovery to 2023-12-15 16:24:00+00
[...]

PR #945, opened 2024-06-20
Always add an explicit time zone to `recovery_target_time`

barthisrael:
Previous to this commit there could be cases where Postgres would be misconfigured by `barman recover`.

For example, assume you had an environment like this:

* Barman host configured with time zone UTC+2, e.g. `Europe/Berlin`;`
* Postgres cluster with default value for timezone GUC, i.e. GMT value.

If we attempted a `barman recover` execution with `--target-time` set, but without an explicit time zone, recovery could fail:

* `barman recover` output would say that recovery target time was set with `+02:00`;
* However `postgresql.auto.conf` would not contain `+02:00`.

In practice either the recovery would fail because of an invalid recovery target, or the cluster would be recovered to a point-in-time different from the user needs. Neither of these cases are welcome.

This commit fixes that issue by reusing the `target_datetime` key from the `recovery_info` variable. That entry contains the target time with a time zone explicitly set, even if the user specified no time zone through the command-line option.

Note: a warning is issued if the user specified no time zone, and Barman assumed its host time zone.

References: BAR-185 #881.

### DIFF
--- a/barman/recovery_executor.py
+++ b/barman/recovery_executor.py
@@ -492,6 +492,12 @@ class RecoveryExecutor(object):
                         tzinfo=dateutil.tz.tzlocal()
                     )
 
+                    output.warning(
+                        "No time zone has been specified through '--target-time' "
+                        "command-line option. Barman assumed the same time zone from "
+                        "the Barman host.",
+                    )
+
                 # Check if the target time is reachable from the
                 # selected backup
                 if backup_info.end_time > target_datetime:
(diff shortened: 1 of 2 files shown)


gcalacoci (approved):
LGTM. Good catch.

barman/recovery_executor.py
```diff
@@ -1048,7 +1048,19 @@ def _generate_recovery_conf(
...
+            # option, which may be without a time zone. When writing the actual Postgres
+            # configuration we should use a value with an explicit time zone set, so we
+            # avoid hitting pitfalls. We use the 'target_datetime' which was prevously
+            # added to 'recovery_info'. It already handles the cases where the user
+            # specifies no time zone, and uses the Barman host time zone as a fallback.
+            # In short: if 'target_time' is present it means the user asked for a
+            # specific point in time, but we need a sanitized value to use in the
+            # Postgres configuration, so we use 'target_datetime'.
+            # See '_set_pitr_targets'.
+            recovery_conf_lines.append(
```
martinmarques, line 1061: We should emit a warning when user doesn't provide a TZ and barman appends it to the recovery time

martinmarques (changes requested):

barman/recovery_executor.py
barthisrael: Added that and squashed the commits.

gcalacoci (approved):
LGTM now with the warning

martinmarques (approved):

PR #947, opened 2024-06-21
Restrict the use of keep annotations to full backups only

bleidens:
An incremental backup with the keep annotation that has its parent deleted would become an orphan backup and therefore useless for recovery. To avoid this, we should restrict the use of keep annotations to full backups only. This adds a new check prior to starting the annotation process to make sure we do not allow keeping a backup if it has a parent backup, raising an error if so.

References: BAR-184

### DIFF
--- a/barman/cli.py
+++ b/barman/cli.py
@@ -1820,6 +1820,13 @@ def keep(args):
             ) % (backup_info.backup_id, backup_info.status)
             output.error(msg)
             output.close_and_exit()
+        if backup_info.parent_backup_id:
+            msg = (
+                "Unable to execute the keep command on backup %s: is an incremental backup.\n"
+                "Only full backups are eligible for the use of the keep command."
+            ) % (backup_info.backup_id)
+            output.error(msg)
+            output.close_and_exit()
         backup_manager.keep_backup(backup_info.backup_id, args.target)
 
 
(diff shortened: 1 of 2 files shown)


barman/cli.py
```diff
@@ -1820,6 +1820,15 @@ def keep(args):
             ) % (backup_info.backup_id, backup_info.status)
             output.error(msg)
             output.close_and_exit()
+        # To do: consider empty parent backups will also return None
+        parent_backup = backup_info.get_parent_backup_info()
+        if parent_backup is not None:
+            msg = (
+                "Cannot add keep to backup %s because it has a parent backup %s. "
```
gcalacoci, line 1827:
```suggestion
                "Unable to apply the keep flag to backup %s: is an incremental backup. "
```

barman/cli.py
```diff
@@ -1820,6 +1820,15 @@ def keep(args):
...
+                "Only full backups can be kept."
```
gcalacoci, line 1828:
```suggestion
                "Only full backups are eligible to the use of the keep flag."
```

barman/cli.py
```diff
@@ -1820,6 +1820,15 @@ def keep(args):
...
+            ) % (backup_info.backup_id, parent_backup.backup_id)
```
gcalacoci, line 1829:
```suggestion
            ) % (backup_info.backup_id)
```

gcalacoci (changes requested):
I've reworded the error message.

Regarding your doubts about the location of the check, IMHO your reasoning about applying the check before starting any annotation is valid and in general, the sooner we identify issues like this one, the better.

barthisrael (changes requested):
The error messages was changed -- unit tests need to be changed accordingly.
After that, squash commits.

barthisrael (approved):
Looks good.

gcalacoci (approved):
LGTM now

PR #950, opened 2024-06-24
Modify retention policies so incremental backups are ignored

andremagui:
Starting from Postgres version 17, the introduction of incremental backups ensures a continuous chain of backups beginning with a full backup and extending through several incremental backups, specifically when backup_method="postgres". Consequently, any retention policy should exclusively target full backups for
deletion, as incremental backups rely on their parent chain of backups to remain valid and usable.

This commit:
1. Add new method to RetentionPolicy class
	1.1 propagate_retention_status_to_children

- This new method is used to set the retention status for incremental backups. This is needed because we want to control the retention status of the incremental backup based on its parent backup (AKA full backup). This will propagate the parent retention status to its children incremental backups. For KEEP status of the parent, we propagate the VALID status to the incrementals since the incrementals will always depend on the full backup.

2. Modify method _backup_report of RecoveryWindowRetentionPolicy and RedundancyRetentionPolicy

- These were modified so we could add the logic for skipping incremental backups and only account full backups for retention policy. Whatever status the full backup is reported, the value is propagated to its children incremental backups (KEEP status for parent will propagate VALID for incrementals).

4. Add properties to infofile.py LocalBackupInfo
	3.1 is_incremental

- 	This property was created to verify if the BackupInfo is an incremental backup.

	3.2 has_children

- 	This property was created to verify if the BackupInfo has any children (incremental backups).

5. Add new parameter to walk_backups_tree in infofile.py

- This parameter was added so we can control whether to return the current backup info (self) calling the method.

References: BAR-166

Signed-off-by: Andre <andre.marchesini@enterprisedb.com>

### DIFF
--- a/barman/infofile.py
+++ b/barman/infofile.py
@@ -731,6 +731,24 @@ class LocalBackupInfo(BackupInfo):
                 e,
             )
 
+    @property
+    def is_incremental(self):
+        """
+        Only checks if the backup_info is an incremental backup
+
+        :return bool: ``True`` if this backup has a parent, ``False`` otherwise.
+        """
+        return self.parent_backup_id is not None
+
+    @property
+    def has_children(self):
+        """
+        Only checks if the backup_info has children
+
+        :return bool: ``True`` if this backup has at least one child, ``False`` otherwise.
+        """
+        return self.children_backup_ids is not None
+
     def get_list_of_files(self, target):
         """
         Get the list of files for the current backup
@@ -878,16 +896,19 @@ class LocalBackupInfo(BackupInfo):
 
         return None
 
-    def walk_backups_tree(self):
+    def walk_backups_tree(self, return_self=True):
         """
         Walk through all the children backups of the current backup.
 
         .. note::
-            The objects are returned with a bottom-up approach, including all children
-            backups plus the caller backup.
+            The objects are returned with a bottom-up approach, including all
+            children backups plus the caller backup.
 
-        :yield: a generator of :class:`LocalBackupInfo` objects for each backup, walking
-           from the leaves to self.
+        :param bool return_self: Whether to return the current backup.
+            Default to ``True``.
+
+        :yields: a generator of :class:`LocalBackupInfo` objects for each
+            backup, walking from the leaves to self.
         """
 
         if self.children_backup_ids:
@@ -897,6 +918,8 @@ class LocalBackupInfo(BackupInfo):
                     backup_id=child_backup_id,
                 )
                 yield from backup_info.walk_backups_tree()
+        if not return_self:
+            return
         yield self
 
     def walk_to_root(self):
--- a/barman/retention_policies.py
+++ b/barman/retention_policies.py
@@ -126,6 +126,33 @@ class RetentionPolicy(with_metaclass(ABCMeta, object)):
         """
         return "%s %s %s" % (self.mode, self.value, self.unit)
 
+    def _propagate_retention_status_to_children(self, backup_info, report, ret_status):
+        """
+        Propagate retention status to all backups in the tree.
+
+        .. note::
+            This has a side-effect. It modifies or add data to *report* dict.
+
+        :param barman.infofile.BackupInfo backup_info: The object we want to
+            propagate the RETENTION STATUS from.
+        :param dict[str, str] report: The report data structure to be modified.
+            Each key is the ID of a backup, and its value is the retention status
+            of that backup.
+        :param str ret_status:  The status of the backup according to retention
+            policies
+        """
+        # As KEEP status doesn't make sense for incremental backups, we simply
+        # set them as VALID if their root full backup has KEEP annotation
+        if ret_status in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
+            ret_status = BackupInfo.VALID
+        backup_tree = backup_info.walk_backups_tree(return_self=False)
+        for backup in backup_tree:
+            report[backup.backup_id] = ret_status
+            _logger.debug(
+                "Propagating %s retention status of backup %s to %s."
+                % (ret_status, backup_info.backup_id, backup.backup_id)
+            )
+
 
 class RedundancyRetentionPolicy(RetentionPolicy):
     """
@@ -170,6 +197,13 @@ class RedundancyRetentionPolicy(RetentionPolicy):
         # NOTE: reverse key orders (simulate reverse chronology)
         i = 0
         for bid in sorted(backups.keys(), reverse=True):
+            if backups[bid].is_incremental:
+                _logger.debug(
+                    "Ignoring incremental backup %s. The retention status will"
+                    " be propagated from %s."
+                    % (backups[bid], backups[bid].parent_backup_id)
+                )
+                continue
             if backups[bid].status == BackupInfo.DONE:
                 keep_target = self.server.get_keep_target(bid)
                 if keep_target == KeepManager.TARGET_STANDALONE:
@@ -185,6 +219,14 @@ class RedundancyRetentionPolicy(RetentionPolicy):
                 i = i + 1
             else:
                 report[bid] = BackupInfo.NONE
+
+            if backups[bid].has_children:
+                status = report[bid]
+                self._propagate_retention_status_to_children(
+                    backup_info=backups[bid],
+                    report=report,
+                    ret_status=status,
+                )
         return report
 
     def _wal_report(self):
@@ -271,6 +313,13 @@ class RecoveryWindowRetentionPolicy(RetentionPolicy):
         valid = 0
         # NOTE: reverse key orders (simulate reverse chronology)
         for bid in sorted(backups.keys(), reverse=True):
+            if backups[bid].is_incremental:
+                _logger.debug(
+                    "Ignoring incremental backup %s. The retention status will"
+                    " be propagated from %s."
+                    % (backups[bid], backups[bid].parent_backup_id)
+                )
+                continue
             # We are interested in DONE backups only
             if backups[bid].status == BackupInfo.DONE:
                 keep_target = self.server.get_keep_target(bid)
@@ -350,6 +399,14 @@ class RecoveryWindowRetentionPolicy(RetentionPolicy):
                         found = True
             else:
                 report[bid] = BackupInfo.NONE
+
+            if backups[bid].has_children:
+                status = report[bid]
+                self._propagate_retention_status_to_children(
+                    backup_info=backups[bid],
+                    report=report,
+                    ret_status=status,
+                )
         return report
 
     def _wal_report(self):
--- a/tests/test_barman_cloud_backup_delete.py
+++ b/tests/test_barman_cloud_backup_delete.py
@@ -103,12 +103,57 @@ class TestCloudBackupDelete(object):
             for name, path in file_paths.items()
         )
 
-    def _create_backup_metadata(
-        self, backup_ids, begin_wals=None, end_wals=None, is_snapshot_backup=False
+    def _create_backup_info(
+        self,
+        backup_id,
+        begin_wal=None,
+        end_wal=None,
+        backup_name=None,
+        parent_backup_id=None,
+        children_backup_ids=None,
+        is_incremental=False,
+        has_children=False,
     ):
         """
-        Helper for tests which creates mock BackupFileInfo and BackupInfo objects
-        which are returned in a dict keyed by backup_id.
+        Helper for tests which creates mock BackupInfo objects.
+
+        If *begin_wal* is set then the backup's begin_wal and
+        timeline values will be set according to the wal name.
+
+        This is used by tests for two purposes:
+          1. Helper to create a backup_metadata and mock CloudBackupCatalog.
+          2. Providing the data needed to create backup_metadata.
+        """
+        backup_info = mock.MagicMock(name="backup_info", snapshots_info=None)
+        backup_info.backup_id = backup_id
+        if backup_name:
+            backup_info.backup_name = backup_name
+        backup_info.status = "DONE"
+        backup_info.end_time = datetime.datetime.strptime(
+            backup_id, "%Y%m%dT%H%M%S"
+        ) + datetime.timedelta(hours=1)
+
+        backup_info.begin_wal = begin_wal
+        backup_info.timeline = None
+        if backup_info.begin_wal:
+            backup_info.timeline = int(backup_info.begin_wal[:8])
+        backup_info.end_wal = end_wal
+        backup_info.is_incremental = is_incremental
+        backup_info.has_children = has_children
+        backup_info.parent_backup_id = parent_backup_id
+        backup_info.children_backups_ids = children_backup_ids
+        return backup_info
+
+    def _create_backup_metadata(
+        self,
+        backup_ids,
+        begin_wals=None,
+        end_wals=None,
+        is_snapshot_backup=False,
+    ):
+        """
+        Helper for tests which creates mock BackupFileInfo which are
+        returned in a dict keyed by backup_id.
 
         If begin_wals has an entry for a given backup_id then the begin_wal and
         timeline values will be set according to the wal name in begin_wals.
@@ -121,6 +166,7 @@ class TestCloudBackupDelete(object):
             begin_wals = {}
         if end_wals is None:
             end_wals = {}
+
         backup_metadata = {}
         for backup in backup_ids:
             backup_name = None
@@ -141,23 +187,12 @@ class TestCloudBackupDelete(object):
                     }
                 )
             backup_metadata[backup_id]["files"] = mock_backup_files
-            backup_info = mock.MagicMock(name="backup_info", snapshots_info=None)
-            backup_info.backup_id = backup_id
-            if backup_name:
-                backup_info.backup_name = backup_name
-            backup_info.status = "DONE"
-            backup_info.end_time = datetime.datetime.strptime(
-                backup_id, "%Y%m%dT%H%M%S"
-            ) + datetime.timedelta(hours=1)
-            try:
-                backup_info.begin_wal = begin_wals[backup_id]
-                backup_info.timeline = int(backup_info.begin_wal[:8])
-            except KeyError:
-                pass
-            try:
-                backup_info.end_wal = end_wals[backup_id]
-            except KeyError:
-                pass
+            backup_info = self._create_backup_info(
+                backup_id,
+                begin_wals.get(backup_id),
+                end_wals.get(backup_id),
+                backup_name,
+            )
             backup_metadata[backup_id]["info"] = backup_info
         return backup_metadata
 
--- a/tests/test_retention_policies.py
+++ b/tests/test_retention_policies.py
@@ -261,6 +261,12 @@ class TestRetentionPolicies(object):
         assert empty_report_child == BackupInfo.NONE
 
     def test_first_backup(self, server):
+        """
+        Basic unit test of method first_backup
+
+        This method tests the retrieval of the first backup using both
+        RedundancyRetentionPolicy and RecoveryWindowRetentionPolicy
+        """
         rp = RetentionPolicyFactory.create(
             "retention_policy", "RECOVERY WINDOW OF 4 WEEKS", server
         )
@@ -268,12 +274,21 @@ class TestRetentionPolicies(object):
 
         # Build a BackupInfo object with status to DONE
         backup_info = build_test_backup_info(
-            server=server, backup_id="test0", end_time=datetime.now(tzlocal())
+            server=server,
+            backup_id="test0",
+            end_time=datetime.now(tzlocal()) - timedelta(days=1),
+        )
+        # Build another BackupInfo object with status to DONE taken one day after
+        backup_info2 = build_test_backup_info(
+            server=server, backup_id="test1", end_time=datetime.now(tzlocal())
         )
 
         # instruct the get_available_backups method to return a map with
         # our mock as result and minimum_redundancy = 1
-        server.get_available_backups.return_value = {"test_backup": backup_info}
+        server.get_available_backups.return_value = {
+            "test_backup": backup_info,
+            "test_backup2": backup_info2,
+        }
         server.config.minimum_redundancy = 1
         # execute retention policy report
         report = rp.first_backup()
@@ -285,14 +300,12 @@ class TestRetentionPolicies(object):
         )
         assert isinstance(rp, RedundancyRetentionPolicy)
 
-        # Build a BackupInfo object with status to DONE
-        backup_info = build_test_backup_info(
-            server=server, backup_id="test1", end_time=datetime.now(tzlocal())
-        )
-
         # instruct the get_available_backups method to return a map with
         # our mock as result and minimum_redundancy = 1
-        server.get_available_backups.return_value = {"test_backup": backup_info}
+        server.get_available_backups.return_value = {
+            "test_backup": backup_info,
+            "test_backup2": backup_info2,
+        }
         server.config.minimum_redundancy = 1
 
         # execute retention policy report
--- a/tests/test_retention_policies.py
+++ b/tests/test_retention_policies.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Barman.  If not, see <http://www.gnu.org/licenses/>.
 
+import itertools
 import logging
 import re
 from datetime import datetime, timedelta
@@ -35,6 +36,7 @@ from testing_helpers import build_mocked_server, build_test_backup_info
 
 
 class TestRetentionPolicies(object):
+
     @pytest.fixture
     def server(self):
         backup_manager = mock.Mock()
@@ -180,49 +182,83 @@ class TestRetentionPolicies(object):
 
         # Build a BackupInfo object with status to DONE
         backup_info = build_test_backup_info(
-            server=server, backup_id="test1", end_time=datetime.now(tzlocal())
+            server=server,
+            backup_id="test_backup",
+            end_time=datetime.now(tzlocal()),
+            parent_backup_id=None,
+            children_backup_ids=["test_backup_child"],
+        )
+
+        # Build a CHILD BackupInfo object with status to DONE
+        child_backup_info = build_test_backup_info(
+            server=server,
+            backup_id="test_backup_child",
+            end_time=datetime.now(tzlocal()) + timedelta(days=1),
+            parent_backup_id="test_backup",
+            children_backup_ids=None,
         )
 
         # instruct the get_available_backups method to return a map with
         # our mock as result and minimum_redundancy = 1
-        server.get_available_backups.return_value = {"test_backup": backup_info}
+        server.get_available_backups.return_value = {
+            "test_backup": backup_info,
+            "test_backup_child": child_backup_info,
+        }
         server.config.minimum_redundancy = 1
-        # execute retention policy report
-        report = rp.backup_status("test_backup")
 
-        assert report == "VALID"
+        # execute retention policy report on parent
+        report_parent = rp.backup_status("test_backup")
+
+        assert report_parent == "VALID"
+
+        # execute retention policy report on child
+        report_child = rp.backup_status("test_backup_child")
+
+        assert report_child == "VALID"
         # Force context of retention policy for testing purposes.
         # Expect the method to return a BackupInfo.NONE value
         rp.context = "invalid"
-        empty_report = rp.backup_status("test_backup")
+        empty_report_parent = rp.backup_status("test_backup")
 
-        assert empty_report == BackupInfo.NONE
+        assert empty_report_parent == BackupInfo.NONE
+
+        empty_report_child = rp.backup_status("test_backup_child")
+
+        assert empty_report_child == BackupInfo.NONE
 
         rp = RetentionPolicyFactory.create(
             "retention_policy", "RECOVERY WINDOW OF 4 WEEKS", server=server
         )
         assert isinstance(rp, RecoveryWindowRetentionPolicy)
 
-        # Build a BackupInfo object with status to DONE
-        backup_info = build_test_backup_info(
-            server=server, backup_id="test1", end_time=datetime.now(tzlocal())
-        )
-
         # instruct the get_available_backups method to return a map with
         # our mock as result and minimum_redundancy = 1
-        server.get_available_backups.return_value = {"test_backup": backup_info}
+        server.get_available_backups.return_value = {
+            "test_backup": backup_info,
+            "test_backup_child": child_backup_info,
+        }
         server.config.minimum_redundancy = 1
-        # execute retention policy report
-        report = rp.backup_status("test_backup")
 
-        assert report == "VALID"
+        # execute retention policy report on parent
+        report_parent = rp.backup_status("test_backup")
+
+        assert report_parent == "VALID"
+
+        # execute retention policy report on child
+        report_child = rp.backup_status("test_backup_child")
+
+        assert report_child == "VALID"
 
         # Force context of retention policy for testing purposes.
         # Expect the method to return a BackupInfo.NONE value
         rp.context = "invalid"
-        empty_report = rp.backup_status("test_backup")
+        empty_report_parent = rp.backup_status("test_backup")
 
-        assert empty_report == BackupInfo.NONE
+        assert empty_report_parent == BackupInfo.NONE
+
+        empty_report_child = rp.backup_status("test_backup_child")
+
+        assert empty_report_child == BackupInfo.NONE
 
     def test_first_backup(self, server):
         rp = RetentionPolicyFactory.create(
@@ -263,6 +299,265 @@ class TestRetentionPolicies(object):
         report = rp.first_backup()
 
         assert report == "test_backup"
+
+    @pytest.mark.parametrize(
+        ("retention_policy", "retention_status"),
+        itertools.product(
+            ("RECOVERY WINDOW OF 4 WEEKS", "REDUNDANCY 2"),
+            (
+                BackupInfo.OBSOLETE,
+                BackupInfo.VALID,
+                BackupInfo.POTENTIALLY_OBSOLETE,
+                BackupInfo.KEEP_FULL,
+                BackupInfo.KEEP_STANDALONE,
+                BackupInfo.NONE,
+            ),
+        ),
+    )
+    @mock.patch("barman.retention_policies._logger.debug")
+    @mock.patch("barman.infofile.LocalBackupInfo.walk_backups_tree")
+    def test__propagate_retention_status_to_children(
+        self,
+        mock_walk_backups_tree,
+        mock_logger,
+        retention_policy,
+        retention_status,
+        server,
+        tmpdir,
+    ):
+        """
+        Unit test of method _propagate_retention_status_to_children
+        """
+
+        # Use this to Build a chain of incrementals BackupInfo objects in
+        # post-order up to the root.
+        chain = {
+            "b3": """parent_backup_id=b2
+                   children_backup_ids=None
+                   status=DONE""",
+            "b6": """parent_backup_id=b2
+                   children_backup_ids=None
+                   status=DONE""",
+            "b2": """parent_backup_id=root
+                   children_backup_ids=b3,b6
+                   status=DONE""",
+            "b5": """parent_backup_id=b4
+                   children_backup_ids=None
+                   status=DONE""",
+            "b4": """parent_backup_id=root
+                   children_backup_ids=b5
+                   status=DONE""",
+            "root": """parent_backup_id=None
+                     children_backup_ids=b2,b4
+                     status=DONE""",
+        }
+        backup_chain = {}
+        for bkp in chain:
+            infofile = tmpdir.mkdir(bkp).join("backup.info")
+            infofile.write(chain[bkp])
+            b_info = build_test_backup_info(
+                backup_id=bkp,
+                server=server,
+            )
+            backup_chain[bkp] = b_info
+
+        root = backup_chain["root"]
+        mock_walk_backups_tree.return_value = iter(list(backup_chain.values())[:-1])
+
+        rp = RetentionPolicyFactory.create(
+            "retention_policy", retention_policy, server=server
+        )
+
+        report = {}
+        rp._propagate_retention_status_to_children(root, report, retention_status)
+
+        mock_walk_backups_tree.assert_called_once()
+
+        mock_logger.call_count == 5
+        # For full backups with status KEEP, we propagate VALID status to children
+        if retention_status in (BackupInfo.KEEP_FULL, BackupInfo.KEEP_STANDALONE):
+            retention_status = BackupInfo.VALID
+        for backup_id in report:
+            mock_logger.assert_any_call(
+                "Propagating %s retention status of backup root to %s."
+                % (retention_status, backup_id)
+            )
+
+        for backup in report:
+            assert report[backup] == retention_status
+
+    def test_redundancy_report_with_incrementals(self, server, caplog):
+        """
+        Test of the management of the minimum_redundancy parameter
+        into the backup_report method of the RedundancyRetentionPolicy class
+
+        """
+        rp = RetentionPolicyFactory.create(
+            "retention_policy", "REDUNDANCY 2", server=server
+        )
+        assert isinstance(rp, RedundancyRetentionPolicy)
+
+        backups_data = {
+            "20240628T000000": {
+                "parent_backup_id": None,
+                "children_backup_ids": ["20240628T120000"],
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=6, days=1),
+            },
+            "20240628T120000": {
+                "parent_backup_id": "20240628T000000",
+                "children_backup_ids": None,
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=6),
+            },
+            "20240629T000000": {
+                "parent_backup_id": None,
+                "children_backup_ids": ["20240629T120000"],
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=5, days=1),
+            },
+            "20240629T120000": {
+                "parent_backup_id": "20240629T000000",
+                "children_backup_ids": None,
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=5),
+            },
+            "20240630T000000": {
+                "parent_backup_id": None,
+                "children_backup_ids": None,
+                "end_time": datetime.now(tzlocal()),
+            },
+        }
+
+        available_backups = {}
+        for bkp_id, info in backups_data.items():
+            available_backups[bkp_id] = build_test_backup_info(
+                backup_id=bkp_id,
+                server=server,
+                parent_backup_id=info["parent_backup_id"],
+                children_backup_ids=info["children_backup_ids"],
+                end_time=info["end_time"],
+            )
+
+        # instruct the get_available_backups method to return a map with
+        # our mock as result and minimum_redundancy = 1
+        server.get_available_backups.return_value = available_backups
+
+        server.config.minimum_redundancy = 1
+
+        # execute retention policy report
+        report = rp.report()
+        # check that our mock is valid for the retention policy because
+        # the total number of valid backups is lower than the retention policy
+        # redundancy.
+        assert report == {
+            "20240630T000000": "VALID",
+            "20240629T000000": "VALID",
+            "20240629T120000": "VALID",
+            "20240628T000000": "OBSOLETE",
+            "20240628T120000": "OBSOLETE",
+        }
+
+        # Expect a ValueError if passed context is invalid
+        with pytest.raises(ValueError):
+            rp.report(context="invalid")
+        # Set a new minimum_redundancy parameter, enforcing the usage of the
+        # configuration parameter instead of the retention policy default
+        server.config.minimum_redundancy = 3
+        # execute retention policy report
+        rp.report()
+        # Check for the warning inside the log
+        caplog.set_level(logging.WARNING)
+
+        log = caplog.text
+        assert log.find(
+            "WARNING  Retention policy redundancy (2) "
+            "is lower than the required minimum redundancy (3). "
+            "Enforce 3."
+        )
+
+    def test_recovery_window_report_with_incrementals(self, server, caplog):
+        """
+        Test of the management of the minimum_redundancy parameter
+        into the backup_report method of the RecoveryWindowRetentionPolicy class
+
+        """
+        rp = RetentionPolicyFactory.create(
+            "retention_policy", "RECOVERY WINDOW OF 4 WEEKS", server=server
+        )
+        assert isinstance(rp, RecoveryWindowRetentionPolicy)
+
+        backups_data = {
+            "20240628T000000": {
+                "parent_backup_id": None,
+                "children_backup_ids": ["20240628T120000"],
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=6, days=1),
+            },
+            "20240628T120000": {
+                "parent_backup_id": "20240628T000000",
+                "children_backup_ids": None,
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=6),
+            },
+            "20240629T000000": {
+                "parent_backup_id": None,
+                "children_backup_ids": ["20240629T120000"],
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=5, days=1),
+            },
+            "20240629T120000": {
+                "parent_backup_id": "20240629T000000",
+                "children_backup_ids": None,
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=5),
+            },
+            "20240630T000000": {
+                "parent_backup_id": None,
+                "children_backup_ids": None,
+                "end_time": datetime.now(tzlocal()),
+            },
+        }
+
+        available_backups = {}
+        for bkp_id, info in backups_data.items():
+            available_backups[bkp_id] = build_test_backup_info(
+                backup_id=bkp_id,
+                server=server,
+                parent_backup_id=info["parent_backup_id"],
+                children_backup_ids=info["children_backup_ids"],
+                end_time=info["end_time"],
+            )
+
+        # instruct the get_available_backups method to return a map with
+        # our mock as result and minimum_redundancy = 1
+        server.get_available_backups.return_value = available_backups
+
+        server.config.minimum_redundancy = 1
+        server.config.name = "test"
+
+        # execute retention policy report
+        report = rp.report()
+        # check that our mock is valid for the retention policy because
+        # the total number of valid backups is lower than the retention policy
+        # redundancy.
+        assert report == {
+            "20240630T000000": "VALID",
+            "20240629T000000": "VALID",
+            "20240629T120000": "VALID",
+            "20240628T000000": "OBSOLETE",
+            "20240628T120000": "OBSOLETE",
+        }
+
+        # Expect a ValueError if passed context is invalid
+        with pytest.raises(ValueError):
+            rp.report(context="invalid")
+        # Set a new minimum_redundancy parameter, enforcing the usage of the
+        # configuration parameter instead of the retention policy default
+        server.config.minimum_redundancy = 4
+        # execute retention policy report
+        rp.report()
+        # Check for the warning inside the log
+        caplog.set_level(logging.WARNING)
+        log = caplog.text
+        warn = (
+            r"WARNING  .*Keeping obsolete backup 20240628T000000 for "
+            r"server test \(older than .*\) due to minimum redundancy "
+            r"requirements \(4\)\n"
+        )
+        assert re.search(warn, log)
 
 
 class TestRedundancyRetentionPolicyWithKeepAnnotation(object):
--- a/tests/test_infofile.py
+++ b/tests/test_infofile.py
@@ -1127,14 +1127,25 @@ class TestLocalBackupInfo:
             side_effect=provide_child_backup_info,
         ):
             # Call the `walk_backups_tree` method on the root backup info
-            backups = list(root_backup_info.walk_backups_tree())
+            backups = list(root_backup_info.walk_backups_tree(return_self=True))
             # Assert that the backups are returned in the correct order
             # We want to walk through the tree in a depth-first post order,
             # so leaf nodes are visited first, then their parent, and so on.
+            assert len(backups) == 4
             assert backups[0].backup_id == "child_backup3"
             assert backups[1].backup_id == "child_backup1"
             assert backups[2].backup_id == "child_backup2"
             assert backups[3].backup_id == "root_backup"
+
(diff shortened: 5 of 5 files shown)


andremagui:
Since this ticket is about ignoring incremental backups, i thought about implementing this in the retention_policies.py instead of doing it in the backup.py -> cron_retention_policy() method.

barman/backup.py
```diff
@@ -134,6 +134,21 @@ def get_available_backups(self, status_filter=DEFAULT_STATUS_FILTER):
                 backups[key] = value
         return backups
 
+    def get_available_full_backups(self, status_filter=DEFAULT_STATUS_FILTER):
```
gcalacoci, line 137: not sure about this. Have you thought about implementing this as an option for the `get_available_backups` method? something like:
```python
get_available_backups(self, status_filter=DEFAULT_STATUS_FILTER, incrementals=false)
```

Also please don't assume that the above is correct by default, search for possible implications in the code if you haven't already.

gcalacoci: just to be clear... my general sensation is that we are sorta duplicating code just adding `_full` in the name methods... duplicating is not even the right word here... is more... overloading? is clear what I mean?

barthisrael: Right, I've been taking a look and a I have a similar sensation. Maybe we ware able to modify the existing methods they look less like copy'n'paste.

andremagui: Yeah, i thought about modifying the existing get_avaialable_backups! I just thought about creating this other one so i would not change the signature of the other one and because this existing method is used in several places and i would end up having to change them all! But i agree about being a little repetitive here! Also this new method uses the existing one. If you guys agree, i can modify the existing method and remove this new one! What you guys think?

barman/retention_policies.py
```diff
@@ -455,8 +455,10 @@ def retention_policy(self):
         return self.server.config.retention_policy
 
     @property
-    def available_backups(self):
-        return self.server.get_available_backups(BackupInfo.STATUS_NOT_EMPTY)
+    def available_backups(self, ignore_incrementals):
```
andremagui, line 458: i guess properties cannot have parameters directly, right? So should i create another property available_full_backups? Or any ideas here?

barthisrael: Yeah, we cannot have parameters in properties :s

barthisrael (changes requested):
Posted a couple of suggestions.

barthisrael:
I tested this and the implementation is broken. I took the following backups:

```console
$ egrep "children|parent" */backup.info
20240627T001815/backup.info:children_backup_ids=20240627T001833
20240627T001815/backup.info:parent_backup_id=None
20240627T001833/backup.info:children_backup_ids=20240627T001836
20240627T001833/backup.info:parent_backup_id=20240627T001815
20240627T001836/backup.info:children_backup_ids=None
20240627T001836/backup.info:parent_backup_id=20240627T001833
20240627T001840/backup.info:children_backup_ids=20240627T001846
20240627T001840/backup.info:parent_backup_id=None
20240627T001846/backup.info:children_backup_ids=20240627T001849
20240627T001846/backup.info:parent_backup_id=20240627T001840
20240627T001849/backup.info:children_backup_ids=None
20240627T001849/backup.info:parent_backup_id=20240627T001846
20240627T001852/backup.info:children_backup_ids=20240627T002214
20240627T001852/backup.info:parent_backup_id=None
20240627T002214/backup.info:children_backup_ids=20240627T002217
20240627T002214/backup.info:parent_backup_id=20240627T001852
20240627T002217/backup.info:children_backup_ids=None
20240627T002217/backup.info:parent_backup_id=20240627T002214
```

When I attempt to run `barman list-backup` to check their status, the command fails:

```console
$ barman list-backup pg17
EXCEPTION: '20240627T002217'
See log file for more details.
```

The logs:

```console
$ tail -10 /var/log/barman/barman.log
2024-06-27 00:27:41,768 [154975] barman.cli ERROR: '20240627T002217'
See log file for more details.
  File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 2416, in main
    args.func(args)
  File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 595, in list_backups
    server.list_backups()
  File "/home/vagrant/Repositories/edb/barman/barman/server.py", line 1770, in list_backups
    and retention_status[backup.backup_id] != BackupInfo.VALID
KeyError: '20240627T002217'
```

The problem is `Server.list_backup` is comparing two different lists of backups:

* `backups` variable contains them all;
* `retention_status` variable contains only the full;

I think we should change the implementation here.

Instead of using `source = self.server.full_backups` in `barman.retention_policies.RetentionPolicy.report`, I think we should still use all of the backups. However, when it comes to actually taking them into consideration for retention policies, we should only care about full backups.

Said in other words:

* Ignore incremental backups when checking for redundancy or recovery window, and take only full backups into consideration for that purpose;
* Whenever a full backup is considered obsolete by the retention policy, mark all of its children in the tree as obsolete too.

Thoughts @andremagui @gcalacoci ?

gcalacoci:
> I think we should change the implementation here.
>
> Instead of using source = self.server.full_backups in barman.retention_policies.RetentionPolicy.report, I think we should still use all of the backups. However, when it comes to actually taking them into consideration for retention policies, we should only care about full backups.
>
> Said in other words:
>
> * Ignore incremental backups when checking for redundancy or recovery window, and take only full backups into consideration for that purpose;
> * Whenever a full backup is considered obsolete by the retention policy, mark all of its children in the tree as obsolete too.

Yeah totally makes sense and I agree

barman/backup.py
```diff
@@ -371,29 +377,6 @@ def get_last_backup_id(self, status_filter=DEFAULT_STATUS_FILTER):
...
-        )
-
-        if len(available_full_backups) == 0:
-            return None
-
-        backup_infos = sorted(
-            available_full_backups, key=lambda backup_info: backup_info.backup_id
-        )
-        return backup_infos[-1].backup_id
-
```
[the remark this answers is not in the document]
andremagui: Yes, i totally mixed up things here! Thanks for pointing it out!

barman/backup.py
```diff
@@ -112,12 +112,16 @@ def mode(self):
...
+    def get_available_backups(
+        self, status_filter=DEFAULT_STATUS_FILTER, ignore_incrementals=False
+    ):
         """
         Get a list of available backups
 
         :param status_filter: default DEFAULT_STATUS_FILTER. The status of
             the backup list returned
+        :param ignore_incrementals: default False. Ignore incremental backups.
+            When set to True, this will return only full backups.
```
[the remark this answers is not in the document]
andremagui: Goog call

andremagui:
> I tested this and the implementation is broken. I took the following backups:
> 
> ```
> $ egrep "children|parent" */backup.info
> 20240627T001815/backup.info:children_backup_ids=20240627T001833
> 20240627T001815/backup.info:parent_backup_id=None
> 20240627T001833/backup.info:children_backup_ids=20240627T001836
> 20240627T001833/backup.info:parent_backup_id=20240627T001815
> 20240627T001836/backup.info:children_backup_ids=None
> 20240627T001836/backup.info:parent_backup_id=20240627T001833
> 20240627T001840/backup.info:children_backup_ids=20240627T001846
> 20240627T001840/backup.info:parent_backup_id=None
> 20240627T001846/backup.info:children_backup_ids=20240627T001849
> 20240627T001846/backup.info:parent_backup_id=20240627T001840
> 20240627T001849/backup.info:children_backup_ids=None
> 20240627T001849/backup.info:parent_backup_id=20240627T001846
> 20240627T001852/backup.info:children_backup_ids=20240627T002214
> 20240627T001852/backup.info:parent_backup_id=None
> 20240627T002214/backup.info:children_backup_ids=20240627T002217
> 20240627T002214/backup.info:parent_backup_id=20240627T001852
> 20240627T002217/backup.info:children_backup_ids=None
> 20240627T002217/backup.info:parent_backup_id=20240627T002214
> ```
> 
> When I attempt to run `barman list-backup` to check their status, the command fails:
> 
> ```
> $ barman list-backup pg17
> EXCEPTION: '20240627T002217'
> See log file for more details.
> ```
> 
> The logs:
> 
> ```
> $ tail -10 /var/log/barman/barman.log
> 2024-06-27 00:27:41,768 [154975] barman.cli ERROR: '20240627T002217'
> See log file for more details.
> Traceback (most recent call last):
>   File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 2416, in main
>     args.func(args)
>   File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 595, in list_backups
>     server.list_backups()
>   File "/home/vagrant/Repositories/edb/barman/barman/server.py", line 1770, in list_backups
>     and retention_status[backup.backup_id] != BackupInfo.VALID
> KeyError: '20240627T002217'
> ```
> 
> The problem is `Server.list_backup` is comparing two different lists of backups:
> 
> * `backups` variable contains them all;
> * `retention_status` variable contains only the full;
> 
> I think we should change the implementation here.
> 
> Instead of using `source = self.server.full_backups` in `barman.retention_policies.RetentionPolicy.report`, I think we should still use all of the backups. However, when it comes to actually taking them into consideration for retention policies, we should only care about full backups.
> 
> Said in other words:
> 
> * Ignore incremental backups when checking for redundancy or recovery window, and take only full backups into consideration for that purpose;
> * Whenever a full backup is considered obsolete by the retention policy, mark all of its children in the tree as obsolete too.
> 
> Thoughts @andremagui @gcalacoci ?

Makes total sense and it is safer!

barman/backup.py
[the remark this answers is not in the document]
barthisrael: This might fall apart in the end, though 😆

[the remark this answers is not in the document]
andremagui: yeah

barman/retention_policies.py
```diff
@@ -126,6 +126,40 @@ def to_json(self):
...
+        len_format = 2 if len(backup_id_tree) <= 7 else 3
+        msg = "Propagating %s annotation of backup %s to n=%s children %s." % (
+            annotation,
+            backup_info.backup_id,
+            str(len(backup_id_tree)),
+            ",".join(backup_id_tree[:len_format])
+            + "..."
+            + ",".join(backup_id_tree[-len_format:]),
+        )
+        _logger.debug(msg)
```
andremagui, line 147: would this be enough? Any suggestion on how to get better logging? Perhaps in another place in the code?

barman/retention_policies.py
```diff
@@ -170,6 +204,12 @@ def _backup_report(self, source):
         # NOTE: reverse key orders (simulate reverse chronology)
         i = 0
         for bid in sorted(backups.keys(), reverse=True):
+            if self.has_parent(backups[bid]):
+                _logger.debug(
```
andremagui, line 208: I inserted this one as well for both classes

barman/retention_policies.py
```diff
@@ -126,6 +126,40 @@ def to_json(self):
...
+            str(len(backup_id_tree)),
+            ",".join(backup_id_tree[:len_format])
+            + "..."
+            + ",".join(backup_id_tree[-len_format:]),
+        )
+        _logger.debug(msg)
+
+    def has_children(self, backup_info):
+        return (
+            isinstance(backup_info, BackupInfo)
```
andremagui, line 151: I used all this checks because there were a lot of tests braking in the TestCloudBackupDelete class if they are not used. Took a lot of time to find that this was the problem braking the tests.

barman/retention_policies.py
```diff
@@ -126,6 +126,40 @@ def to_json(self):
...
+            and hasattr(backup_info, "children_backup_ids")
+            and backup_info.children_backup_ids is not None
+        )
+
+    def has_parent(self, backup_info):
+        return (
+            isinstance(backup_info, BackupInfo)
```
andremagui, line 158: also here

barthisrael:
Something seems still broken here.

* On master:

```console
$ barman list-backups pg17
pg17 20240628T000532 'yet_another_full' - Thu Jun 27 22:05:34 2024 - Size: 251.7 MiB - WAL Size: 0 B - WAITING_FOR_WALS
pg17 20240628T000524 'another_incremental' - Thu Jun 27 22:05:26 2024 - Size: 35.7 MiB - WAL Size: 32.0 MiB
pg17 20240628T000511 'another_full' - Thu Jun 27 22:05:13 2024 - Size: 251.7 MiB - WAL Size: 32.0 MiB
```

* On the dev branch:

```console
$ barman list-backups pg17
pg17 20240628T000532 'yet_another_full' - Thu Jun 27 22:05:34 2024 - Size: 251.7 MiB - WAL Size: 0 B - WAITING_FOR_WALS
EXCEPTION: '20240628T000524'
See log file for more details.
```

```console
$ tail -10 /var/log/barman/barman.log
2024-06-28 03:11:04,095 [28079] barman.cli ERROR: '20240628T000524'
See log file for more details.
  File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 2416, in main
    args.func(args)
  File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 595, in list_backups
    server.list_backups()
  File "/home/vagrant/Repositories/edb/barman/barman/server.py", line 1762, in list_backups
    and retention_status[backup.backup_id] != BackupInfo.VALID
KeyError: '20240628T000524'
```

barthisrael:
The issue mentioned through the previous reply occurs as soon as I have the first incremental backup.

barman/retention_policies.py
```diff
@@ -126,6 +126,59 @@ def to_json(self):
...
+            annotation,
+            backup_info.backup_id,
+            str(len(backup_id_tree)),
+            ",".join(backup_id_tree[:len_format])
+            + "..."
+            + ",".join(backup_id_tree[-len_format:]),
+        )
+        _logger.debug(msg)
+
+    def has_children(self, backup_info):
```
barthisrael, line 156: A method like this seems to fit best in `LocalBackupInfo`.
Maybe we could eliminate both `has_children` and `has_parent` and introduce a `LocalBackupInfo.is_incremental` method.

barman/retention_policies.py
```diff
@@ -126,6 +126,59 @@ def to_json(self):
...
+        :param barman.infofile.BackupInfo backup_info: The object we want to
+            do the check.
+        """
+        return (
+            isinstance(backup_info, BackupInfo)
+            and hasattr(backup_info, "children_backup_ids")
+            and backup_info.children_backup_ids is not None
+        )
+
+    def has_parent(self, backup_info):
```
barthisrael, line 169: A method like this seems to fit best in `LocalBackupInfo`.
Maybe we could eliminate both `has_children` and `has_parent` and introduce a `LocalBackupInfo.is_incremental` method.

barman/retention_policies.py
```diff
@@ -182,9 +241,20 @@ def _backup_report(self, source):
...
+                            backup_info=backups[bid],
+                            report=report,
+                            annotation=BackupInfo.OBSOLETE,
+                        )
+                    else:
+                        _logger.debug("No children to propagate annotation...")
                 i = i + 1
             else:
                 report[bid] = BackupInfo.NONE
+
```
barthisrael, line 257: Maybe we could change things a bit and process like this:

* Skip the backup in the beginning, as you are already doing -- maybe by using the new method of `LocalBackupInfo` that I mentioned in the previous comment;
* During the loop, try to identify the expected status for the children;
* After the loop, loop over `walk_backups_tree` and set the status accordingly

barman/retention_policies.py
```diff
@@ -333,6 +420,17 @@ def _backup_report(self, source):
...
+                            # If backup is postgres with incremental backup children
+                            # create tree o backups and make them all OBSOLETE.
+                            if self.has_children(backups[bid]):
+                                self.propagate_annotation_to_children(
+                                    backup_info=backups[bid],
+                                    report=report,
+                                    annotation=BackupInfo.OBSOLETE,
+                                )
+                            else:
+                                _logger.debug("No children to propagate annotation...")
```
barthisrael, line 433: Same comment here about how to process the loop.

barthisrael (changes requested):
As I mentioned before, the code is buggy.

Besides that, this feels like overcomplicated to me. I think we are able to get a cleaner and better performant solution.

I posted a few comments along the way with a few ideas.

Thoughts?

barman/retention_policies.py
```diff
@@ -126,6 +126,59 @@ def to_json(self):
         """
         return "%s %s %s" % (self.mode, self.value, self.unit)
 
+    def propagate_annotation_to_children(self, backup_info, report, annotation):
```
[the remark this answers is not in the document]
gcalacoci: agreed

barman/retention_policies.py
```diff
@@ -126,6 +126,59 @@ def to_json(self):
...
+            propagate annotations from.
+        :param report dict[str, str]: The report data structure to be modified.
+        :param annotation str:  The status of the backup according to retention
+            policies
+        """
+        backup_id_tree = [
+            backup.backup_id for backup in list(backup_info.walk_backups_tree())
+        ][:-1]
+        for _id in backup_id_tree:
+            report[_id] = annotation
```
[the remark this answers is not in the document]
gcalacoci: also is a bit difficult to read

barman/retention_policies.py
```diff
@@ -126,6 +126,59 @@ def to_json(self):
...
+        len_format = 2 if len(backup_id_tree) <= 7 else 3
+        msg = "Propagating %s annotation of backup %s to n=%s children %s." % (
+            annotation,
+            backup_info.backup_id,
+            str(len(backup_id_tree)),
+            ",".join(backup_id_tree[:len_format])
+            + "..."
+            + ",".join(backup_id_tree[-len_format:]),
+        )
```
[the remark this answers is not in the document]
gcalacoci: agreed

barman/retention_policies.py
gcalacoci: I do agree this should be on the backup info..
let's consider to have properties for all... we do check if a backup has children's or a parent often. and a boolean has_ something method or is_ something make things more readable/easy.

gcalacoci:
I added some comments to the PR. overall I do feel that we should sorta "restart" from the drawing board. I share @barthisrael sensation about this being inefficient and buggy and sometimes just starting over is the way to go

barman/retention_policies.py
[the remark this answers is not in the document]
barthisrael: Yeah, that thought also came to my mind at some point.

andremagui:
I have made the changes according to the reviews! I will rebase everything afterwards when all is OK! @barthisrael @gcalacoci I will still look into some tests that might need to be added to this

andremagui:
> Something seems still broken here.
> 
> * On master:
> 
> ```
> $ barman list-backups pg17
> pg17 20240628T000532 'yet_another_full' - Thu Jun 27 22:05:34 2024 - Size: 251.7 MiB - WAL Size: 0 B - WAITING_FOR_WALS
> pg17 20240628T000524 'another_incremental' - Thu Jun 27 22:05:26 2024 - Size: 35.7 MiB - WAL Size: 32.0 MiB
> pg17 20240628T000511 'another_full' - Thu Jun 27 22:05:13 2024 - Size: 251.7 MiB - WAL Size: 32.0 MiB
> ```
> 
> * On the dev branch:
> 
> ```
> $ barman list-backups pg17
> pg17 20240628T000532 'yet_another_full' - Thu Jun 27 22:05:34 2024 - Size: 251.7 MiB - WAL Size: 0 B - WAITING_FOR_WALS
> EXCEPTION: '20240628T000524'
> See log file for more details.
> ```
> 
> ```
> $ tail -10 /var/log/barman/barman.log
> 2024-06-28 03:11:04,095 [28079] barman.cli ERROR: '20240628T000524'
> See log file for more details.
> Traceback (most recent call last):
>   File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 2416, in main
>     args.func(args)
>   File "/home/vagrant/Repositories/edb/barman/barman/cli.py", line 595, in list_backups
>     server.list_backups()
>   File "/home/vagrant/Repositories/edb/barman/barman/server.py", line 1762, in list_backups
>     and retention_status[backup.backup_id] != BackupInfo.VALID
> KeyError: '20240628T000524'
> ```

Hey this bug is related to the fact that list backups uses the report in the beginning and it checks the available_backups with the ones that were reported. So actually we need to add all incrementals as well in the report or change the list_backups method from Server. Or even create another retention status for incrementals that are not touched and add them to the report. What are your thoughts on this?

barman/infofile.py
```diff
@@ -878,7 +898,7 @@ def get_child_backup_info(self, child_backup_id):
 
         return None
 
-    def walk_backups_tree(self):
+    def walk_backups_tree(self, return_root=True):
```
andremagui, line 901: I ended up touching this function! Would it be ok to make the modification in this PR?

barman/retention_policies.py
```diff
@@ -170,6 +189,12 @@ def _backup_report(self, source):
         # NOTE: reverse key orders (simulate reverse chronology)
         i = 0
         for bid in sorted(backups.keys(), reverse=True):
+            if backups[bid].is_incremental and len(backups) > 1:
```
andremagui, line 192: I added this `len(backups) > 1` because of the `backup_status` method. It takes into account a dict with only one element! If it happens to be an incremental backup, it will be skipped and we will not get the expected result of getting the status of the backup!

barman/retention_policies.py
```diff
@@ -185,6 +210,16 @@ def _backup_report(self, source):
                 i = i + 1
             else:
                 report[bid] = BackupInfo.NONE
+
+            if backups[bid].has_children:
+                status = report[bid]
+                if report[bid] in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
```
andremagui, line 216: Here, if the parent has children but reported as KEEP_*, we have to propagate some status to the children, otherwise we will end up having that bug in list_buckets. list_buckets takes into account all available_backups. I thought that if parent has KEEP status, the incremental should be VALID since we only delete incrementals through the deletion of its parent. As soon as a parent gets obsolete, automatically the children will change status as well. So when a parent backup is VALID or KEEP_STANDALONE or KEEP_FULL, the children will be VALID. 
Please, what are your thoughts on this? Would this make sense?

barman/retention_policies.py
```diff
@@ -271,6 +306,12 @@ def _backup_report(self, source):
         valid = 0
         # NOTE: reverse key orders (simulate reverse chronology)
         for bid in sorted(backups.keys(), reverse=True):
+            if backups[bid].is_incremental and len(backups) > 1:
```
andremagui, line 309: same here

barman/retention_policies.py
```diff
@@ -350,6 +391,16 @@ def _backup_report(self, source):
                         found = True
             else:
                 report[bid] = BackupInfo.NONE
+
+            if backups[bid].has_children:
+                status = report[bid]
+                if report[bid] in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
```
andremagui, line 397: Also here

tests/test_backup.py
```diff
@@ -743,6 +743,79 @@ def test_cron_retention_only_deletes_OBSOLETE_backups(
...
+            call(
+                available_backups["obsolete_backup"],
+                skip_wal_cleanup_if_standalone=False,
+            ),
+        ]
+        delete_backup.assert_has_calls(expected_calls, any_order=False)
+
     @patch("barman.backup.BackupManager.delete_backup")
     @patch("barman.backup.BackupManager.get_available_backups")
     def test_cron_retention_skip_OBSOLETE_backups_if_lock(
```
andremagui, line 821: I was going to create a test for this as well but i thought that what is testing is more related to the fact that an OBSOLETE backup  should be skipped on retetion when a lock is acquired. So the delete will not happen

tests/test_barman_cloud_backup_delete.py
```diff
@@ -103,24 +103,73 @@ def _get_mock_backup_files(self, file_paths):
...
+            backup_info.end_wal = end_wals[backup_id]
+        except KeyError:
+            pass
+        backup_info.is_incremental = is_incremental
+        backup_info.has_children = has_children
+        backup_info.parent_backup_id = parent_backup_id
+        backup_info.children_backups_ids = children_backup_ids
+        return backup_info
+
+    def _create_backup_metadata(
```
andremagui, line 154: Refactored this a bit to have more flexibility on creating backup infos.

barman/infofile.py
```diff
@@ -731,6 +731,26 @@ def __init__(self, server, info_file=None, backup_id=None, **kwargs):
                 e,
             )
 
+    @property
+    def is_incremental(self):
+        """
+        Only checks if the backup_info is an incremental backup
+
+        :param barman.infofile.BackupInfo backup_info: The object we want to
+            do the check.
```
barthisrael, line 740:
```suggestion
        :return bool: ``True`` if this backup is incremental, ``False`` otherwise.
```

There is no `backup_info` param, and we are missing the return value.

barman/infofile.py
```diff
@@ -731,6 +731,26 @@ def __init__(self, server, info_file=None, backup_id=None, **kwargs):
...
+        """
+        return self.parent_backup_id is not None
+
+    @property
+    def has_children(self):
+        """
+        Only checks if the backup_info has children
+
+        :param barman.infofile.BackupInfo backup_info: The object we want to
+            do the check.
```
barthisrael, line 750:
```suggestion
        :return bool: ``True`` if this backup has at least one child, ``False`` otherwise.
```

There is no `backup_info` param, and we are missing the return value.

barman/infofile.py
barthisrael: I think so -- we are going to use it here after all, and the original ticket used to create this method has already been merged.

barman/retention_policies.py
```diff
@@ -126,6 +126,25 @@ def to_json(self):
         """
         return "%s %s %s" % (self.mode, self.value, self.unit)
 
+    def propagate_retention_status_to_children(self, backup_info, report, ret_status):
+        """
+        Propagate retention status to all backups in the chain.
+        This has a side-effect. It modifies or add data to report dict.
```
barthisrael, line 132:
```suggestion
        Propagate retention status to all backups in the chain.

        .. note::
            This has a side-effect. It modifies or add data to *report* dict.
```

This sounds more like a note.
Besides that, sphinx recommend using `*` for quoting parameters.

barman/retention_policies.py
```diff
@@ -126,6 +126,25 @@ def to_json(self):
...
+
+        :param barman.infofile.BackupInfo backup_info: The object we want to
+            propagate the RETENTION STATUS from.
+        :param report dict[str, str]: The report data structure to be modified.
```
barthisrael, line 136:
```suggestion
        :param report dict[str, str]: The report data structure to be modified.
            Each key is the ID of a backup, and its value is the retention status
            of that backup.
```

barman/retention_policies.py
barthisrael: I'm not sure if I understand your concerns.

That should work fine at a first glance. I mean, this is the code:

```python
    def backup_status(self, backup_id):
        """Report the status of a backup according to the retention policy"""
        source = self.server.available_backups
        if self.context == "BASE":
            return self._backup_report(source)[backup_id]
```

It runs `_backup_report` passing all available backups, and returns a single key. IAC the `_backup_report` is still called with them all.

barman/retention_policies.py
barthisrael, line 216:
```suggestion
                if status in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
```

We can skip the lookup in this line 😆

barthisrael: At a first glance I think it makes sense to have keep backups' children as `VALID`.

barman/retention_policies.py
barthisrael: As I mentioned before, it doesn't make sense to me based on my understanding of the code.

barman/retention_policies.py
barthisrael: Right, makes sense at a first glance.

barthisrael (changes requested):
I still need to check the unit tests, but I'm providing some thoughts on the source code.

This looks much more promising now!

barman/retention_policies.py
barthisrael:
```suggestion
                if status in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
```

Same comment about the lookup.

barman/infofile.py
gcalacoci: yeah seems good to me doing it here

barman/retention_policies.py
gcalacoci: Yeah VALID is the way to go imho.

barman/retention_policies.py
andremagui: You are correct!

barman/infofile.py
```diff
@@ -878,7 +896,7 @@ def get_child_backup_info(self, child_backup_id):
 
         return None
 
-    def walk_backups_tree(self):
+    def walk_backups_tree(self, return_root=True):
```
barthisrael, line 899: Modify the `:yields:` and add `:param bool return_root:` to the docstring.

barman/retention_policies.py
```diff
@@ -126,6 +126,29 @@ def to_json(self):
         """
         return "%s %s %s" % (self.mode, self.value, self.unit)
 
+    def propagate_retention_status_to_children(self, backup_info, report, ret_status):
+        """
+        Propagate retention status to all backups in the chain.
```
barthisrael, line 131:
```suggestion
        Propagate retention status to all backups in the tree.
```

"Tree" sounds like a more appropriate word here. What do you think?
I would use "chain" when building a "single direction" path, with no options to go "left" or "right", if you get what I mean?

barman/retention_policies.py
```diff
@@ -126,6 +126,29 @@ def to_json(self):
...
+
+        .. note::
+            This has a side-effect. It modifies or add data to *report* dict.
+
+        :param barman.infofile.BackupInfo backup_info: The object we want to
+            propagate the RETENTION STATUS from.
+        :param report dict[str, str]: The report data structure to be modified.
```
barthisrael, line 138:
```suggestion
        :param dict[str, str] report: The report data structure to be modified.
```

Type comes before the name.

barman/retention_policies.py
```diff
@@ -126,6 +126,29 @@ def to_json(self):
...
+            Each key is the ID of a backup, and its value is the retention status
+            of that backup.
+        :param ret_status str:  The status of the backup according to retention
```
barthisrael, line 141:
```suggestion
        :param str ret_status:  The status of the backup according to retention
```

Type comes before the name.

barman/retention_policies.py
```diff
@@ -126,6 +126,29 @@ def to_json(self):
...
+            policies
+        """
+        backup_id_tree = backup_info.walk_backups_tree(return_root=False)
+        for backup in backup_id_tree:
```
barthisrael, line 145:
```suggestion
        backup_tree = backup_info.walk_backups_tree(return_root=False)
        for backup in backup_tree:
```

It's not a "backup ID tree" anymore :)

barman/retention_policies.py
```diff
@@ -185,6 +214,16 @@ def _backup_report(self, source):
                 i = i + 1
             else:
                 report[bid] = BackupInfo.NONE
+
+            if backups[bid].has_children:
+                status = report[bid]
+                if status in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
+                    status = BackupInfo.VALID
```
barthisrael, line 221: Maybe we want to push this "use VALID if status in X or Y" from here to `propagate_retention_status_to_children`?

I mean, we perform this logic twice in this file, related to the same thing.

barman/retention_policies.py
```diff
@@ -170,6 +193,12 @@ def _backup_report(self, source):
         # NOTE: reverse key orders (simulate reverse chronology)
         i = 0
         for bid in sorted(backups.keys(), reverse=True):
+            if backups[bid].is_incremental:
+                _logger.debug(
+                    "Ignoring incremental backup %s from retention policy report."
```
barthisrael, line 198: I wonder if this log message will sound confusing to whoever reads the log later?
I mean, we are indeed ignoring this in the main loop, but that's because we are going to handle this through other means.

I wonder if we have better ways to express that, or if we should leave it as is -- or remove it. Thoughts?

barman/retention_policies.py
```diff
@@ -126,6 +126,29 @@ def to_json(self):
         """
         return "%s %s %s" % (self.mode, self.value, self.unit)
 
+    def propagate_retention_status_to_children(self, backup_info, report, ret_status):
```
barthisrael, line 129:
```suggestion
    def _propagate_retention_status_to_children(self, backup_info, report, ret_status):
```

I think this should be an "internal" method as at a first glance we won't be calling this from the outside.

tests/test_backup.py
```diff
@@ -743,6 +743,80 @@ def test_cron_retention_only_deletes_OBSOLETE_backups(
...
+        available_backups.update(available_children)
+        get_available_backups.return_value = available_backups
+
+        children_retention_policies = {
+            "child_0": BackupInfo.VALID,
+            "child_1": BackupInfo.VALID,
+            "child_2": BackupInfo.VALID,
+            "child_3": BackupInfo.VALID,
+            "child_4": BackupInfo.OBSOLETE,
+            "child_5": BackupInfo.VALID,
```
barthisrael, line 802: Based on the code, I was expecting `child_3` to be `None` and `child_5` to be `POTENTIALLY_OBSOLETE`.

IAC I can see you are using this for mocking so whatever value here won't reflect the actual values computed by the code. But as most of the backups are "reflecting" what the code would do, maybe we could do something similar for these, just in case?

barthisrael (commented):
I need to finish reviewing your unit tests, but I'm posting what I have so far.

barman/retention_policies.py
andremagui: Yeah, chain is linear indeed! Thanks

barman/retention_policies.py
andremagui:
```
_logger.debug(
                    "Ignoring incremental backup %s. It will have the retention"
                    " policy status propagated from %s."
                    % (backups[bid], backups[bid].parent_backup_id)
                )
```
OR 

```
_logger.debug(
                    "Ignoring incremental backup %s. The retention status will"
                    " be propagated from %s if parent status is VALID,"
                    " OBSOLETE, POTENTIALLY OBSOLETE. If parent has any KEEP*"
                    " status, incremental backup will be VALID."
                    % (backups[bid], backups[bid].parent_backup_id)
                )
```
@barthisrael @gcalacoci

barthisrael: I think the first option should be ok for now.
If we think this is flooding the logs, we may choose to remove this message too.
I'm interested in @gcalacoci 's thoughts later as well.

barthisrael:
Have you by chance applied changes locally and forgot to commit?
I can see several conversations marked as resolved, but no actual changes to the code.

Besides that, I'd like to suggest a couple of things:

1. If we send a suggestion through the PR, you can use the `Add suggestion to batch` button of each of them followed by `Commit suggestions` button to apply them all, and `git pull` to fetch the commit created by GitHub. That avoids having to write the same code again on your local repo and tends to be faster for you.
2. Avoid forced pushes in between the reviews. When you force push, that makes it harder for us to re-review the code. Let's only use force pushes when rebasing the branch on top of another branch, or at the end of the PR, right before merging it.

barman/infofile.py
```diff
@@ -878,16 +896,19 @@ def get_child_backup_info(self, child_backup_id):
...
         Walk through all the children backups of the current backup.
 
         .. note::
-            The objects are returned with a bottom-up approach, including all children
-            backups plus the caller backup.
+            The objects are returned with a bottom-up approach, including all
+            children backups plus the caller backup.
+
+        :param bool return_self: Whether to return the current backup.
+            Default to True.
```
barthisrael, line 908:
```suggestion
            Default to ``True``.
```

barman/infofile.py
```diff
@@ -897,6 +918,8 @@ def walk_backups_tree(self):
                     backup_id=child_backup_id,
                 )
                 yield from backup_info.walk_backups_tree()
+        if return_self is False and self.parent_backup_id is None:
```
barthisrael, line 921:
```suggestion
        if not return_self:
```

This should be enough I guess?

This way we also allow bringing all the children of an intermediate child in the backup tree without returning the own child itself.

barman/retention_policies.py
```diff
@@ -126,6 +126,31 @@ def to_json(self):
...
+
+        :param barman.infofile.BackupInfo backup_info: The object we want to
+            propagate the RETENTION STATUS from.
+        :param dict[str, str] report: The report data structure to be modified.
+            Each key is the ID of a backup, and its value is the retention status
+            of that backup.
+        :param str ret_status:  The status of the backup according to retention
+            policies
+        """
+        if ret_status in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
```
barthisrael, line 144:
```suggestion
        # As KEEP status doesn't make sense for incremental backups, we simply
        # set them as VALID if their root full backup has KEEP annotation
        if ret_status in (BackupInfo.KEEP_STANDALONE, BackupInfo.KEEP_FULL):
```

Maybe it's worth it a comment.

barman/retention_policies.py
```diff
@@ -170,6 +195,15 @@ def _backup_report(self, source):
...
         i = 0
         for bid in sorted(backups.keys(), reverse=True):
+            if backups[bid].is_incremental:
+                _logger.debug(
+                    "Ignoring incremental backup %s. The retention status will"
+                    " be propagated from %s if parent status is VALID,"
+                    " OBSOLETE, POTENTIALLY OBSOLETE. If parent has any KEEP*"
+                    " status, incremental backup will be VALID."
+                    % (backups[bid], backups[bid].parent_backup_id)
+                )
```
barthisrael, line 205:
```suggestion
                _logger.debug(
                    "Ignoring incremental backup %s. The retention status will"
                    " be propagated from %s."
                    % (backups[bid], backups[bid].parent_backup_id)
                )
```

Let's use the same log message in both cases (recovery window and redundancy checks)

barthisrael (changes requested):
Some comments about the code.

barthisrael:
The implementation itself seems to be working fine, at least with a redundancy retention policy.

For example, with 4 full backups, one of them with `KEEP`, and retention`redundancy 2`:

```console
$ barman list-backup pg17
pg17 20240702T224908 'yet_another_incr_backup' - Tue Jul  2 20:49:10 2024 - Size: 35.3 MiB - WAL Size: 0 B
pg17 20240702T224830 'yet_another_full_backup' - Tue Jul  2 20:48:33 2024 - Size: 251.2 MiB - WAL Size: 32.0 MiB
pg17 20240702T224631 'another_incr_backup_1' - Tue Jul  2 20:46:32 2024 - Size: 35.3 MiB - WAL Size: 48.0 MiB
pg17 20240702T224353 'another_full_backup' - Tue Jul  2 20:43:55 2024 - Size: 251.2 MiB - WAL Size: 32.0 MiB
pg17 20240702T224345 'incr_backup_2' - Tue Jul  2 20:43:47 2024 - Size: 35.3 MiB - WAL Size: 32.0 MiB - OBSOLETE
pg17 20240702T224339 'incr_backup_1' - Tue Jul  2 20:43:41 2024 - Size: 35.3 MiB - WAL Size: 32.0 MiB - OBSOLETE
pg17 20240702T224322 'full_backup' - Tue Jul  2 20:43:24 2024 - Size: 251.2 MiB - WAL Size: 32.0 MiB - OBSOLETE
pg17 20240702T223430 'child2.1' - Tue Jul  2 20:34:32 2024 - Size: 35.3 MiB - WAL Size: 48.0 MiB
pg17 20240702T223422 'child2' - Tue Jul  2 20:34:23 2024 - Size: 35.3 MiB - WAL Size: 32.0 MiB
pg17 20240702T223414 'child1.1.2' - Tue Jul  2 20:34:15 2024 - Size: 35.3 MiB - WAL Size: 32.0 MiB
pg17 20240702T223408 'child1.1.1' - Tue Jul  2 20:34:10 2024 - Size: 35.3 MiB - WAL Size: 32.0 MiB
pg17 20240702T223359 'child1.1' - Tue Jul  2 20:34:00 2024 - Size: 35.3 MiB - WAL Size: 48.0 MiB
pg17 20240702T223325 'child1' - Tue Jul  2 20:33:26 2024 - Size: 35.3 MiB - WAL Size: 32.0 MiB
pg17 20240702T223314 'root' - Tue Jul  2 20:33:16 2024 - Size: 251.2 MiB - WAL Size: 32.0 MiB - KEEP:STANDALONE
```

When `barman cron` runs, it deletes the `OBSOLETE` backups:

```console
$ barman cron
Starting WAL archiving for server pg17
Enforcing retention policy: removing backup 20240702T224322 for server pg17
Deleting backup 20240702T224322 for server pg17
Deleted backup 20240702T224322 (start time: Tue Jul  2 23:02:19 2024, elapsed time: less than one second)
Enforcing retention policy: removing backup 20240702T224339 for server pg17
Deleting backup 20240702T224339 for server pg17
Deleted backup 20240702T224339 (start time: Tue Jul  2 23:02:19 2024, elapsed time: less than one second)
Enforcing retention policy: removing backup 20240702T224345 for server pg17
Deleting backup 20240702T224345 for server pg17
Deleted backup 20240702T224345 (start time: Tue Jul  2 23:02:19 2024, elapsed time: less than one second)
```

Tomorrow I'll take a look at unit tests again.
I'm fine about the code changes.

tests/test_barman_cloud_backup_delete.py
```diff
@@ -103,24 +103,73 @@ def _get_mock_backup_files(self, file_paths):
             for name, path in file_paths.items()
         )
 
-    def _create_backup_metadata(
-        self, backup_ids, begin_wals=None, end_wals=None, is_snapshot_backup=False
+    def _create_backup_info(
+        self,
+        backup_id,
+        begin_wals=None,
+        end_wals=None,
```
barthisrael, line 110:
```suggestion
        begin_wal=None,
        end_wal=None,
```

It seems a bit weird to pass the whole dict of `begin_wals` and `end_wals` to this method, if it's supposed to handle only a single backup info.

I'd say to pass only the value for that backup info, if that's the case.

barthisrael: The docstring needs to be adjusted accordingly.

tests/test_barman_cloud_backup_delete.py
```diff
@@ -103,24 +103,73 @@ def _get_mock_backup_files(self, file_paths):
...
         This is used by tests for two purposes:
-          1. Creating a mock CloudBackupCatalog.
-          2. Providing the data needed to verify that the expected backups were deleted.
+          1. Helper to create a backup_metadata and mock CloudBackupCatalog.
+          2. Providing the data needed to create backup_metadata.
         """
         if begin_wals is None:
             begin_wals = {}
         if end_wals is None:
             end_wals = {}
```
barthisrael, line 130:
```suggestion
```

We would no longer deal with this as a dict.

tests/test_barman_cloud_backup_delete.py
```diff
@@ -103,24 +103,73 @@ def _get_mock_backup_files(self, file_paths):
...
+        ) + datetime.timedelta(hours=1)
+        try:
+            backup_info.begin_wal = begin_wals[backup_id]
+            backup_info.timeline = int(backup_info.begin_wal[:8])
+        except KeyError:
+            pass
+        try:
+            backup_info.end_wal = end_wals[backup_id]
+        except KeyError:
+            pass
```
barthisrael, line 147:
```suggestion
        if begin_wal:
            backup_info.begin_wal = begin_wal
            backup_info.timeline = int(backup_info.begin_wal[:8])

        if end_wal:
            backup_info.end_wal = end_wal
```

We would no longer deal with this as a dict.

tests/test_backup.py
```diff
@@ -743,6 +743,80 @@ def test_cron_retention_only_deletes_OBSOLETE_backups(
...
+        delete_backup.call_count == 2
+        expected_calls = [
+            call(available_backups["child_4"], skip_wal_cleanup_if_standalone=False),
+            call(
+                available_backups["obsolete_backup"],
+                skip_wal_cleanup_if_standalone=False,
+            ),
+        ]
+        delete_backup.assert_has_calls(expected_calls, any_order=False)
+
```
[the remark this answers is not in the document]
andremagui: I added this because cron_retetion_policy() calls the report() method. So i wanted to be sure that the child was deleted with its parent calling the delete twice. But now RE-reading the name of the test, i am thinking over what you pointed out. Maybe we could add the children parts to the existing test?

tests/test_barman_cloud_backup_delete.py
```diff
@@ -141,23 +190,9 @@ def _create_backup_metadata(
...
-                backup_info.begin_wal = begin_wals[backup_id]
-                backup_info.timeline = int(backup_info.begin_wal[:8])
-            except KeyError:
-                pass
-            try:
-                backup_info.end_wal = end_wals[backup_id]
-            except KeyError:
-                pass
+            backup_info = self._create_backup_info(
+                backup_id, begin_wals, end_wals, backup_name
```
barthisrael, line 194:
```suggestion
                backup_id, begin_wals.get(backup_id), end_wals.get(backup_id), backup_name
```

That logic which replaces `begin_wals` and `end_wals` that were `None` with an empty dict (`{}`) would need to be performed in the beginning of this function.

barthisrael (changes requested):
Suggestions for improving the refactoring applied on `_create_backup_metadata`.

tests/test_backup.py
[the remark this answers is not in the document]
barthisrael: What I mean is: `cron_retention_policy` doesn't know which type of backup it is dealing with:

* Is it rsync?
* Is it Postgres 16 or older?
* Is it Postgres 17 full?
* Is it Postgres 17 incremental?

That information doesn't matter for that method. The only thing that matters is:

* Is the backup `OBSOLETE`?

The logic which you mentioned belongs to `_backup_report`. That's the method we should check if it is performing as expected.

I mean, make sure `_backup_report` is marking the children nodes in the tree with the expected status.

In this case, the test on `cron_retention_policy` seems redundante because we are artificially mocking everything, including the return of `report` method.

The idea is that we get to something like this:

* When testing `_backup_report`: assuming I have this set of full and incremental backups, I would expect backups X, Y and Z to be marked as status "SOME STATUS";
* When testing `cron_retention_policy`: assuming `report` gives me this set of backups with their status, I remove backups X, Y and Z.

Does that make it clearer what I have in mind?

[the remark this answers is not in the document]
barthisrael: > When testing cron_retention_policy: assuming report gives me this set of backups with their status, I remove backups X, Y and Z.

FWIW this is already covered by tests like `test_cron_retention_only_deletes_OBSOLETE_backups`.

tests/test_retention_policies.py
```diff
@@ -180,49 +181,74 @@ def test_backup_status(self, server):
...
-        backup_info = build_test_backup_info(
-            server=server, backup_id="test1", end_time=datetime.now(tzlocal())
-        )
+        # execute retention policy report on parent
+        report_parent = rp.backup_status("test_backup")
 
-        # instruct the get_available_backups method to return a map with
-        # our mock as result and minimum_redundancy = 1
-        server.get_available_backups.return_value = {"test_backup": backup_info}
-        server.config.minimum_redundancy = 1
```
barthisrael, line 214: I can see you refactored this test a bit to avoid duplication of commands when setting up the mocks.

Although I think this like `server.config.minimum_redundancy = 1` should not be removed as part of that refactoring.

barthisrael (changes requested):
Review of `test_backup_status`

tests/test_barman_cloud_backup_delete.py
```diff
@@ -103,12 +103,57 @@ def _get_mock_backup_files(self, file_paths):
...
+        parent_backup_id=None,
+        children_backup_ids=None,
+        is_incremental=False,
+        has_children=False,
+    ):
+        """
+        Helper for tests which creates mock BackupInfo objects.
+
+        If begin_wal has an entry for a backup_id then the begin_wal and
+        timeline values will be set according to the wal name.
```
barthisrael, line 121:
```suggestion
        If *begin_wal* is set then the backup's begin_wal and
        timeline values will be set according to the wal name.
```

It is no longer a dict heh

barthisrael (changes requested):
Last comment about `_create_backup_info`.

tests/test_backup.py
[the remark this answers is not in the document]
gcalacoci: Yeah seems not necessary to me

tests/test_retention_policies.py
barthisrael: Never mind, I can see this is also set above, this was a duplication in the original code.

tests/test_retention_policies.py
```diff
@@ -232,12 +258,27 @@ def test_first_backup(self, server):
...
+            parent_backup_id="test_backup",
+            children_backup_ids=None,
+        )
         # instruct the get_available_backups method to return a map with
         # our mock as result and minimum_redundancy = 1
-        server.get_available_backups.return_value = {"test_backup": backup_info}
+        server.get_available_backups.return_value = {
+            "test_backup": backup_info,
+            "test_backup_child": child_backup_info,
+        }
```
barthisrael, line 281: This doesn't seem to add value to the unit test from what I can see.

I mean, `test_first_backup` simply checks if the `_first_backup` attribute will contain the first `VALID` backup in the catalog. With that in mind, the incremental which you add to the list won't even be taken into account.

barthisrael (changes requested):
Review of `test_first_backup` unit test.

I think we should undo the changes here, see my comment.

tests/test_retention_policies.py
gcalacoci: yup. I agree with this.
the above modifications do not really add something useful to the test (that... is worryingly missing any docstring and difficult to understand) I think we should not apply this code.
@barthisrael @andremagui

tests/test_retention_policies.py
```diff
@@ -249,21 +290,259 @@ def test_first_backup(self, server):
...
         server.config.minimum_redundancy = 1
 
         # execute retention policy report
         report = rp.first_backup()
 
         assert report == "test_backup"
 
+    @pytest.mark.parametrize(
+        "retention_policy", ("RECOVERY WINDOW OF 4 WEEKS", "REDUNDANCY 2")
+    )
```
barthisrael, line 304:
```suggestion
    @pytest.mark.parametrize(
        ("retention_policy", "retention_status"),
        itertools.product(
            ("RECOVERY WINDOW OF 4 WEEKS", "REDUNDANCY 2"),
            (
                BackupInfo.OBSOLETE,
                BackupInfo.VALID,
                BackupInfo.POTENTIALLY_OBSOLETE,
                BackupInfo.KEEP_FULL,
                BackupInfo.KEEP_STANDALONE,
                BackupInfo.NONE,
            ),
        ),
    )
```

Instead of using a fixed `RETENSION_STATUS`, let's play with actual valid values for `_propagate_retention_status_to_children` method and check how the method behaves when it receives them.

Here I'm only providing how to parametrize the test. After you apply this suggestion you will need to change the unit test itself to use this new parameter.

barthisrael (changes requested):
Enhancements to unit test `test__propagate_retention_status_to_children`

barthisrael (changes requested):
Review of unit tests `test_redundancy_report_with_incrementals` and `test_recovery_window_report_with_incrementals`.

tests/test_retention_policies.py
andremagui: @gcalacoci @barthisrael  the goal here was just to be more precise in the test. I just think that if you have a get_available_backups.return_value with only ONE backup_info, this will always be the first backup

tests/test_retention_policies.py
```diff
@@ -249,21 +290,259 @@ def test_first_backup(self, server):
...
+        backups_data = {
+            "20240628T000000": {
+                "parent_backup_id": None,
+                "children_backup_ids": ["20240628T120000"],
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=6),
+            },
+            "20240628T120000": {
+                "parent_backup_id": "20240628T000000",
+                "children_backup_ids": None,
+                "end_time": datetime.now(tzlocal()) - timedelta(weeks=6, days=1),
```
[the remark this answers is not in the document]
andremagui: Sorry, it was suposed to be a +

barthisrael (changes requested):
The changes to the source code and to the unit tests look good to me.

My last round of suggestions related to the commits and to the PR themselves.

We have the following commits:

1. 081c6ac3803867e0cd9e3db63c401ee21c966ae9: modifies the source code to implement the changes of this PR;
2. 017bd474195cdced29413bdda019924267cbc002: adds unit tests to cover the new code;
3. 366346a0e8424c22316364bfb1b20024543a04db: fixes unit tests that became broken because of the new code;
4. 722189b744014ef735ee59c2dde5a83764f9707e: provides enhancements to `test_first_backup` unit test, which is a side change.

I propose the following changes:

1. Have 366346a0e8424c22316364bfb1b20024543a04db as second commit and 017bd474195cdced29413bdda019924267cbc002 as third commit. Logically speaking, it seems to make more sense if we "first fix broken things", then "cover new code";
2. 081c6ac3803867e0cd9e3db63c401ee21c966ae9, 017bd474195cdced29413bdda019924267cbc002 and 366346a0e8424c22316364bfb1b20024543a04db all have the same commit body (at least the beginning of them all). I think for the first one it sounds ok, but for the second and third commit the commit body should be something else, reflecting what the commit is actually changing. For example, the commit title of one of them says `Add and modify unit tests accordingly to implementations`, but then the commit body seems not connected with that, if you get what I mean? I'd say to use something like `This commit fixes unit tests which got broken because of changes introduced to retention policies to handle incremental backups. The previous unit tests got broken because of X, Y and Z, and we did A, B and C to fix them`;
3. In the first commit (081c6ac3803867e0cd9e3db63c401ee21c966ae9) we have the following mentions:

> This commit:
> 1. Add new method to RetentionPolicy class
> 	1.1 propagate_retention_status_to_children
> 2. Modify method _backup_report of RecoveryWindowRetentionPolicy
> 	and RedundancyRetentionPolicy
> 3. Add properties to infofile.py LocalBackupInfo
> 	3.1 is_incremental
> 	3.2 has_children
> 4. Add new parameter to walk_backups_tree in infofile.py
> 	so we can control whether to yield the parent also

That is on the right path, but incomplete. I'm missing a more detailed explanation for each change that is outlined. For example: "propagate_retention_status_to_children" was added -- why? what is its job? Similar thoughts for other items.

(This last change can be reflected in the PR body later, too)

barthisrael (approved):
Looks great :)

gcalacoci (approved):
Definitely better than previous iterations, leaner and cleaner both in terms of working code and testing code.
LGTM

PR #952, opened 2024-06-25
When recovering with `--no-get-wal` and `--target-lsn`, bring only the WALs up to the specified LSN

barthisrael:
Previous to this PR, when the user specified `--target-lsn` and `--no-get-wal` in `barman recover` command, Barman was copying all WAL files from Barman to Postgres.

As we are able to infere the required WAL files based on the specified `--target-lsn`, this PR changes Barman so it copies only the required WAL files instead of them all.

Notes:

* We also added a missing unit test to cover the case where `--target-xid` is specified;
* This PR is based on #951. That PR needs to be merged before this PR.

References: BAR-190.

### DIFF
--- a/barman/recovery_executor.py
+++ b/barman/recovery_executor.py
@@ -255,6 +255,9 @@ class RecoveryExecutor(object):
                     self.server.get_required_xlog_files(
                         backup_info,
                         target_tli,
+                        None,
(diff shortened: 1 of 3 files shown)


gcalacoci (approved):
LGTM

PR #955, opened 2024-07-01
create new recovery executor that uses `pg_combinebackup`

andremagui:
pg_combinebackup utility is coming to Postgresql 17 with the introduction of incremental backup to reconstruct a synthetic full backup from a sequence of incremental backups and the previous full backup.  

This PR:
- Create a new command wrapper for the `pg_combinebackup` utility.
- Create a new recovery executor, which does the following:
   1. Get the backup chain and combine them into a synthetic backup in a common staging path
   2. Reuse the recover method from the `RecoveryExecutor` super class, by passing the newly created synthetic backup;
   3. In the copy phase, we check if is remote or local recovery:
      a. if it is remote, we reuse the rsync logic already present in the parent recovery executor;
      b. if it is local, we move the data directory and tablespaces from staging to the destination.

Note: we are reusing the `--recovery-staging-path` config options to point the place where `pg_combinebackup` should create the synthetic backup in the Barman host. Originally that config option was used only to point the path in the Postgres host where a backup should be decompressed.

References: BAR-167 & BAR-168

### DIFF
--- a/barman/cli.py
+++ b/barman/cli.py
@@ -909,9 +909,9 @@ def recover(args):
         )
         output.close_and_exit()
 
-    # If the backup to be recovered is compressed then there are additional
-    # checks to be carried out
-    if backup_id.compression is not None:
+    # If the backup to be recovered is compressed or an incremental then
+    # there are additional checks to be carried out
+    if backup_id.compression is not None or backup_id.parent_backup_id is not None:
         # Set the recovery staging path from the cli if it is set
         if args.recovery_staging_path is not None:
             try:
@@ -922,21 +922,34 @@ def recover(args):
                 output.error("Cannot parse recovery staging path: %s", str(exc))
                 output.close_and_exit()
             server.config.recovery_staging_path = recovery_staging_path
-        # If the backup is compressed but there is no recovery_staging_path
-        # then this is an error - the user *must* tell barman where recovery
-        # data can be staged.
+
+        # If no staging path is found then we need to display a proper error
+        # according to the backup type (compressed or incremental)
         if server.config.recovery_staging_path is None:
-            output.error(
-                "Cannot recover from backup '%s' of server '%s': "
-                "backup is compressed with %s compression but no recovery "
-                "staging path is provided. Either set recovery_staging_path "
-                "in the Barman config or use the --recovery-staging-path "
-                "argument.",
-                args.backup_id,
-                server.config.name,
-                backup_id.compression,
-            )
-            output.close_and_exit()
+            if backup_id.parent_backup_id is not None:
+                output.error(
+                    "Cannot recover from backup '%s' of server '%s': "
+                    "backup will be combined with pg_combinebackup in the "
+                    "barman host but no recovery staging path is provided. "
+                    "Either set recovery_staging_path in the Barman config "
+                    "or use the --recovery-staging-path argument.",
+                    args.backup_id,
+                    server.config.name,
+                )
+                output.close_and_exit()
+
+            if backup_id.compression is not None:
+                output.error(
+                    "Cannot recover from backup '%s' of server '%s': "
+                    "backup is compressed with %s compression but no recovery "
+                    "staging path is provided. Either set recovery_staging_path "
+                    "in the Barman config or use the --recovery-staging-path "
+                    "argument.",
+                    args.backup_id,
+                    server.config.name,
+                    backup_id.compression,
+                )
+                output.close_and_exit()
 
     # decode the tablespace relocation rules
     tablespaces = {}
--- a/barman/command_wrappers.py
+++ b/barman/command_wrappers.py
@@ -1150,6 +1150,64 @@ class PgVerifyBackup(PostgreSQLClient):
             self.args += args
 
 
+class PgCombineBackup(PostgreSQLClient):
+    """
+    Wrapper class for the ``pg_combinebackup`` system command
+    """
+
+    COMMAND_ALTERNATIVES = ["pg_combinebackup"]
+
+    def __init__(
+        self,
+        destination,
+        command,
+        tbs_mapping=None,
+        connection=None,
+        version=None,
+        app_name=None,
+        check=True,
+        args=None,
+        **kwargs
+    ):
+        """
+        Constructor
+
+        :param str destination: destination directory path
+        :param str command: the command to use
+        :param None|Dict[str, str] tbs_mapping: used for tablespace
+        :param PostgreSQL connection: an object representing
+          a database connection
+        :param Version version: the command version
+        :param str app_name: the application name to use for the connection
+        :param bool check: check if the return value is in the list of
+          allowed values of the :class:`Command` obj
+        :param None|List[str] args: additional arguments
+        """
+        PostgreSQLClient.__init__(
+            self,
+            connection=connection,
+            command=command,
+            version=version,
+            app_name=app_name,
+            check=check,
+            **kwargs
+        )
+
+        # Set the backup destination
+        self.args = ["--output=%s" % destination]
+
+        # The tablespace mapping option is repeated once for each tablespace
+        if tbs_mapping:
+            for tbs_source, tbs_destination in tbs_mapping.items():
+                self.args.append(
+                    "--tablespace-mapping=%s=%s" % (tbs_source, tbs_destination)
+                )
+
+        # Manage additional args
+        if args:
+            self.args += args
+
+
 class BarmanSubProcess(object):
     """
     Wrapper class for barman sub instances
--- a/barman/infofile.py
+++ b/barman/infofile.py
@@ -973,7 +973,7 @@ class SyntheticBackupInfo(LocalBackupInfo):
     ):
         """
         Stores meta information about a single synthetic backup.
-        
+
         .. note::
             A synthetic backup is a base backup which was artificially created
             through ``pg_combinebackup``. A synthetic backup is not part of
--- a/barman/infofile.py
+++ b/barman/infofile.py
@@ -922,13 +922,22 @@ class LocalBackupInfo(BackupInfo):
             return
         yield self
 
-    def walk_to_root(self):
+    def walk_to_root(self, return_self=True):
         """
         Walk through all the parent backups of the current backup.
+
+        .. note::
+            The objects are returned with a bottom-up approach, including all
+            parents backups plus the caller backup if *return_self* is ``True``.
+
+        :param bool return_self: Whether to return the current backup.
+            Default to ``True``.
 
         :yield: a generator of :class:`LocalBackupInfo` objects for each parent backup.
         """
 
+        if return_self:
+            yield self
         backup_info = self.get_parent_backup_info()
         while backup_info:
             yield backup_info
@@ -956,3 +965,41 @@ class LocalBackupInfo(BackupInfo):
         ):
             return True
         return False
+
+
+class SyntheticBackupInfo(LocalBackupInfo):
+    def __init__(
+        self, server, base_directory, backup_id=None, info_file=None, **kwargs
+    ):
+        """
+        Stores meta information about a single synthetic backup.
+        
+        .. note::
+            A synthetic backup is a base backup which was artificially created
+            through ``pg_combinebackup``. A synthetic backup is not part of
+            the Barman backup catalog, and only exists so we are able to
+            recover a backup created by ``pg_combinebackup`` utility, as
+            almost all functions and methods require a backup info object.
+
+        The only difference from this class to its parent :class:`LocalBackupInfo`
+        is that it accepts a custom base directory for the backup as synthetic
+        backups are expected to live on directories other than the default
+        ``<server_name>/base`` path.
+
+        :param barman.server.Server server: the server that owns the
+            synthetic backup
+        :param str base_directory: the root directory where this synthetic
+            backup resides, essentially an override to the
+            ``server.config.basebackups_directory`` configuration.
+        :param str|None backup_id: the backup id of this backup
+        :param None|str|TextIO info_file: path or file descriptor of an existing
+            synthetic ``backup.info`` file
+        """
+        self.base_directory = base_directory
+        super(SyntheticBackupInfo, self).__init__(
+            server, info_file, backup_id, **kwargs
+        )
+
+    def get_basebackup_directory(self):
+        """Get the backup directory based on its base directory"""
+        return os.path.join(self.base_directory, self.backup_id)
--- a/barman/recovery_executor.py
+++ b/barman/recovery_executor.py
@@ -24,6 +24,8 @@ from __future__ import print_function
 
 import collections
 import datetime
+
+from functools import partial
 import logging
 import os
 import re
@@ -37,7 +39,7 @@ import dateutil.tz
 
 from barman import output, xlog
 from barman.cloud_providers import get_snapshot_interface_from_backup_info
-from barman.command_wrappers import RsyncPgData
+from barman.command_wrappers import PgCombineBackup, RsyncPgData
 from barman.config import RecoveryOptions
 from barman.copy_controller import RsyncCopyController
 from barman.exceptions import (
@@ -58,8 +60,8 @@ from barman.compression import (
     NoneCompression,
 )
 import barman.fs as fs
-from barman.infofile import BackupInfo, LocalBackupInfo
-from barman.utils import force_str, mkpath
+from barman.infofile import BackupInfo, LocalBackupInfo, SyntheticBackupInfo
+from barman.utils import force_str, mkpath, total_seconds
 
 # generic logger for this module
 _logger = logging.getLogger(__name__)
@@ -1417,6 +1419,7 @@ class TarballRecoveryExecutor(RemoteConfigRecoveryExecutor):
         output.info(
             "Staging compressed backup files on the recovery host in: %s", staging_dir
         )
+
         recovery_info["cmd"].create_dir_if_not_exists(staging_dir, mode="700")
         recovery_info["cmd"].validate_file_mode(staging_dir, mode="700")
         recovery_info["staging_dir"] = staging_dir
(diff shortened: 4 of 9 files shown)


barman/recovery_executor.py
```diff
@@ -1807,13 +1807,94 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        starting from the full backup, through the parent backups up until the provided
+        incremental backup.
+
+        :param backup_info: The backup_info object for the incremental backup.
+        :return: A list of backup_info objects from the root to the provided incremental backup.
+        """
+        # Convert the generator to a list and reverse it to start from the root backup
+        return [
+            backup
+            for backup in reversed(list(backup_info.walk_to_root(include_self=True)))
```
andremagui, line 1851: check [dev/bar-168](https://github.com/EnterpriseDB/barman/pull/950/files#diff-458020cb725a00dcf8f7c7a3902a83a1aa05ff62d48dea5e1e1442cf3ad9e40bR141 ). Maybe there is a way that you do not need to turn the generator into the list. Maybe find a way to run a for loop and insert each value in the beginning of a list instead of creating the list first and then reversing it.

barman/backup.py
```diff
@@ -1628,3 +1634,62 @@ def verify_backup(self, backup_info):
...
+        :param str synthetic_backup_dir: the synthetic backup directory
+        """
+        synthetic_backup_info = SynthethicBackupInfo(
+            server=self.server,
+            backup_dir=synthetic_backup_dir,
+        )
+
+        # set server parameters
+        strategy = PostgresCombineBackupStrategy(self.server.postgres, self.config.name)
+        strategy.set_pg_metadata(synthetic_backup_info)
```
andremagui, line 1666: IMHO i think we should not use the _pg_get_metadata from the strategy because it ends up accessing the postgres server for information. What if we are doing a recover and for any reason the postgres server is down or the configuration was changed. We will end up with weird things in the backup.info. I think we should reuse as many configs as possible from existing backups. Does it make sense?

barman/backup.py
```diff
@@ -1628,3 +1634,62 @@ def verify_backup(self, backup_info):
...
+
+        # set wal-related parameters from the last incremetal backup
+        wal_fields = (
+            "begin_wal",
+            "begin_xlog",
+            "begin_offset",
+            "end_wal",
```
andremagui, line 1673: I would say from the test i did that begin_wal, begin_xlog its correct to take from the last incremental! Would this be correct for the others as well?

barman/backup.py
```diff
@@ -1628,3 +1634,62 @@ def verify_backup(self, backup_info):
...
+            "end_offset",
+        )
+        for field in wal_fields:
+            last_backup_val = getattr(backup_info, field, None)
+            synthetic_backup_info.set_attribute(field, last_backup_val)
+
+        # calculate the backup size
+        backup_size = 0
+        backup_dest = synthetic_backup_info.get_basebackup_directory()
+        for dir_path, _, file_names in os.walk(backup_dest):
```
andremagui, line 1684: Maybe this can be done using the walk_to_root method and summing the size field from backup.info?! This is looping over the dir and opening each file to get the stat. Dunno, just a thought!

barman/backup.py
gustabowill: It makes sense! Yeah we really shouldn't rely on any connection to the server to get this done.

barman/backup.py
gustabowill: Hmmmm I'm not sure about it, but I don't think the `pg_combinebackup` works as simple as putting all data together, I wouldn't assume the synthetic backup size will always be exactly the sum of all backups in the chain 🤔

barman/infofile.py
```diff
@@ -955,8 +955,11 @@ def __init__(self, server, backup_dir, info_file=None, **kwargs):
         # the backup_id is then the directory name where the backup resides
         backup_id = os.path.basename(os.path.normpath(backup_dir))
 
-        super(SynthethicBackupInfo, self).__init__(
-            server=server, info_file=info_file, backup_id=backup_id, **kwargs
+        super(SyntheticBackupInfo, self).__init__(
```
andremagui, line 958: I think i did not understand this part. 
we are going to have a command like `barman recover [OPTIONS] SERVER_NAME BACKUP_ID DESTINATION_DIRECTORY`. Why use backup_dir here? The DESTINATION_DIRECTORY is something like path/to/backup and backup_id is already passed here. Can't we just use those directly?

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,321 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+            backup_id=backup_info.backup_id,
+        )
+        synthetic_backup_info.load(filename=backup_info.filename)
+
+        dest_dirs = [synthetic_backup_info.get_data_directory()]
+
+        # Maps the tablespaces from the old backup directory to the new synthetic
+        # backup directory. This mapping is passed to the pg_combinebackup as input
+        tbs_map = {}
+        if backup_info.tablespaces:
```
andremagui, line 1892: using the backup_info instead of synthetic is better for testing and also i think it does not matter which to use here since synthetic its almost a copy of backup_info. It could be safer to use backup_info in case of any erros when creating the synthetic.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,321 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        :param barman.infofile.LocalBackupInfo backup_info: an incremental
+            backup in the chain
+        :param str dest: the directory where the synthetic backup is going
+            to be mounted on
+        :return barman.infofile.SyntheticBackupInfo: the backup info file of the
+            combined backup
+        """
+        self.combine_start_time = datetime.datetime.now()
+
+        dest = parse_recovery_staging_path(dest)
```
andremagui, line 1874: this is checking for absolute path!

gustabowill: But I don't think we need here again. This check is done when the user changes the server paramater config and also when he passes it via CLI, so there's basically no way we could have an relative path here, that's why a removed it earilier

andremagui: Oh i did not check the cli! Ok, no problem!

barman/command_wrappers.py
```diff
@@ -1150,6 +1150,71 @@ def __init__(
             self.args += args
 
 
+class PgCombineBackup(PostgreSQLClient):
+    """
+    Wrapper class for the pg_combinebackup system command
```
barthisrael, line 1155:
```suggestion
    Wrapper class for the ``pg_combinebackup`` system command
```

barman/command_wrappers.py
```diff
@@ -1150,6 +1150,71 @@ def __init__(
...
+        :param Dict[str, str] tbs_mapping: used for tablespace
+        :param bool immediate: fast checkpoint identifier for pg_basebackup
+        :param bool check: check if the return value is in the list of
+          allowed values of the Command obj
+        :param barman.compression.PgBaseBackupCompression compression:
+          the pg_basebackup compression options used for this backup
+        :param str parent_backup_manifest_path:
+          the path to a backup_manifest file from a previous backup which can
+          be used to perform an incremental backup
+        :param List[str] args: additional arguments
```
barthisrael, line 1191:
```suggestion
        :param str destination: destination directory path
        :param str command: the command to use
        :param None|Dict[str, str] tbs_mapping: used for tablespace
        :param PostgreSQL connection: an object representing
          a database connection
        :param Version version: the command version
        :param str app_name: the application name to use for the connection
        :param bool check: check if the return value is in the list of
          allowed values of the :class:`Command` obj
        :param None|List[str] args: additional arguments
```

The docstring had arguments in the wrong worder, or arguments from other commands (probably result of a copy'n' paste).

This is a suggestion to fix it.

barthisrael (changes requested):
Review of 81e35307cf067dcee0298fa5f10a9b3027e0d7ec.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        # Get the backup chain data paths to be passed to the pg_combinebackup
+        backups_chain = self._get_backup_chain_path(backup_info)
+
+        self._start_message(synthetic_backup_info)
+
+        pg_combinebackup = PgCombineBackup(
+            destination=synthetic_backup_info.get_data_directory(),
+            command=remote_status["pg_combinebackup_path"],
+            version=remote_status["pg_combinebackup_version"],
+            app_name=self.config.streaming_backup_name,
```
barthisrael, line 1916:
```suggestion
            app_name=None,
```

I would avoid reusing that application name as I guess that can sound confusing to the user if they inspect `pg_stat_replication`. I suggested `None` because that's what `PgVerifyBackup` class does.

If we want to have `application_name` in the connection string, I would say to use something else other than the streaming backup name.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+
+    def _move_to_destination(self, source, destination, exclude_dirs=[]):
+        """
+        Move a single directory or file to its destination directory
+
+        :param barman.infofile.SyntheticBackupInfo backup_info: the synthetic
+            backup info file
+        :param barman.infofile.Tablespace: the tablespace to be moved
+        :param dict[str,str]|None tablespaces_mapping: a tablespace
+            name -> location map (for relocation)
```
barthisrael, line 2001: I'd say this docstring needs a complete rewrite.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        """
+        for file_or_dir in os.listdir(source):
+            if file_or_dir not in exclude_dirs:
```
barthisrael, line 2004: We name the argument `exclude_dirs`, but this logic can also exclude files if their name matches anything in `exclude_dirs`.

Going further, let's use a `set` instead of a `list`, so we have faster lookups.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+                file_or_dir_path = os.path.join(source, file_or_dir)
+                try:
+                    shutil.move(file_or_dir_path, destination)
+                except shutil.Error:
+                    output.error(
+                        "Destination directory '%s' must be empty." % destination
+                    )
+                    output.close_and_exit()
+
+    def _get_backup_chain_path(self, backup_info):
```
barthisrael, line 2014:
```suggestion
    def _get_backup_chain_paths(self, backup_info):
```

It seems we are returning the path for each backup in the chain, so a plural sounds better to me here.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        """
+        Get the path of the backup chain, from the full backup to the specified
+        incremental backup.
```
barthisrael, line 2017:
```suggestion
        Get the path of each backup in the chain, from the full backup to
        the specified incremental backup.
```

Sounds clearer like this.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+
+        :param barman.infofile.LocalBackupInfo backup_info: The incremental backup
```
barthisrael, line 2019:
```suggestion
        :param barman.infofile.SyntheticBackupInfo backup_info: The incremental backup
```

From what I can see this will always be a `SynheticBackupInfo` instance.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        """
```
barthisrael, line 2020: Missing `:return:` in the docstring.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        return reversed(
+            [backup.get_data_directory() for backup in backup_info.walk_to_root()]
+        )
+
+    def _prepare_destination(self, dest_dir):
+        """
+        Prepare the destination of a backup data/tablespace.
```
barthisrael, line 2027:
```suggestion
        Prepare the destination of a backup base/tablespace directory.
```

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+
+        This method is responsible for removing a directory if it already
+        exists and for ensuring the correct permissions on the directory
```
barthisrael, line 2030:
```suggestion
        This method is responsible for removing a directory if it already
        exists, then (re)creating it and ensuring the correct permissions
        on the directory.
```

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        # Remove all the destination directories and reinit the backup
+        for d in dest_dirs:
+            self._prepare_destination(d)
+
+    def _fetch_remote_status(self):
+        """
+        Gather info from the remote server.
+
+        This method does not raise any exception in case of errors,
+        but set the missing values to None in the resulting dictionary.
```
barthisrael, line 2069:
```suggestion
        but set the missing values to ``None`` in the resulting dictionary.
```

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        """
```
barthisrael, line 2070: Missing `:return:`.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+
+        pg_version = None
+        if self.server.postgres is not None:
+            pg_version = self.server.postgres.server_major_version
+
+        # Is the tablespace mapping option supported?
+        if pgcombinebackup_version >= "9.4":
+            remote_status["pg_combinebackup_tbls_mapping"] = True
+        else:
+            remote_status["pg_combinebackup_tbls_mapping"] = False
```
barthisrael, line 2101: `pg_combinebackup` was introduced on Postgres 17, and there is no version without support for tablespace relocation.

I think we can get rid of `pg_combinebackup_tbls_mapping`.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,327 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+
+        # If any of the two versions is unknown, we can't compare them
+        if pgcombinebackup_version is None or pg_version is None or pg_version < "17":
+            # Return here. We are unable to retrieve
+            # pg_combinebackup or PostgreSQL versions
+            return remote_status
```
barthisrael, line 2107: This `if` statement sounds useless.

barman/recovery_executor.py
andremagui: This returns a reversed iterator! Does it look ok to set docstring like this? Any suggestion?

barman/recovery_executor.py
andremagui: dict[str, str|bool] is this ok? Usually when i used docstrings and annotations we use that typing module with List[str, Union[str,bool]].

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,317 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+            for tablespace in backup_info.tablespaces:
+                source = backup_info.get_data_directory(tablespace_oid=tablespace.oid)
+                destination = synthetic_backup_info.get_data_directory(
+                    tablespace_oid=tablespace.oid
+                )
+                tbs_map[source] = destination
+                dest_dirs.append(destination)
+
+        # Prepare the destination directories for pgdata and tablespaces
+        for d in dest_dirs:
```
gcalacoci, line 1904: let's avoid using single char vars... I know that this happens in the already existing code, but is wrong and I would like to avoid this.
```suggestion
        for dir in dest_dirs:
```

gcalacoci (changes requested):
I've added a small change request and asked a question about a possible default value for `recovery_staging_path`

barman/recovery_executor.py
barthisrael:
```suggestion
        :return Iterator[barman.infofile.LocalBackupInfo]: iterator of paths to
            backups in the chain, going from the full to the incremental backup
            pointed by *backup_info*
        """
```

You can use something like this, for example.

barman/cli.py
```diff
@@ -922,21 +922,34 @@ def recover(args):
                 output.error("Cannot parse recovery staging path: %s", str(exc))
                 output.close_and_exit()
             server.config.recovery_staging_path = recovery_staging_path
-        # If the backup is compressed but there is no recovery_staging_path
-        # then this is an error - the user *must* tell barman where recovery
-        # data can be staged.
+
+        # If no staging path is found then we need to display a proper error
+        # according to the backup type (compressed or incremental)
         if server.config.recovery_staging_path is None:
```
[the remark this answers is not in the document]
barthisrael: IIRC `recovery_staging_path` is a path in the target host when used for compressed backups.
So, for compressed backups, it's an intermediate directory where Barman copies the compressed backups to, then later the backup is decompressed and moved to the final destination pointed by `barman recover`.

A default value could make the experience when recovering incremental backup more seamless. But I don't think we should assume paths inside Barman home for that GUC -- unless we set a "default" only when recovering an incremental backup. I mean something like:

* Is incremental backup recovery?
  * Is recovery staging path not set?
    * Maybe issue a warning
    * Set recovery staging path as you proposed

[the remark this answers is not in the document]
andremagui: I think @gustabowill wanted to follow the tarball standards, so it does not have a default value, but requires a recovery_staging_path, right?? IMO setting the recovery_staging_path for the recovery is a little useless in this case because we combine in the staging and delete it after the copying/moving is done to the final destination. So for the UX, it is almost hidden (only have messages as output.info) because he/she won't have access to the combined backup directory after all. 
But i think a default value is nice to have! Then we do not require the user to set recovery_staging_path at all, only if he/she actually needs it. Then we could follow @barthisrael previous suggestion.

What are your thoughts? Can i go ahead with @barthisrael suggestion? @martinmarques @barthisrael  @gcalacoci  @gustabowill  @bleidens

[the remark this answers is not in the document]
gcalacoci: my question was around "should we change the actual behavior?"
the fact is: if we don't set it up, we error out. 

Is this what we want or do we want to change?
> IMO setting the recovery_staging_path for the recovery is a little useless in this case because we combine in the staging and delete it after the copying/moving is done to the final destination.

True that is something that get's whiped out after the recovery.
BUT, while you are recovering in an emergency, the less stuff you configure, the better.

BTW, good feedback so far guys.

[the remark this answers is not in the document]
barthisrael: > True that is something that get's whiped out after the recovery.
> BUT, while you are recovering in an emergency, the less stuff you configure, the better.

Right!

That's why I thought of having a default only when recovering an incremental backup, which default can be more predictable/acceptable in that case.

A `WARNING`, or even an `INFO` message, would be cool to have in the output, just so the user is aware of what's going on.
I think it is worth it a note in the docs later, too, so the user is aware that we will temporarily consume some considerable disk space when combining backups, and the location is set by `--recovery-staging-path`, defaulting to location "X".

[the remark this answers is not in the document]
gustabowill: I'm starting to wonder if it our previous idea of having a specific config for the combine staging directory was maybe the best option 🤔. It seems good to have a default value for this, but it seems like it makes sense for one use of `--recovery-staging-path` and not all, maybe having a specific param for the combine directory would make it easier to set a default and also facilitate explanations in the docs later?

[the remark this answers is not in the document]
martinmarques: I strongly feel that having a default value will lead to users not thinking about it, and having it fill the disk where they have their backups and WALs.
I vote to make this a mandatory option when recovering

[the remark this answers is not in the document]
gcalacoci: well, before that was something used only for tar stuff... that was sort of ok... but the way I see incremental going in the future is possibly going to be how people are going to do backups by default, because is faster and uses less space.
THIS changes the importance of the option, IMHO.

@barthisrael having a default only for incrementals is possibly confusing IMHO.

@martinmarques I see your point, and is something that "scares me" too... that's why I started this conversation.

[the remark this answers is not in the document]
barthisrael: > well, before that was something used only for tar stuff... that was sort of ok... but the way I see incremental going in the future is possibly going to be how people are going to do backups by default, because is faster and uses less space.
THIS changes the importance of the option, IMHO.
> 
> @barthisrael having a default only for incrementals is possibly confusing IMHO.

Yeah, maybe if we decide for using a default, a separate option, let's say `--combine-staging-path` would suit best now. As I mentioned initially, I thought there would be room for reusing `--recovery-staging-path`, until we eventually find that would not fit 😆 

> @martinmarques I see your point, and is something that "scares me" too... that's why I started this conversation.
Yeah, when reviewing the PR I was ok with forcing the user to choose a path as that can consume a lot of disk space. I'm afraid that if we use defaults, that can be "destructive" for the user environment.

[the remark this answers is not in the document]
gcalacoci: I propose to continue with the actual behaviour and then test the whole incremental backup and recovery experience manually a couple of times to get a feeling of the real-world usage.

After that, we can re-evaluate this. I don't want to introduce too much complexity or problems now

@martinmarques @andremagui @gustabowill @bleidens @barthisrael

barman/recovery_executor.py
andremagui: I have set the `exclude_dirs` to `exclude_path_names`. How does it sound?

barthisrael: Better :)

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,318 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+
+        # Then procede to move the content of the data directory
+        # We don't move the pg_tblspc as the _prepare_tablespaces method called earlier
+        # in the process already created this directory and required symlinks in the destination
+        data_source = backup_info.get_data_directory()
+        self._move_to_destination(
+            source=data_source, destination=dest, exclude_path_names={"pg_tblspc"}
+        )
+
+    def _move_to_destination(self, source, destination, exclude_path_names={}):
```
barthisrael, line 1996:
```suggestion
    def _move_to_destination(self, source, destination, exclude_path_names=set()):
```

While `{"pg_tblspc"}` does create a `set`, `{}` creates an empty dict. Let's use `set()` to create an empty set properly.

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,318 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        """
+        Move a single directory or file to its destination directory
```
barthisrael, line 1998:
```suggestion
        Move all files and directories contained within *source* to *destination*.
```

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,318 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+
+        :param str source: the source path to be moved
+        :param str destination: the destination path to move the source to
```
barthisrael, line 2001:
```suggestion
        :param str destination: the destination directory path where to move the
            files and directories contained within *source*
```

barman/recovery_executor.py
```diff
@@ -1807,13 +1810,318 @@ def _backup_copy(self, backup_info, dest, remote_command=None, **kwargs):
...
+        # Test pg_combinebackup existence
+        version_info = PgCombineBackup.get_version_info(self.server.path)
+
+        if version_info["full_path"]:
+            remote_status["pg_combinebackup_installed"] = True
+            remote_status["pg_combinebackup_path"] = version_info["full_path"]
+            remote_status["pg_combinebackup_version"] = version_info["full_version"]
+        else:
+            remote_status["pg_combinebackup_installed"] = False
+            return remote_status
```
barthisrael, line 2098:
```suggestion
            return remote_status
```

Doesn't add value. The next instructions is exactly the same.

tests/test_recovery_executor.py
barthisrael, line 1: I'm missing unit tests for:

* `IncrementalRecoveryExecutor._move_to_destination`
* `IncrementalRecoveryExecutor._prepare_destination`
* `IncrementalRecoveryExecutor._retry_handler`
* `IncrementalRecoveryExecutor._start_message`
* `IncrementalRecoveryExecutor._end_message`

The last 2 seem useless, I know. But good to have just in case we accidentally change some message which a user script could be eventually looking for in the future.

barthisrael (changes requested):

tests/test_recovery_executor.py
andremagui: Those were added in a000c5aff6bd149dfdda38a1bebbad540a67d90d

barthisrael:
I sent a commit with a suggestion for `test_cli` unit tests (basically combining 3 different methods into a single one which uses `pytest.parametrize`).

tests/test_cli.py
```diff
@@ -815,13 +876,67 @@ def test_recover_recovery_staging_path(
...
+        with pytest.raises(
+            SystemExit,
+        ):
+            recover(mock_recover_args)
+
+            error_mock.assert_called_once_with(
+                "Cannot recover from backup 'test_backup_id' of server "
+                "'test_server': backup status is not DONE"
+            )
+
```
barthisrael, line 939:
```suggestion
```
I don't understand what this test is checking. It seems to me we can remove this.

barthisrael (changes requested):
Review of `test_cli.py`.

tests/test_recovery_executor.py
```diff
@@ -2312,3 +2332,706 @@ def test_simple_file_mangeling(self, tmpdir):
...
+        )
+
+        mock_prepare_dest.assert_has_calls(
+            prepare_calls["with_mapping"], any_order=False
+        )
+        mock_move_to_dest.assert_has_calls(move_calls["with_mapping"], any_order=False)
+
+    def test__backup_copy_remote(self, server, executor):
+        """
+        Unit test for the _backup_copy method with tablespaces.
```
barthisrael, line 2653:
```suggestion
        Unit test for the _backup_copy method with remote option.
```

barthisrael (changes requested):
Review of `test_recovery_executor.py`

barthisrael:
I submitted a couple of fixes to the unit tests along with a final round of review/suggestions.
After the suggestions are applied, I think we can work on making the commit log cleaner and proceed with merging.

tests/test_cli.py
andremagui: Changed the name of the method to test_recover_backup_status and kept the test!

barthisrael (approved):
I've been performing several tests with this PR and it seems to behave consistently and as expected in all cases.

* Recovery with no staging path -- errors out because staging path is missing:

```console
$ barman recover pg17 incremental /home/vagrant/data_restore
ERROR: Cannot recover from backup 'incremental' of server 'pg17': backup will be combined with pg_combinebackup in the barman host but no recovery staging path is provided. Either set recovery_staging_path in the Barman config or use the --recovery-staging-path argument.
```

* Local recovery with different user and without ts mapping -- errors out because original ts path is not writable:

```console
$ barman recover pg17 incremental --recovery-staging-path /tmp  /home/vagrant/data_restore
Start combining backup via pg_combinebackup for backup 20240723T152539 on /tmp
End combining backup via pg_combinebackup for backup 20240723T152539
Starting local restore for server pg17 using backup 20240723T152539
Destination directory: /home/vagrant/data_restore
Using safe horizon time for smart rsync copy: 2024-07-23 13:25:39.491584+00:00
ERROR: unable to prepare 'custom_ts_1' tablespace (destination '/var/lib/pgsql/17/tablespaces/custom_ts_1'): mkdir execution failed
```

* Local recovery with different user and ts mapping -- succeeds:

```console
$ barman recover pg17 incremental --recovery-staging-path /tmp /home/vagrant/data_restore --tablespace custom_ts_1:/home/vagrant/tablespaces/custom_ts_1 --tablespace custom_ts_2:/home/vagrant/tablespaces/custom_ts_2
Start combining backup via pg_combinebackup for backup 20240723T152539 on /tmp
End combining backup via pg_combinebackup for backup 20240723T152539
Starting local restore for server pg17 using backup 20240723T152539
Destination directory: /home/vagrant/data_restore
	24678, custom_ts_1, /home/vagrant/tablespaces/custom_ts_1
	24686, custom_ts_2, /home/vagrant/tablespaces/custom_ts_2
Copying the base backup.
Copying required WAL segments.
Generating archive status files
Identify dangerous settings in destination directory.

Recovery completed (start time: 2024-07-23 16:05:47.810279+02:00, elapsed time: 1 second)
Your PostgreSQL server has been successfully prepared for recovery!
```

* "Remote" recovery without ts mapping -- succeeds:

```console
$ barman recover pg17 incremental --recovery-staging-path /tmp --remote-ssh-command "ssh postgres@localhost" /var/lib/pgsql/17/data
Start combining backup via pg_combinebackup for backup 20240723T152539 on /tmp
End combining backup via pg_combinebackup for backup 20240723T152539
Starting remote restore for server pg17 using backup 20240723T152539
Destination directory: /var/lib/pgsql/17/data
Remote command: ssh postgres@localhost
Using safe horizon time for smart rsync copy: 2024-07-23 13:25:39.491584+00:00
	24678, custom_ts_1, /var/lib/pgsql/17/tablespaces/custom_ts_1
	24686, custom_ts_2, /var/lib/pgsql/17/tablespaces/custom_ts_2
Copying the base backup.
Copying required WAL segments.
Generating archive status files
Identify dangerous settings in destination directory.

Recovery completed (start time: 2024-07-23 16:06:14.782003+02:00, elapsed time: 20 seconds)
Your PostgreSQL server has been successfully prepared for recovery!
```

* "Remote" recovery with ts mapping -- succeeds:

```console
$ barman recover pg17 incremental --recovery-staging-path /tmp --remote-ssh-command "ssh postgres@localhost" /var/lib/pgsql/17/data_restore --tablespace custom_ts_1:/var/lib/pgsql/17/random_ts
Start combining backup via pg_combinebackup for backup 20240723T152539 on /tmp
End combining backup via pg_combinebackup for backup 20240723T152539
Starting remote restore for server pg17 using backup 20240723T152539
Destination directory: /var/lib/pgsql/17/data_restore
Remote command: ssh postgres@localhost
Using safe horizon time for smart rsync copy: 2024-07-23 13:25:39.491584+00:00
	24678, custom_ts_1, /var/lib/pgsql/17/random_ts
	24686, custom_ts_2, /var/lib/pgsql/17/tablespaces/custom_ts_2
Copying the base backup.
Copying required WAL segments.
Generating archive status files
Identify dangerous settings in destination directory.

Recovery completed (start time: 2024-07-23 16:06:52.320212+02:00, elapsed time: 22 seconds)
Your PostgreSQL server has been successfully prepared for recovery!
```

I think this PR is a good shape to be merged.

barthisrael:
(FWIW I was able to start all "succeeded" Postgres instances from the tests shared through the previous comment).

barman/cli.py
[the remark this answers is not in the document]
martinmarques: I think that as long as you can put a value as a configuration option in the .conf, the user will only get an exception on the first try without setting the staging dir, then they configure it and done.

martinmarques (changes requested):

gcalacoci:
I promised @barthisrael to look at this before leaving for PTO 😄 , and I always keep my promises!

Important things first: 
I've executed the unit tests, set up a pg17 and executed a local backup and restore without tablespaces, I didn't go into deep testing like @barthisrael has done.

I like this PR and to me, it is in good shape, just need to apply @martinmarques suggestion and squash in 2 or 3 commits instead of 7, once those things are done, consider this comment as my approval!

barthisrael:
Thanks for also taking a look at this @gcalacoci -- and enjoy your Vacations 😆 !

I agree with applying the suggestion from Martin, but I guess we should have those 7 commits, though.
This PR is big, and thinking of it now we could've split this into at least 3 cards:

* One for creating a command wrapper for `pg_combinebackup`, which has been done through the first commit;
* One for aborting a recover operation if no `--recovery-staging-path` is provided, which has been done through the second commit;
* One for creating the recovery executor, which has been done through the third commit;

The fourth commit is about fixing tests which got broken by the commits in this PR, and the fifth about introducing unit tests to cover the changes from this PR.

The 6th is an "unrelated change" which improves the unit test suite, and the 7th is an "unrelated change" which was introduced by Actions.

To sum up my view of this: yes, I'm also bothered about the amount of commits, but I guess that's a consequence of the big task which was not properly split by us.

With that in mind I'd say to proceed like this:

1. Apply the suggestion from @martinmarques and squash that into the third commit (321ccce4d5ef25d7241c879d1c1cfa2ca89f9560);
2. Remove the changes related with `command_wrappers.py` file from the third commit (321ccce4d5ef25d7241c879d1c1cfa2ca89f9560), and "squash" those changes into the first commit (d6985887677dbb51bbb45b42c6ce309ec1a966fa). What I mean here is: the first commit is in charge of adding the `pg_combinebackup` command wrapper, so the suggestion that I made in regards to its docstring should be part of that commit, not of the one which adds a recovery executor;
3. Reword a bit the second commit (d2f07ae7b5a1240f93aa09ef6a553139d994b787). It says we are modifying the `recover` method to use `--recovery-staging-path`, but in fact we are only making sure the value is set when recovering an incremental backup, and aborting execution if that's not the case.

Makes sense?

barman/infofile.py
```diff
@@ -956,3 +965,41 @@ def is_full_and_eligible_for_incremental(self):
...
+            A synthetic backup is a base backup which was artificially created
+            through ``pg_combinebackup``. A synthetic backup is not part of
+            the Barman backup catalog, and only exists so we are able to
+            recover a backup created by ``pg_combinebackup`` utility, as
+            almost all functions and methods require a backup info object.
+
+        The only difference from this class to its parent :class:`LocalBackupInfo`
+        is that it accepts a custom base directory for the backup as synthetic
+        backups are expected to live on directories other than the default
+        ``<server_name>/base`` path.
```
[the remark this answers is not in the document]
gustabowill: I don't think these docstrings were wrong 🤔

The standard path of a backup in the catalog is <server_name> -> base -> <backup_id>, which means the base path is <server_name>/base, which is where all the backup folders live. What this comment was saying is that a synthetic backup can have a different base path for its folder, so whole path would be <custom_base_path> -> <backup_id>, in this case the base path is just <custom_base_path> and the backup folder lives there. So I think it makes sense the way it was. @martinmarques @barthisrael @andremagui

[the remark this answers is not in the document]
barthisrael: That's true, that's the other way around.

barman/infofile.py
```diff
@@ -956,3 +965,41 @@ def is_full_and_eligible_for_incremental(self):
...
+            A synthetic backup is a base backup which was artificially created
+            through ``pg_combinebackup``. A synthetic backup is not part of
+            the Barman backup catalog, and only exists so we are able to
+            recover a backup created by ``pg_combinebackup`` utility, as
+            almost all functions and methods require a backup info object.
+
+        The only difference from this class to its parent :class:`LocalBackupInfo`
+        is that it accepts a custom base directory for the backup as synthetic
+        backups are expected to live on directories other than the default
+        ``<server_name>/<backup_id>/base`` path.
```
barthisrael, line 987:
```suggestion
        ``<server_name>/base`` path.
```

PR #958, opened 2024-07-04
Require cramjam lib to be >= 2.7.0

gcalacoci:
The library `python-snappy` internally uses the library cramjam from versions > 0.6.1.
This doesn't work well with versions of cramjam < 2.7.0 as the structure of the code is different than the one used by python-snappy.

References: bar-156

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,11 @@ setup(
             "grpcio",
             "google-cloud-compute",  # requires minimum python3.7
         ],
-        "snappy": ["python-snappy==0.6.1"],
+        "snappy": [
+            'python-snappy==0.6.1; python_version<"3.7"',
+            'python-snappy; python_version>="3.7"',
+            'cramjam >= 2.7.0; python_version>="3.7"',
+        ],
     },
     platforms=["Linux", "Mac OS X"],
     classifiers=[


barthisrael (changes requested):
Changes look good.

I would extend a bit the commit message with a note explaining that cramjan >= 2.7.0 is not available for Python 3.6, so we need to pin python-snappy to an older version (0.6.1) in that case, to avoid the broken relationship of newer snappy with older cramjan

gcalacoci:
> Changes look good.
> 
> I would extend a bit the commit message with a note explaining that cramjan >= 2.7.0 is not available for Python 3.6, so we need to pin python-snappy to an older version (0.6.1) in that case, to avoid the broken relationship of newer snappy with older cramjan

I'll rework the commit message. re-reading it doesn't seem clear as it was during the writing

gcalacoci:
> > Changes look good.
> > I would extend a bit the commit message with a note explaining that cramjan >= 2.7.0 is not available for Python 3.6, so we need to pin python-snappy to an older version (0.6.1) in that case, to avoid the broken relationship of newer snappy with older cramjan
> 
> I'll rework the commit message. re-reading it doesn't seem clear as it was during the writing

re-reading my commit message:
```
Also for python < 3.7 new `cramjam` libs are not available.

So we pinned `python-snappy` to 0.6.1 for python < 3.7, and set use
the new `python-snappy` and `cramjam` versions for python >= 3.7 .
```
@barthisrael you think is not understandable enough?

barthisrael:
@gcalacoci I think the second looks good.

barthisrael (approved):
Looks great!

Issue #943, opened 2024-06-20
Collecting ideas on how to speed up recovery


toydarian:
I have a fairly large database and currently replaying WALs takes over 24 hours.  
I'm trying to find ways to speed this up and would appreciate some input.  
Up until now, I have found two possible improvements that could be done on the `barman-wal-restore` script.

 - The first one is in [`try_deliver_from_spool`](https://github.com/EnterpriseDB/barman/blob/master/barman/clients/walrestore.py#L237) where the file is actually copied instead of moved. Assuming we don't run on a copy-on-write file-system and the spool is on the same file-system as `pg_wal`, it would be faster to move or hard-link the file instead of copying.
 - As far as I understand it, even when running with multiple parallel processes fetching files, the script will fetch n files, then PostgreSQL will replay those n files and then ask for the next one, where `barman-wal-restore` will fetch the next batch of n files and the whole thing starts from the beginning. I wonder if it would be possible to continuously fetch files, so the database never has to wait for any files to get delivered.

Just to be clear, I don't expect anybody to implement any of this. I'm collecting ideas which I plan to implement myself.

PR #960, opened 2024-07-08
uses shutil.move instead of copy and unlink

toydarian:
This should speed up serving wal-files from the spool-directory, in case `pg_wal` and the spool-directory are on the same filesystem, as in this case the file is renamed instead of copied.  
If those directories are on different file-systems, this doesn't change anything, as `shutil.move` will do the same 2-step-process (copy and remove) in that case. [Docs for reference](https://docs.python.org/3/library/shutil.html#shutil.move).

Also see ticket #943

### DIFF
--- a/barman/clients/walrestore.py
+++ b/barman/clients/walrestore.py
@@ -31,6 +31,7 @@ import shutil
 import subprocess
 import sys
 import time
+from pathlib import Path
 
 import barman
 from barman.utils import force_str
@@ -243,15 +244,14 @@ def try_deliver_from_spool(config, dest_file):
     :param argparse.Namespace config: the configuration from command line
     :param dest_file: The destination file object
     """
-    spool_file = os.path.join(config.spool_dir, config.wal_name)
+    spool_file = Path(config.spool_dir, config.wal_name)
 
     # id the file is not present, give up
-    if not os.path.exists(spool_file):
+    if not spool_file.exists():
         return
 
     try:
-        shutil.copyfileobj(open(spool_file, "rb"), dest_file)
-        os.unlink(spool_file)
+        shutil.move(spool_file, dest_file)
         sys.exit(0)
     except IOError as e:
         exit_with_error(
--- a/barman/clients/walrestore.py
+++ b/barman/clients/walrestore.py
@@ -31,7 +31,6 @@ import shutil
 import subprocess
 import sys
 import time
-from pathlib import Path
 
 import barman
 from barman.utils import force_str
@@ -75,7 +74,7 @@ def main(args=None):
         return  # never reached
 
     # If the file is present in SPOOL_DIR use it and terminate
-    try_deliver_from_spool(config, dest_file)
+    try_deliver_from_spool(config, dest_file.name)
 
     # If required load the list of files to download in parallel
     additional_files = peek_additional_files(config)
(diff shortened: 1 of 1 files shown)


barman/clients/walrestore.py
```diff
@@ -243,15 +244,14 @@ def try_deliver_from_spool(config, dest_file):
     :param argparse.Namespace config: the configuration from command line
     :param dest_file: The destination file object
     """
-    spool_file = os.path.join(config.spool_dir, config.wal_name)
+    spool_file = Path(config.spool_dir, config.wal_name)
```
gcalacoci, line 247: Do you have a specific reason for using `Pathlib` instead of `os.path` here?

We are planning to do this and other changes in the whole codebase at once, so if is not necessary for your patch we would like to avoid introducing this kind of changes right now.

I hope that his makes sense to you.

barman/clients/walrestore.py
```diff
@@ -243,15 +244,14 @@ def try_deliver_from_spool(config, dest_file):
...
 
     # id the file is not present, give up
-    if not os.path.exists(spool_file):
+    if not spool_file.exists():
         return
 
     try:
-        shutil.copyfileobj(open(spool_file, "rb"), dest_file)
-        os.unlink(spool_file)
+        shutil.move(spool_file, dest_file)
```
gcalacoci, line 254: We discussed this and yes we do agree that this change is a performance improvement in the same fs scenario you described. I'm fine with merging this once we remove the `Pathlib` usage above

gcalacoci (changes requested):

barman/clients/walrestore.py
toydarian: I prefer `Pathlib` and in that case used it to work around typing-issues. Reverted that change.

I also found another issues I addressed in this commit.

barman/clients/walrestore.py
```diff
@@ -241,21 +241,20 @@ def try_deliver_from_spool(config, dest_file):
     return otherwise.
 
     :param argparse.Namespace config: the configuration from command line
-    :param dest_file: The destination file object
+    :param dest_file: The path to the destination file
     """
-    spool_file = os.path.join(config.spool_dir, config.wal_name)
+    spool_file = str(os.path.join(config.spool_dir, config.wal_name))
```
gcalacoci, line 246: isn't the return value of `os.path.join` already a string?

barman/clients/walrestore.py
gcalacoci: I do prefer `Pathlib` too, but barman is has a really old codebase that we kept compatible with pyhton 2 for a long time. Said so, we really want to bring in some "modern" changes into the barman code soon :)

barman/clients/walrestore.py
toydarian: No, it's a `LiteralString`. Probably okay, but I try keeping the types correct.

gcalacoci (approved):
LGTM

martinmarques:
The only thing I'd like to see added is a note to the docs regarding the performance, but conversations are happening inside the team to redesign the whole docs, so I'm going to add an internal ticket for us to track this little piece of doc changes to the architecture part.

martinmarques (approved):

barthisrael (approved):

PR #961, opened 2024-07-08
Configurable timeout when awaiting snapshot completion on AWS

mikewallace1979:
Adds a new config option `aws_await_snapshots_timeout` and corresponding
barman-cloud-backup option `--aws-await-snapshots-timeout` which can be
used to set the amount of time Barman will wait for snapshots to reach a
completed state.

This has a default value of 3600 which is significantly higher than the
300 second timeout resulting from the original boto3 default `Delay` of
15s and `MaxAttempts` of 40.

---

Tested by:

- Provisioning a new EC2 instance and immediately shutting it down.
- Creating three gp2 EBS volumes of 4096GB size and attaching them to the above instance at mount points `/dev/sdb`, `/dev/sdc` and `/dev/sdd`.
- Configuring Barman to take snapshot backups of this instance but using a fake `ssh_command` of `ssh barman@localhost` so that the volume mount checks are using the local host and not the real instance.
- Creating a mock `findmnt` script and putting it on the path of the barman user - the script is just `echo $4 rw,nosuid,nodev,noexec,relatime`, so it always appears to find whatever device it is asked about.

This is enough to allow `barman backup` and `barman-cloud-backup` to complete a snapshot backup, where the snapshots of the actual volumes in AWS are taken and the new timeout option can be observed to be working.

The backup operation took between 1 and 2 minutes. I tested with `aws_await_snapshots_timeout` values of below 60s and observed backups to fail, tested with the default value and observed backups to succeed, and tested with values above 120s and observed backups to succeed.


barthisrael (commented):
Overall looks good -- had only a couple of questions related with the snapshot interface.

ruimarinho:
Just hit this issue and would love to see this merged/released. First time snapshots with the disk size I'm using takes much longer than the default 300 seconds, so a configurable timeout would be great!

martinmarques:
> Just hit this issue and would love to see this merged/released. First time snapshots with the disk size I'm using takes much longer than the default 300 seconds, so a configurable timeout would be great!

We are targeting this for the next release. That will be, hopefully, next month

barman/cloud_providers/aws_s3.py
```diff
@@ -462,7 +463,7 @@ class AwsCloudSnapshotInterface(CloudSnapshotInterface):
         https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-creating-snapshot.html
     """
 
-    def __init__(self, profile_name=None, region=None):
+    def __init__(self, profile_name=None, region=None, await_snapshots_timeout=3600):
```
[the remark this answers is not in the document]
mikewallace1979: Yes! Thanks for raising this.

barman/cloud_providers/__init__.py
```diff
@@ -197,7 +197,12 @@ def get_snapshot_interface(config):
     elif config.cloud_provider == "aws-s3":
         from barman.cloud_providers.aws_s3 import AwsCloudSnapshotInterface
 
-        return AwsCloudSnapshotInterface(config.aws_profile, config.aws_region)
+        args = [config.aws_profile, config.aws_region]
+        # aws_await_snapshots_timeout is only supported by barman-cloud-backup so not
+        # always present
+        if config is not None and hasattr(config, "aws_await_snapshots_timeout"):
+            args += [config.aws_await_snapshots_timeout]
```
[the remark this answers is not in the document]
barthisrael: Great, thanks for clarifying!

barman/cloud_providers/aws_s3.py
```diff
@@ -462,18 +463,30 @@ class AwsCloudSnapshotInterface(CloudSnapshotInterface):
...
         # If a specific region was provided then this overrides any region which may be
         # defined in the profile
         self.region = region or self.session.region_name
         self.ec2_client = self.session.client("ec2", region_name=self.region)
+        self.await_snapshots_timeout = await_snapshots_timeout
+
+    def _get_waiter_config(self):
+        delay = 15
+        # Use ceil so that we always wait for at least the specified timeout
+        max_attempts = math.ceil(self.await_snapshots_timeout / delay)
```
barthisrael, line 485: Should we treat a `await_snapshots_timeout` of `0` in a "special" way, or we just let it pass over?

barthisrael (commented):
@mikewallace1979 thanks for the clarifications and changes, the PR looks mostly good!

I posted a final question, and took the liberty of sending a commit with docs changes. What do you think of this?

barman/cloud_providers/aws_s3.py
mikewallace1979: This is a good question.

If `await_snapshots_timeout` is `0` then currently `max_attempts` will work out as `0` as well.

As it happens, the current implementation will still make one attempt if we pass `"MaxAttempts": 0` in the `WaiterConfig`: https://github.com/boto/botocore/blob/547b20801770c8ea4255ee9c3b809fea6b9f6bc4/botocore/waiter.py#L379

However, despite being a de facto part of the boto3 interface, `WaiterConfig` [remains undocumented](https://github.com/boto/boto3/issues/712) so we shouldn't rely on the current implementation.

I think deliberately setting `MaxAttempts` to `1` in this case would be worthwhile just to be safe.

barthisrael: Cool, agree with you!

barthisrael (approved):
Looks good!

PR #966, opened 2024-07-23
Get database cluster size and add to `backup.info`

andremagui:
After the introduction of incremental backups in Postgres 17, there is the need to review how we display information about backups, specially when using the `show-backup` command. Because of this need, we are adding the `cluster size` field to the backup.info. 
This new field is intended to be used in the calculation of the deduplication_ratio of backups and to show the estimated size of the Postgres cluster at the time the backup was taken. For this, we are reusing an existing property of the `PostgreSQLConnection` class called `current_size`.

The tests broken by the changes on the code were fixed.

Unit tests for the `current_size` property were added because they were found to missing from the test suite and the unit test for the `_pg_get_metadata` method of the `BackupStrategy` class was extended because of the changes in the method. 

References: BAR-215

### DIFF
--- a/barman/backup_executor.py
+++ b/barman/backup_executor.py
@@ -1787,6 +1787,9 @@ class BackupStrategy(with_metaclass(ABCMeta, object)):
             summarize_wal = self.postgres.get_setting("summarize_wal")
             backup_info.set_attribute("summarize_wal", summarize_wal)
 
+        # Set total size of the PostgreSQL server
+        backup_info.set_attribute("cluster_size", self.postgres.current_size)
+
     @staticmethod
     def _backup_info_from_start_location(backup_info, start_info):
         """
--- a/tests/test_postgres.py
+++ b/tests/test_postgres.py
@@ -1748,6 +1748,72 @@ class TestPostgres(object):
         # THEN a None value was returned
         assert has_monitoring is None
 
+    @pytest.mark.parametrize(
+        ("backup_privilege", "expected", "should_fetchone"),
+        [(False, None, False), (True, 2048, True)],
+    )
+    @patch("barman.postgres.PostgreSQLConnection.connect")
+    @patch(
+        "barman.postgres.PostgreSQLConnection.has_backup_privileges",
+        new_callable=PropertyMock,
+    )
+    def test_current_size(
+        self,
+        has_backup_privileges_mock,
+        conn_mock,
+        backup_privilege,
+        expected,
+        should_fetchone,
+    ):
+        """
+        Test the current_size method
+        """
+        any_size = 2048
+
+        has_backup_privileges_mock.return_value = backup_privilege
+
+        # Build a server
+        server = build_real_server()
+        cursor_mock = conn_mock.return_value.cursor.return_value
+        cursor_mock.fetchone.side_effect = [[any_size]]
+
+        result = server.postgres.current_size
+        assert result == expected
+        if should_fetchone:
+            cursor_mock.execute.assert_called_once_with(
+                "SELECT sum(pg_tablespace_size(oid)) FROM pg_tablespace"
+            )
+        else:
+            cursor_mock.assert_not_called()
+
+    @pytest.mark.parametrize(
+        "expected_error", (psycopg2.Error, PostgresConnectionError)
+    )
+    @patch("barman.postgres._logger.debug")
+    @patch("barman.postgres.PostgreSQLConnection.connect")
+    @patch(
+        "barman.postgres.PostgreSQLConnection.has_backup_privileges",
+        new_callable=PropertyMock,
+    )
+    def test_current_size_error(
+        self, has_backup_privileges_mock, conn_mock, logger_mock, expected_error
+    ):
+        """
+        Test the current_size method
+        """
+        has_backup_privileges_mock.return_value = True
+
+        # Build a server
+        server = build_real_server()
+        cursor_mock = conn_mock.return_value.cursor.return_value
+        cursor_mock.fetchone.side_effect = expected_error
+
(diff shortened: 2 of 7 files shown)


barman/backup_executor.py
```diff
@@ -1787,6 +1787,9 @@ def _pg_get_metadata(self, backup_info):
             summarize_wal = self.postgres.get_setting("summarize_wal")
             backup_info.set_attribute("summarize_wal", summarize_wal)
 
+        # Set total size of the PostgreSQL server
+        backup_info.set_attribute("cluster_size", self.postgres.current_size)
```
barthisrael, line 1791: `current_size` uses the total size of tablespaces.

```sql
postgres=# SELECT sum(pg_database_size(oid)) FROM pg_database;
    sum
-----------
 665407665
(1 row)
```

```sql
postgres=# SELECT sum(pg_tablespace_size(oid)) FROM pg_tablespace ;
    sum
-----------
 666034993
(1 row)
```

Who should we rely on, databases or tablespaces?

(Just joking).

Good catch on using that existing property.

barthisrael (commented):
The code changes seems fine.
I'll take a closer look at the unit tests tomorrow.

I'd suggest providing a few more details in some commit messages, e.g.:

* cc14dffe034107b8a24c0af93762fe345cba15c6: provide a heads up on why we are adding that new field, what will that be used for?

barman/backup_executor.py
andremagui: I found this property but also checked the pg_database_size! What i could find about differences between is that pg_tablespace_size also sums the size of global objects, that's why the size is greater using tablespaces!

tests/test_postgres.py
```diff
@@ -1748,6 +1748,26 @@ def test_has_monitoring_privileges_exception(self, conn_mock, mock_is_superuser)
...
+    @patch("barman.postgres.PostgreSQLConnection.connect")
+    def test_current_size(self, conn_mock, has_backup_privileges_mock):
+        """
+        Test the current_size method
+        """
+        size = 2048
+        has_backup_privileges_mock.return_value = True
+        # Build a server
+        server = build_real_server()
+        conn_mock.return_value.server_version = 110000
```
martinmarques, line 1761: I suppose this was copied from other mocks (didn't check), but shouldn't we test with newer versions? Maybe it's ok/better to test with the oldest supported version.

P.S.: we should have named this sever_version_num which is the field in pg_settings. But that's outside the scope of this PR

andremagui: We can remove the server_version! It is not mentioned in the method itself! I got it wrongly from the xlog_segment_size property! I will fix

tests/test_postgres.py
```diff
@@ -1748,6 +1748,25 @@ def test_has_monitoring_privileges_exception(self, conn_mock, mock_is_superuser)
         # THEN a None value was returned
         assert has_monitoring is None
 
+    @patch("barman.postgres.PostgreSQLConnection.has_backup_privileges")
+    @patch("barman.postgres.PostgreSQLConnection.connect")
+    def test_current_size(self, conn_mock, has_backup_privileges_mock):
```
barthisrael, line 1753: For completeness, you can add test cases for:

* `has_backup_privileges` == `True`
* `PostgresConnectionError` and `psycopg2.Error` being triggered.

barthisrael (changes requested):
The implementation itself works fine. I can see the field being set independently of the type of the backup. Here I have 4 backups (one "full" and one "incremental" for each method):

```console
$ find ~/barman_home/ -name backup.info -print -exec bash -c "grep -i cluster {}; echo" \;
/home/vagrant/barman_home/pg17/base/20240724T164407/backup.info
cluster_size=666034993

/home/vagrant/barman_home/pg17/base/20240724T164417/backup.info
cluster_size=666034993

/home/vagrant/barman_home/support/base/20240724T164351/backup.info
cluster_size=666034993

/home/vagrant/barman_home/support/base/20240724T164556/backup.info
cluster_size=666034993
```

Besides the comment in the `test_postgres.py` file, I propose the following changes:

* Squash 52a36ab9f7ab5d6eb77c467287f2f977a0c5fdac and 6049129353aa00270149bc33a99ae6ce5077f8da
* Cover the changes of `_pg_get_metadata` in `test__pg_get_metadata`
* Add a note to 70be908e58405e7b6fd00196de6b7d4121db1a51 to make it clear that that commit is adding unit tests which were missing in the test suite, for code which is reused by this PR

martinmarques (approved):

barthisrael (changes requested):
This looks good. Now, just a final round of "nitpicking" 😄 

* c2dd301c47cd59eefc76058bc4cfc2b225c3a744:
   * We mention "This new field is intended to be used in the calculation of the deduplication_ratio of incremental backups.". We also intend to use that for showing the estimated cluster size on `show-backup` output.
   * Maybe it's worth it to update the PR description to contain this information as well. The PR description is missing a `References` BTW.
* Switch order between 8d74d0776fada650da4729d0facef5d43b8c740a and 7b42685c5ee5a2fad65da6e6ca147f89bbf4b21d: thinking of a "logical flow of changes", I'd say it makes more sense to first fix what got broken, then introduce new tests;
* 8d74d0776fada650da4729d0facef5d43b8c740a:
   * This doesn't only add unit tests for `current_size` method, but also extends the tests of `_pg_get_metadata`;
   * Regarding the note "Note: The unit tests for this method were added because they were found to be missing from the test suite coverage.", I would make that note more clear in the sense that it refers to tests for the existing `cluster_size` which were missing before you started reusing it for filling the `backup.info` new field

barthisrael (approved):
Looks good!

PR #967, opened 2024-07-23
Output info from the check-backup command

gustabowill:
The `check-backup` command can mark a backup as failed if it notices that some of its required WAL files are missing. This works correctly, however no output is provided, so the only way to check why the backup has been marked as failed is by executing a `show-backup` on the backup. This commit fixes this issue by outputting the reason as an error message in the CLI.

Note: I though about using debug or warning log level, but considering that, as I've seen, there's a change of this happening also during a backup, I think it's better to display it with an error level instead. Also, if by any chance the user runs this manually they should be able to see the failure reason immediately on the screen I think.

References: BAR-213

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -2014,7 +2014,7 @@ class TestServer(object):
     @patch("barman.infofile.BackupInfo.save")
     @patch("os.path.exists")
     def test_check_backup(
-        self, mock_exists, backup_info_save, tmpdir, orig_exists=os.path.exists
+        self, mock_exists, backup_info_save, tmpdir, capsys, orig_exists=os.path.exists
     ):
         """
         Test the check_backup method
@@ -2090,6 +2090,8 @@ class TestServer(object):
             "The first missing WAL file is "
             "000000010000000000000003"
         )
+        _, err = capsys.readouterr()
+        assert backup_info.error in err
         backup_info_save.reset_mock()
 
         # Case 4: the more recent WAL archived is more recent than the end
@@ -2129,6 +2131,8 @@ class TestServer(object):
             "The first missing WAL file is "
             "000000010000000000000004"
         )
+        _, err = capsys.readouterr()
+        assert backup_info.error in err
         backup_info_save.reset_mock()
 
         # Case 4.3: we have all the files, but the backup is marked as
--- a/barman/backup.py
+++ b/barman/backup.py
@@ -1564,6 +1564,10 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
             )
             backup_info.status = BackupInfo.FAILED
             backup_info.save()
+            output.error(
+                "This backup has been marked as FAILED due to the "
+                "following reason: %s" % backup_info.error
+            )
             return
 
         if end_wal <= last_archived_wal:


barthisrael:
I rebased this on top of `master` to get rid of that `black: format` commit, which was already handled before.

barthisrael:
> Note: I though about using debug or warning log level, but considering that, as I've seen, there's a change of this happening also during a backup, I think it's better to display it with an error level instead. Also, if by any chance the user runs this manually they should be able to see the failure reason immediately on the screen I think.

That makes sense!
After all, `backup_info.error` field is filled with that information for later showing through `barman show-backup` command.

barthisrael:
This looks good. A test showing the functionality:

* Get a couple of backups:

```console
$ barman list-backup all
pg17 20240726T210201 - Fri Jul 26 19:02:02 2024 - Size: 35.4 MiB - WAL Size: 0 B - WAITING_FOR_WALS
pg17 20240726T210125 - Fri Jul 26 19:01:28 2024 - Size: 251.3 MiB - WAL Size: 32.0 MiB - WAITING_FOR_WALS
```

* Check WALs needed by one backup:

```console
$ barman show-backup pg17 20240726T210125 | grep WAL
  Status                 : WAITING_FOR_WALS
    Disk usage           : 219.3 MiB (251.3 MiB with WALs)
    Begin WAL            : 0000000100000005000000F1
    End WAL              : 0000000100000005000000F2
    WAL number           : 2
  WAL information:
    WAL rate             : 303.79/hour
```

* Remove the WAL from archive to force the issue:

```console
$ rm ~/barman_home/pg17/wals/0000000100000005/0000000100000005000000F1
```

* Attempt to check the backup:

```console
$ barman check-backup pg17 20240726T210125
ERROR: This backup has been marked as FAILED due to the following reason: At least one WAL file is missing. The first missing WAL file is 0000000100000005000000F1
```

* The same information is present in the logs:

```console
2024-07-26 21:03:59,624 [27471] barman.backup ERROR: This backup has been marked as FAILED due to the following reason: At least one WAL file is missing. The first missing WAL file is 0000000100000005000000F1
```

* And the same information is shown by `barman show-backup` command:

```console
$ barman show-backup pg17 20240726T210125
Backup 20240726T210125:
  Server Name            : pg17
  System Id              : 7382227563826631291
  Status                 : FAILED
  Error                  : At least one WAL file is missing. The first missing WAL file is 0000000100000005000000F1
```

tests/test_server.py
```diff
@@ -2090,6 +2090,8 @@ def mock_os_path_exist(file_name):
             "The first missing WAL file is "
             "000000010000000000000003"
         )
+        _out, err = capsys.readouterr()
```
barthisrael, line 2093:
```suggestion
        _, err = capsys.readouterr()
```

We can just assign to `_` so PyLance stops complaining in the IDE 😄

tests/test_server.py
```diff
@@ -2129,6 +2131,8 @@ def mock_os_path_exist(file_name):
             "The first missing WAL file is "
             "000000010000000000000004"
         )
+        _out, err = capsys.readouterr()
```
barthisrael, line 2134:
```suggestion
        _, err = capsys.readouterr()
```

barthisrael (changes requested):
PR looks good.
Just put a couple of suggestions to silence PyLance in the IDE.

barthisrael (approved):
Looks good!

PR #974, opened 2024-07-31
Adds a keep-alive mechanism to Rsync backups

gustabowill:
This PR adds a keep-alive mechanism to Rsync backups. This is necessary to guarantee that `pg_backup_start()` and `pg_backup_stop()` are executed within the same Postgres session.

This issue has been reported by some users and is most likely due to a firewall rule that kills connections that have been idle for a specified amount of time. In such cases, the stop command ends up being executed in a different connection session than the start command and Postgres gives an error, thus the backup in Barman is considered as `FAILED` since the full process has not been completed successfully and Barman is not able to guarantee its consistency.

d8faad23a4a6c90357a46a73130eba6eeb1d3636 adds the new `keepalive-interval` as both a server config and a `backup` command argument.
eed0c4b4b1e9645db531632e68278782e2269c77 fixes some tests broken due to the new `keepalive-interval` parameter.
3cf515880aa813a40fa519f81f3637d8465b3bf8 implements a keep-alive mechanism on Rsync backups by sending heartbeat queries at a specified interval using the current connection.
65fbd6a77f1ace125e28f0b76447dca71421656e adds tests to the keep-alive mechanism introduced in the previous commit.

References: BAR-192.


gustabowill:
While I'm not able to replicate the firewall scenario locally, an alternative way to test this is just by passing a very small number to `keepalive-interval` on a Rsync backup. Then you can check the debug logs in the log file, which will indicate the query is being executed, or check the query history with something like `log_min_duration_statement = 0 `in the Postgres server.

barthisrael:
I initially thought your code would break because of the lack of locking when reusing Python objects, like Postgres connections, but turns out `psycopg2` made our lives easier in that sense as they implement thread-safety in the module, so we don't need to care about that -- see https://www.psycopg.org/docs/usage.html#thread-and-process-safety.

However, I was not able to infere from their release notes when that thread-safety mechanism was introduced -- so I guess that's since the beginning?

Just to be on the safe side, I checked the behavior with the oldest `psycopg2` provided by the supported operating systems repositorie, which is `2.7.5` (on RHEL 8).

I installed that version of `psycopg2` and ran a Python script that spawns several threads to "concurrently" create several tables (size between 1MB and 350MB each table) using the same connection object. As described in the module docs, it seems that `psycopg2` is indeed thread-safe and that it serializes the access to the connection object, thus executing one command after another.

That shouldn't be an issue for us as we are executing either `pg_backup_start`/`pg_backup_stop` or `SELECT 1` queries through that reused connection, function calls that should be fast enough.

In short what I mean is: we might be safe as `psycopg2` seems to serialize access to the Python connection object, and our queries are fast enough so the queries won't wait long to execute.

barman/postgres.py
```diff
@@ -1813,3 +1839,72 @@ def stop_exclusive_backup(self):
...
+        manager 4) Now the main thread have to wait for more 200 seconds until this
+        thread ends sleeping, even though all the work is already finished. To solve
+        this, instead of sleeping 300 seconds directly, we sleep 1 second 300 times
+        and each time we check if the thread should still be alive.
+        """
+        for _ in range(self.interval):
+            if self._stop_thread.is_set():
+                return
+            time.sleep(1)
+
```
barthisrael, line 1891: What about using `self._stop_thread.wait(self.interval)` instead?
That method does what you are trying to accomplish here, i.e. "sleep for N seconds unless something wakes it up" (`set` wakes it up, see https://docs.python.org/3/library/threading.html#threading.Event.set).

barthisrael (changes requested):

barman/postgres.py
```diff
@@ -1813,3 +1839,72 @@ def stop_exclusive_backup(self):
...
+    def _run_keep_alive(self):
+        """Runs the keepalive until a stop-thread event is set"""
+        while not self._stop_thread.is_set():
+            if not self.postgres.has_connection:
+                # Wait for the connection to be opened on the main thread
+                continue
+            self.postgres.send_heartbeat_query()
```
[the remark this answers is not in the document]
barthisrael: I think we can keep that.

Maybe we can add a note to the commit message that introduces the method, just so it's clear why we opted out for keeping the return value.

barman/postgres.py
barthisrael: Yes, I'd say to use the mechanism exposed by the `threading` module. The code will essentially do something similar, but in a cleaner way.

barthisrael: We can remove the function and directly call `.wait(self.interval)` in `_run_keep_alive`.

gustabowill:
@barthisrael Just applied all suggestions so far.

barman/postgres.py
```diff
@@ -498,6 +501,10 @@ def connect(self):
                 raise PostgresAppNameError(force_str(e).strip())
         return self._conn
 
+    @property
+    def has_connection(self):
```
barthisrael, line 505: Add a docstring.

barman/postgres.py
```diff
@@ -1602,6 +1609,24 @@ def get_synchronous_standby_names(self):
             # the format of the synchronous_standby_names content
             return [x.strip().strip('"') for x in names_list.split(",")]
 
+    def send_heartbeat_query(self):
+        """
+        Sends a heartneat query to the server with the already opened connection.
```
barthisrael, line 1614:
```suggestion
        Sends a heartbeat query to the server with the already opened connection.
```

barman/postgres.py
```diff
@@ -1813,3 +1838,56 @@ def stop_exclusive_backup(self):
...
+class PostgresKeepAlive:
+    """
+    Context manager to maintain a Postgres connection alive.
+
+    A child thread is spawned to execute heartbeat queries in the background
+    at a specified interval during its living context.
+
+    It does not open or close any connections on its own. Instead, it waits for the
+    specified connection to be opened in the main thread, making use of the same
+    current transaction.
```
barthisrael, line 1852:
```suggestion
    current session.
```

Session is a more appropriate word in this context.

barman/postgres.py
```diff
@@ -1813,3 +1838,56 @@ def stop_exclusive_backup(self):
...
+    """
+
+    THREAD_NAME = "barman_keepalive_thread"
```
barthisrael, line 1855: Add `THREAD_NAME` to the class docstring. The syntax is something like this:

```python
"""
(...)
:cvar THREAD_NAME: description...
(...)
"""

File: barman/postgres.py [hunk 7]
```diff
@@ -1813,3 +1838,56 @@ def stop_exclusive_backup(self):
...
+
+    def __init__(self, postgres, interval):
+        """
+        Constructor.
+
+        :param barman.postgres.PostgreSQLConnection postgres: The
+            Postgres connection to keep alive.
+        :param int interval: An interval in seconds at which a
+            heartbeat query will be sent to keep the connection alive.
+            A value <= 0 won't start the keepalive.
```
barthisrael, line 1865:
```suggestion
            A value <= ``0`` won't start the keepalive.
```

Double backticks for literals.

barthisrael (changes requested):

barthisrael:
This seems to work well while everything is fine with the connection.

I don't have a big cluster here to properly test the implementation, so I just simulated that by adding a `time.sleep(60)` in the beginning of the method `barman.backup_executor.RsyncBackupExecutor.backup_copy`.

This was my `barman backup` command:

```console
$ barman backup support --keepalive-interval 3
Starting backup using rsync-concurrent method for server support in /home/vagrant/barman_home/support/base/20240806T231753
Backup start at LSN: 7/84000028 (000000010000000700000084, 00000028)
This is the first backup for server support
WAL segments preceding the current backup have been found:
	000000010000000700000082 from server support has been removed
Starting backup copy via rsync/SSH for 20240806T231753
```

In the meantime I could see the following log entries:

```console
$ tail -20 /var/log/barman/barman.log
2024-08-06 23:17:53,996 [656420] barman.backup_executor INFO: Backup start at LSN: 7/84000028 (000000010000000700000084, 00000028)
2024-08-06 23:17:53,999 [656420] barman.backup_executor INFO: This is the first backup for server support
2024-08-06 23:17:54,010 [656420] barman.backup_executor INFO: 	000000010000000700000082 from server support has been removed
2024-08-06 23:17:54,011 [656420] barman.backup_executor INFO: Starting backup copy via rsync/SSH for 20240806T231753
2024-08-06 23:17:54,061 [607070] barman.command_wrappers INFO: pg17: pg_receivewal: finished segment at 7/84000000 (timeline 1)
2024-08-06 23:17:54,067 [607072] barman.command_wrappers INFO: support: pg_receivewal: finished segment at 7/84000000 (timeline 1)
2024-08-06 23:17:56,759 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:17:59,761 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:02,764 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:05,765 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:08,776 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:11,777 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:14,780 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:17,782 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:20,789 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:23,791 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:26,792 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:29,798 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:32,800 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:18:35,802 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
```

So the thread was correctly sending the heartbeats in the meantime.

This can also be observed later, once the actual copy starts:

```
2024-08-06 23:19:01,236 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/3764
2024-08-06 23:19:01,237 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/3764_fsm
2024-08-06 23:19:01,238 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/3764_vm
2024-08-06 23:19:01,238 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/3766
2024-08-06 23:19:01,243 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/3767
2024-08-06 23:19:01,243 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/3997
2024-08-06 23:19:01,244 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/41088
2024-08-06 23:19:02,814 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:19:05,815 [656420] barman.postgres DEBUG: Sent heartbeat query to maintain the current connection
2024-08-06 23:19:06,167 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/41088_fsm
2024-08-06 23:19:06,168 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/41088_vm
2024-08-06 23:19:06,168 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/41091
2024-08-06 23:19:06,171 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/41094
2024-08-06 23:19:06,695 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/41094_fsm
2024-08-06 23:19:06,696 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/41094_vm
2024-08-06 23:19:06,696 [656808] RsyncPgData DEBUG: >f+++++++++ base/5/4143
```

This is the full output from start to end of a successful backup:

```console
$ barman backup support --keepalive-interval 3
Starting backup using rsync-concurrent method for server support in /home/vagrant/barman_home/support/base/20240806T231753
Backup start at LSN: 7/84000028 (000000010000000700000084, 00000028)
This is the first backup for server support
WAL segments preceding the current backup have been found:
	000000010000000700000082 from server support has been removed
Starting backup copy via rsync/SSH for 20240806T231753
Copy done (time: 11 seconds)
This is the first backup for server support
Asking PostgreSQL server to finalize the backup.
Backup size: 600.5 MiB. Actual size on disk: 600.5 MiB (-0.00% deduplication ratio).
Backup end at LSN: 7/84000158 (000000010000000700000084, 00000158)
Backup completed (start time: 2024-08-06 23:17:54.012874, elapsed time: 1 minute, 14 seconds)
Processing xlog segments from streaming for support
	000000010000000700000083
	000000010000000700000084
```

barthisrael (changes requested):
The code of this PR works great when everything is working fine, i.e. it should avoid that firewalls drop the session held by Barman as it is no longer idle.

However, we need to handle issues in a cleaner way. Currently, if any issue is faced with the connection, as reconnection is not allowed, the thread no longer does anything after that. Besides that, uncaught exceptions are thrown into logs and console ,and the backup copy keeps ongoing even if we know for a fact that the stop backup command will fail.

I suggest the following changes in the behavior so we get a smoother experience:

* Have `send_heartbeat_query` to catch other exceptions as well, avoiding that stack traces go to console and log, but rather that we have a proper error message handling;
* If `send_heartbeat_query` fails to run the heartbeat because of any kind of connection exception, we should abort the backup with a proper message.

barman/cli.py
```diff
@@ -484,6 +484,15 @@ def backup_completer(prefix, parsed_args, **kwargs):
             default=None,
             type=check_non_negative,
         ),
+        argument(
+            "--keepalive-interval",
+            help="An interval, in seconds, at which a heartbeat query will be sent "
+            "to the server to keep the connection alive during an Rsync backup. "
```
martinmarques, line 490: "keep the libpq connection alive", so nobody thinks this is related to the rsync connections

barman/cli.py
```diff
@@ -484,6 +484,15 @@ def backup_completer(prefix, parsed_args, **kwargs):
...
+            "Default is 60 seconds. A value of 0 disables it",
+            dest="keepalive_interval",
+            default=None,
```
martinmarques, line 493: ???
Isn't the default 60?

martinmarques (changes requested):

barman/cli.py
gcalacoci: good point

barman/cli.py
```diff
@@ -534,6 +543,8 @@ def backup(args):
             server.config.basebackup_retry_sleep = args.retry_sleep
         if args.retry_times is not None:
             server.config.basebackup_retry_times = args.retry_times
+        if args.keepalive_interval is not None:
```
gcalacoci, line 546: As mentioned by @martinmarques default should be 60, and here we should check not for none but for 0 value, I think. need to do a bit of research as argparse sometimes is confusing. let's discuss @gustabowill @barthisrael

barthisrael: The arg parsed by `argparse` contains the `default` value when the user specifies nothing. If no `default` is specified, it brings `None`.

```python
import argparse

# Create the parser
parser = argparse.ArgumentParser(description="Example script with two optional integer arguments.")

# Add optional arguments
parser.add_argument(
    '--first',
    type=int,
    default=60,
    help='The first integer argument (default: 60).'
)

parser.add_argument(
    '--second',
    type=int,
    help='The second integer argument (no default value).'
)

# Parse the arguments
args = parser.parse_args()

# Access the arguments
print(f"First argument: {args.first}")
print(f"Second argument: {args.second}")
```

->

```console
$ python test.py
First argument: 60
Second argument: None
```

barman/cli.py
```diff
@@ -484,6 +484,15 @@ def backup_completer(prefix, parsed_args, **kwargs):
             default=None,
             type=check_non_negative,
         ),
+        argument(
+            "--keepalive-interval",
+            help="An interval, in seconds, at which a heartbeat query will be sent "
+            "to the server to keep the connection alive during an Rsync backup. "
+            "Default is 60 seconds. A value of 0 disables it",
```
barthisrael, line 491:
```suggestion
```

I think we can keep the code as it is and remove the information about the default.
We should add the new parameter to the docs, and there we can explain the default value for the setting, and what it does in more details.

gustabowill:
@barthisrael @gcalacoci  I think I've applied everything we discussed so far.

barman/postgres.py
```diff
@@ -1813,3 +1843,107 @@ def stop_exclusive_backup(self):
...
+            from a child thread in Python. A handler for this signal has been set
+            beforehand by the ``_prepare_signal_handler`` method in this class.
+        """
+        os.kill(os.getpid(), signal.SIGUSR1)
+
+    def _run_keep_alive(self):
+        """Runs the keepalive until a stop-thread event is set"""
+        while not self._stop_thread.is_set():
+            if not self.postgres.has_connection:
+                # Wait for the connection to be opened on the main thread
```
barthisrael, line 1923: I wonder if we should have a very short sleep time here before going to the next loop iteration?
I mean, just so we avoid running "continue" statement several times waiting for the connection to be ready.

gcalacoci: 1s is enough

barthisrael:
The implementation works fine.

* I start a backup with an aggressive keep-alive (`barman backup rsync --keepalive-interval 1`)
* I run the following SQL in Postgres shortly after the backup starts:

```sql
postgres=# SELECT pid, pg_terminate_backend(pid) FROM pg_stat_activity WHERE usename = 'barman' ;
   pid   | pg_terminate_backend
---------+----------------------
 1299991 | t
(1 row)
```

* The backup fails with a proper message:

```console
$ barman backup rsync --keepalive-interval 1
Starting backup using rsync-concurrent method for server rsync in /home/vagrant/barman_home/rsync/base/20240814T165704
Backup start at LSN: 0/CE000028 (0000000100000000000000CE, 00000028)
This is the first backup for server rsync
WAL segments preceding the current backup have been found:
	0000000100000000000000CC from server rsync has been removed
Starting backup copy via rsync/SSH for 20240814T165704
ERROR: Backup failed copying files.
DETAILS: Connection to the Postgres server was lost during the backup.
Processing xlog segments from streaming for rsync
	0000000100000000000000CD
```

barthisrael (approved):
Looks good!

gcalacoci (approved):
LGTM, evaluate squashing in 2 commits before merging

gustabowill:
> LGTM, evaluate squashing in 2 commits before merging

You mean just one for the whole implementation and one for the tests? I preferred to keep the changes that touch the CLI in separate commits because (as usual) it broke some existing tests, this way I was able to let the "actual implementation" in the last two commits, instead of having fixes and other things together with that. Idk, what do you think about this structure?

barthisrael:
> You mean just one for the whole implementation and one for the tests? I preferred to keep the changes that touch the CLI in separate commits because (as usual) it broke some existing tests, this way I was able to let the "actual implementation" in the last two commits, instead of having fixes and other things together with that. Idk, what do you think about this structure?

The reasoning makes sense to me.
I had a small chatting with Giulio and we are merging this now.

PR #975, opened 2024-08-01
Fix "backup size" of rsync incremental backups upon deletion of backups

bleidens:
When deleting a backup, if the next backup is rsync it might contain outdated disk usage and deduplication information, which should therefore be re-calculated after the deletion. This occurs because rsync backups have deduplication at the filesystem level when using the "reuse_backup = link" option, and since the deduplication size is calculated at the time the backup is taken, deleting an rsync backup might cause the deduplication information of the subsequent backup to become inaccurate.

With the implemented changes, whenever a backup is deleted from the catalog, if there is a next rsync backup in the server, its disk information will be updated using the new method  `_set_backup_sizes`. This method updates backup information and can optionally also fsync to disk, so `backup_fsync_and_set_sizes` was also updated to reuse it. 

References: BAR-217

### DIFF
--- a/barman/backup.py
+++ b/barman/backup.py
@@ -526,6 +526,7 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
                 remove_until, timelines_to_protect, wal_ranges_to_protect
             ):
                 output.info("\t%s", name)
+
         # As last action, remove the backup directory,
         # ending the delete operation
         try:
@@ -558,6 +559,22 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
                 parent_backup.children_backup_ids = None
             parent_backup.save()
 
+        # rsync backups can have deduplication at filesystem level by using
+        # "reuse_backup = link". The deduplication size is calculated at the
+        # time the backup is taken. If we remove a backup, it may be the case
+        # that the next backup in the catalog is a rsync backup which was taken
+        # with the "link" option. With that possibility in mind, we re-calculate the
+        # deduplicated size of the next rsync backup because the removal of the
+        # previous backup can impact on that number.
+        # Note: we have no straight forward way of identifying if the next rsync
+        # backup in the catalog was taken with "link" or not because
+        # "reuse_backup" value is not stored in the "backup.info" file. In any
+        # case, the "re-calculation" can still be performed even if "link" was not
+        # used, and the only drawback is that we will waste some (small) amount
+        # of CPU/disk usage.
+        if next_backup and next_backup.backup_type == "rsync":
+            self._set_backup_sizes(next_backup)
+
         # Remove the sync lockfile if exists
         sync_lock = ServerBackupSyncLock(
             self.config.barman_lock_directory, self.config.name, backup.backup_id
@@ -588,6 +605,36 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
         script.run()
 
         return True
+
+    def _set_backup_sizes(self, backup_info, fsync=False):
+        """
+        Set the actual size on disk of a backup.
+
+        Optionally fsync all files in the backup.
+
+        :param LocalBackupInfo backup_info: the backup to update
+        :param bool fsync: whether to fsync files to disk
+        """
+        backup_size = 0
+        deduplicated_size = 0
+        backup_dest = backup_info.get_basebackup_directory()
+        for dir_path, _, file_names in os.walk(backup_dest):
+            if fsync:
+                # If fsync, execute fsync() on the containing directory
+                fsync_dir(dir_path)
+            for filename in file_names:
+                file_path = os.path.join(dir_path, filename)
+                # If fsync, execute fsync() on all the contained files
+                file_stat = fsync_file(file_path) if fsync else os.stat(file_path)
+                backup_size += file_stat.st_size
+                # Excludes hard links from real backup size and only counts
+                # unique files for deduplicated size
+                if file_stat.st_nlink == 1:
+                    deduplicated_size += file_stat.st_size
+        # Save size into BackupInfo object
+        backup_info.set_attribute("size", backup_size)
+        backup_info.set_attribute("deduplicated_size", deduplicated_size)
+        backup_info.save()
 
     def validate_backup_args(self, **kwargs):
         """
@@ -1469,23 +1516,10 @@ class BackupManager(RemoteStatusMixin, KeepManagerMixin):
         # Calculate the base backup size
         self.executor.current_action = "calculating backup size"
         _logger.debug(self.executor.current_action)
-        backup_size = 0
-        deduplicated_size = 0
-        backup_dest = backup_info.get_basebackup_directory()
-        for dir_path, _, file_names in os.walk(backup_dest):
-            # execute fsync() on the containing directory
-            fsync_dir(dir_path)
-            # execute fsync() on all the contained files
-            for filename in file_names:
-                file_path = os.path.join(dir_path, filename)
-                file_stat = fsync_file(file_path)
-                backup_size += file_stat.st_size
-                # Excludes hard links from real backup size
-                if file_stat.st_nlink == 1:
-                    deduplicated_size += file_stat.st_size
-        # Save size into BackupInfo object
-        backup_info.set_attribute("size", backup_size)
-        backup_info.set_attribute("deduplicated_size", deduplicated_size)
+        # Set backup sizes with fsync. We need to fsync files here to make sure
+        # the backup files are persisted to disk, so we don't lose the backup in
+        # the event of a system crash.
+        self._set_backup_sizes(backup_info, fsync=True)
         if backup_info.size > 0:
             deduplication_ratio = 1 - (
                 float(backup_info.deduplicated_size) / backup_info.size
--- a/tests/test_backup.py
+++ b/tests/test_backup.py
@@ -26,7 +26,7 @@ import dateutil.parser
 import dateutil.tz
 import mock
 import pytest
-from mock import Mock, patch
+from mock import Mock, patch, call
 from barman.backup import BackupManager
 from barman.lockfile import ServerBackupIdLock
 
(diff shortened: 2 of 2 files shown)


barman/backup.py
```diff
@@ -582,6 +582,12 @@ def delete_backup(self, backup, skip_wal_cleanup_if_standalone=True):
                 e.hook.script,
             )
 
+        # Update next rsync backup size information
+        if backup.config.reuse_backup:
+            next_backup = self.get_next_backup(backup.backup_id)
+            if next_backup:
+                self.backup_fsync_and_set_sizes(next_backup)
```
barthisrael, line 589: We should not rely on volatile information -- the user can change the configuration file.

We should rather rely on the state of the backup, which is stored in its metadata.

Also, simply reusing the `backup_fsync_and_set_sizes` here seems not the best path:

* It will perform a lot of `fsync` syscalls, what puts unnecessary overhead;
* That function has some logging/output which are tied to a backup operation.

I'd say to do something different here, like:

* Create a new method `def _set_backup_sizes(self, backup_info, fsync=False)`
   * It would take care of going through the files
      * If `fsync` then fsync to disk, otherwise simply stat the file
   * Set the attributes in `backup_info` accordingly

That way we could reuse that internal method both when deleting and when backing up.

barthisrael (changes requested):

barman/backup.py
```diff
@@ -582,6 +582,12 @@ def delete_backup(self, backup, skip_wal_cleanup_if_standalone=True):
                 e.hook.script,
             )
 
+        # Update next rsync backup size information
+        if backup.config.reuse_backup:
```
martinmarques, line 586: We should check that the backup is a rsync backup instead of a postgres backup. There is no need to check if reuse backup was set to link, as Israel mentioned

martinmarques: This shoulf be checked from the `backup.info`

barman/backup.py
```diff
@@ -558,6 +559,10 @@ def delete_backup(self, backup, skip_wal_cleanup_if_standalone=True):
                 parent_backup.children_backup_ids = None
             parent_backup.save()
 
+        # Update next rsync backup size information
+        if next_backup and next_backup.backup_type == "rsync":
+            self._set_backup_sizes(next_backup)
```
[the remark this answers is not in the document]
gcalacoci: @bleidens I would add this as comment in the code. As rule of thumb, if you felt the need to specify this with the above comment, possibly you want to specify it in the code too 😄

barman/backup.py
```diff
@@ -1456,23 +1485,11 @@ def backup_fsync_and_set_sizes(self, backup_info):
...
-                # Excludes hard links from real backup size
-                if file_stat.st_nlink == 1:
-                    deduplicated_size += file_stat.st_size
+        # Set backup sizes with fsync
+        self._set_backup_sizes(backup_info, fsync=True)
         # Save size into BackupInfo object
-        backup_info.set_attribute("size", backup_size)
-        backup_info.set_attribute("deduplicated_size", deduplicated_size)
+        backup_info.set_attribute("size", backup_info.size)
+        backup_info.set_attribute("deduplicated_size", backup_info.deduplicated_size)
```
barthisrael, line 1492: This doesn't seem to do anything 😆 
I mean, it essentially sets `backup_info.size` with the value of `backup_info.size`.

barman/backup.py
```diff
@@ -558,6 +559,11 @@ def delete_backup(self, backup, skip_wal_cleanup_if_standalone=True):
                 parent_backup.children_backup_ids = None
             parent_backup.save()
 
+        # Update next rsync backup size information
+        # This assumes that rsync can create hard-links to postgres backups
```
barthisrael, line 563:
```suggestion
        # rsync backups can have deduplication at filesystem level by using
        # "reuse_backup = link". The deduplication size is calculated at the
        # time the backup is taken. If we remove a backup, it may be the case
        # that the next backup in the catalog is a rsync backup which was taken
        # with the "link" option. With that possibility in mind, we re-calculate the
        # deduplicated size of the next rsync backup because the removal of the
        # previous backup can impact on that number.
        # Note: we have no straight forward way of identifying if the next rsync
        # backup in the catalog was taken with "link" or not because
        # "reuse_backup" value is not stored in the "backup.info" file. In any
        # case, the "re-calculation" can still be performed, and the only
        # drawback is that we will waste some (small) amount of CPU/disk
        # usage.
```

@bleidens @gcalacoci I think this provides a more clear idea here.
What do you think?

barman/backup.py
```diff
@@ -589,6 +595,30 @@ def delete_backup(self, backup, skip_wal_cleanup_if_standalone=True):
 
         return True
 
+    def _set_backup_sizes(self, backup_info, fsync=False):
+        """
+        Set the actual size on disk of a backup.
+        Optionally fsync all files in the backup.
```
barthisrael, line 601:
```suggestion

        Optionally fsync all files in the backup.
```

Line-break between summary and "description".

barman/backup.py
```diff
@@ -1456,23 +1486,8 @@ def backup_fsync_and_set_sizes(self, backup_info):
...
-                file_path = os.path.join(dir_path, filename)
-                file_stat = fsync_file(file_path)
-                backup_size += file_stat.st_size
-                # Excludes hard links from real backup size
-                if file_stat.st_nlink == 1:
-                    deduplicated_size += file_stat.st_size
-        # Save size into BackupInfo object
-        backup_info.set_attribute("size", backup_size)
-        backup_info.set_attribute("deduplicated_size", deduplicated_size)
+        # Set backup sizes with fsync
```
barthisrael, line 1489:
```suggestion
        # Set backup sizes with fsync. We need to fsync files here to make sure
        # the backup files are persisted to disk, so we don't lose the backup in
        # the event of a system crash.
```

This looks more complete IMO.
Thoughts?

barthisrael (changes requested):

barman/backup.py
barthisrael: When writing comments to the code, let's first make sure they are needed, and if needed, that they provide enough information for us when reviewing the code in the future.

barman/backup.py
```diff
@@ -589,6 +595,30 @@ def delete_backup(self, backup, skip_wal_cleanup_if_standalone=True):
...
+        deduplicated_size = 0
+        backup_dest = backup_info.get_basebackup_directory()
+        for dir_path, _, file_names in os.walk(backup_dest):
+            if fsync:
+                fsync_dir(dir_path)
+            for filename in file_names:
+                file_path = os.path.join(dir_path, filename)
+                file_stat = fsync_file(file_path) if fsync else os.stat(file_path)
+                backup_size += file_stat.st_size
+                if file_stat.st_nlink == 1:
```
barthisrael, line 616:
```suggestion
                if file_stat.st_nlink == 1:
```

Here I'd say to add the comment back regarding what it means, just so it's handy.

tests/test_backup.py
```diff
@@ -366,6 +366,209 @@ def test_delete_backup(self, mock_available_backups, tmpdir, caplog):
...
         assert child_backup.backup_id not in parent_backup.children_backup_ids
 
+        # Test 8: Update next rsync backup information
+        given_backup = build_test_backup_info(
+            backup_id="parent_backup_id",
+            server=backup_manager.server,
+        )
+        build_backup_directories(given_backup)
+        next_backup = build_test_backup_info(
+            backup_id="next_backup_id",
```
barthisrael, line 376:
```suggestion
        given_backup = build_test_backup_info(
            backup_id="rsync_backup_id",
            server=backup_manager.server,
        )
        build_backup_directories(given_backup)
        next_backup = build_test_backup_info(
            backup_id="next_rsync_backup_id",
```

I think this naming will provide a clearer idea of what we are expecting here.

tests/test_backup.py
```diff
@@ -366,6 +366,209 @@ def test_delete_backup(self, mock_available_backups, tmpdir, caplog):
...
+        )
+        build_backup_directories(next_backup)
+        with patch("barman.backup.BackupManager.get_next_backup") as mock:
+            with patch(
+                "barman.backup.BackupManager._set_backup_sizes"
+            ) as set_backup_sizes:
+                mock.return_value = next_backup
+                deleted = backup_manager.delete_backup(given_backup)
+        assert deleted is True
+        set_backup_sizes.assert_called_with(next_backup)
```
barthisrael, line 387:
```suggestion
                mock.return_value = next_backup
                deleted = backup_manager.delete_backup(given_backup)
                assert deleted is True
                set_backup_sizes.assert_called_with(next_backup)
```

Let's put things in the correct indentation.

barthisrael (commented):
For sake of history: Barbara and I had a small call for guidance on how to simplify/fix the unit tests.

tests/test_backup.py
```diff
@@ -367,6 +367,145 @@ def test_delete_backup(self, mock_available_backups, tmpdir, caplog):
...
+            next_backup.backup_id: next_backup,
+        }
+        with patch("barman.backup.BackupManager.get_next_backup") as get_next_backup:
+            with patch(
+                "barman.backup.BackupManager._set_backup_sizes"
+            ) as set_backup_sizes:
+                get_next_backup.return_value = next_backup
+                deleted = backup_manager.delete_backup(given_backup)
+                assert deleted is True
+                set_backup_sizes.assert_called_with(next_backup)
```
barthisrael, line 392:
```suggestion
                set_backup_sizes.assert_called_once_with(next_backup)
```

I think "once" provides a more precise meaning here.

tests/test_backup.py
```diff
@@ -367,6 +367,145 @@ def test_delete_backup(self, mock_available_backups, tmpdir, caplog):
...
+                    call("/root/dir1/file2.txt"),
+                    call("/root/dir2/subdir/file3.txt"),
+                ]
+            )
+            mock_backup_info.set_attribute.assert_has_calls(
+                [
+                    call("size", 7168),
+                    call("deduplicated_size", 4096),
+                ]
+            )
```
barthisrael, line 488: I think this can be moved out of the `if-else` branch because it's a common assertion.

tests/test_backup.py
```diff
@@ -367,6 +367,145 @@ def test_delete_backup(self, mock_available_backups, tmpdir, caplog):
...
```
barthisrael, line 507: Similar thoughts here.

barthisrael (changes requested):

barthisrael:
> When deleting a backup, if the next is rsync it will contain outdated disk usage and deduplication information. With the implemented changes, whenever a backup is deleted from the catalog, if there is a next rsync backup in the server, its disk information will be updated using the new method _set_backup_sizes. This method updates backup information and can optionally also fsync to disk, so backup_fsync_and_set_sizes was also updated to reuse it.

The PR description needs a few teaks:

* It's missing a "References";
* it mentions "it will contain outdated disk usage and deduplication information". I would replace that "will" with "might", and also expand on why, maybe using a shorter version of the comment that was put in the code;

This new PR description could be relayed to the commit message too, as the commit message is not saying much and would require one to open the code to get a grasp on the changes.

barthisrael:
The implementation itself seems to be working fine. I took 2 rsync backups and played with deletions.

* Initial size:

```console
$ barman show-backup rsync rsync_1 | grep "Base" -A2
  Base backup information:
    Disk usage           : 22.2 MiB (38.2 MiB with WALs)
    Incremental size     : 22.2 MiB (-0.00%)

$ barman show-backup rsync rsync_2 | grep "Base" -A2
  Base backup information:
    Disk usage           : 195.2 MiB (211.2 MiB with WALs)
    Incremental size     : 174.3 MiB (-10.70%)
```

* Sizes after deleting backup `rsync_1`:

```console
$ barman show-backup rsync rsync_2 | grep "Base" -A2
  Base backup information:
    Disk usage           : 195.2 MiB (211.2 MiB with WALs)
    Incremental size     : 195.2 MiB (-0.00%)
```

barthisrael (approved):
Looks good.

Issue #200, opened 2019-02-27
Backup was deleted during recovery process


saifulmuhajir:
We have a server with retention policy for 10 days: `retention_policy: RECOVERY WINDOW OF 10 DAYS`

And here's the list-backup command shows:
```
$ barman list-backup db-test
db-test 20190223T042124 - Sat Feb 23 13:00:28 2019 - Size: 2.4 TiB - WAL Size: 29.8 GiB
db-test 20190220T052453 - Wed Feb 20 14:01:43 2019 - Size: 2.4 TiB - WAL Size: 41.7 GiB
db-test 20190216T105941 - Sat Feb 16 19:43:34 2019 - Size: 2.4 TiB - WAL Size: 43.9 GiB
db-test 20190211T205100 - Sat Feb 12 04:51:30 2019 - Size: 2.4 TiB - WAL Size: 25.9 GiB
```

We tried to restore to a remote server as follows:
```
$ barman recover db-test 20190211T205100 --remote-ssh-command "ssh postgres@destinationserver" /postgres/data96/
Processing xlog segments from streaming for db-test
        000000050000353100000041
Starting remote restore for server db-test using backup 20190211T205100
Destination directory: /postgres/data96/
Copying the base backup.
ERROR: Failure copying base backup: data transfer failure
rsync error:
rsync: change_dir "/barman/db-test/base/20190211T205100/data/" failed: No such file or directory (2)
rsync error: errors selecting input/output files, dirs (code 3) at flist.c(2118) [sender=3.1.2]
```

After some hours, this morning we found out that the recovery process was failed because the backup scheduled was running and the data we are restoring (oldest) was deleted.

```
$ barman list-backup db-test
db-test 20190227T050837 - STARTED
db-test 20190223T042124 - Sat Feb 23 13:00:28 2019 - Size: 2.4 TiB - WAL Size: 29.8 GiB
db-test 20190220T052453 - Wed Feb 20 14:01:43 2019 - Size: 2.4 TiB - WAL Size: 41.7 GiB
db-test 20190216T105941 - Sat Feb 16 19:43:34 2019 - Size: 2.4 TiB - WAL Size: 43.9 GiB
```

It would be good if `barman recover` process touch a _status file_ regarding the backup-id. When auto-delete based on retention policy or `barman delete` command is executed, the file is checked and if it is being used the delete process should fail.

amenonsen:
That makes sense. We'll look at this in conjunction with the `keep` feature (to override the retention policy).

PR #980, opened 2024-08-08
Add note in the manual about recovery near retention policy

andremagui:
This commit Closes: #200 .

It adds a note to the recovery section of the documentation manual. This note informs users about the risk of backup obsolescence and deletion during the recovery process, particularly when retention policy rules are enforced, and provides guidance on how to prevent this issue.

Initially, the team considered implementing an automatic solution for this problem. However, several concerns were raised, including the potential for unintended side effects during recovery, the existing `keep` feature designed to assist with backups, and the need for users to be fully aware of the implications and procedures during recovery—an operation often conducted under stressful situations.

References: BAR-222

### DIFF
--- a/doc/manual/43-backup-commands.en.md
+++ b/doc/manual/43-backup-commands.en.md
@@ -198,6 +198,16 @@ command.
 
 The recovery command has several options that modify the command behavior.
 
+> **IMPORTANT:**
+> If you're unsure whether the backup you want to recover from will be
+> automatically deleted by retention policies during recovery,
+> mark it issuing the [keep command](#keep) using the appropriate target.
+> Once the recovery process is complete and your new PostgreSQL instance
+> has reached the desired recovery target, if you don’t want to
+> keep the backup beyond the retention policy, you can remove `keep`
+> annotation issuing the [keep command](#keep) again using the
+> `--release` option.
+
 ### Remote recovery
 
 Add the `--remote-ssh-command <COMMAND>` option to the invocation


barthisrael (commented):
We will need to revisit this.

This implementation is fragile as it is because the WALs required by the recovery process to complete could be vanished once the KEEP is released.

We need to discuss a different approach.

doc/manual/43-backup-commands.en.md
```diff
@@ -198,6 +198,13 @@ command.
 
 The recovery command has several options that modify the command behavior.
 
+> **NOTE:**
```
gcalacoci, line 201:
```suggestion
> **IMPORTANT:**
```

doc/manual/43-backup-commands.en.md
```diff
@@ -198,6 +198,13 @@ command.
...
+> If you’re not sure whether the backup can be automatically removed by
+> retention policies during recovery, avoid issuing the `recover` command
+> right away. Instead, first mark the backup with `keep` and specify the
+> `target`. Once the recovery process is complete, if you don’t want to
+> keep the backup beyond the retention policy, you can then release it.
```
gcalacoci, line 206: Reworded for clarity
```suggestion
> If you're unsure whether the backup you want to recover from will be
> automatically deleted by retention policies during recovery,
> mark it using the [keep command](#keep) with the appropriate target.
> Once the recovery process is complete, if you don’t want to
> keep anymore the backup beyond the retention policy, you can remove
> the `keep` annotation issuing the [keep command](#keep) again
> using the `--release` option
```
@andremagui @martinmarques @barthisrael , thoughts?

andremagui: Yes, i agree! I thought about fixing 'full' as the target in the note but i was not sure about this.

andremagui: @barthisrael @martinmarques Any comments on this?

andremagui: TBH i think this is an assumption that is valid for people that did not understand the barman recovery process. I mean, if the user does not know what it needs for the recovery process to be complete, than it should keep the backup (source backup) until the new PostgreSQL instance is running (from the destination recovered backup) and tested the PIT target chosen. Does it make sense? If so, we can add more info about this.

gcalacoci: that's why i wrote "recovery process" and not recovery... but looks like is confusing 😄 I don't have a better Idea atm for wording differently

barthisrael: Maybe something like

"Once the recovery process is complete and your Postgres instance has reached the desired recovery target..."

barthisrael (approved):
Looks good.

gcalacoci (approved):

PR #983, opened 2024-08-08
Add new command line argument for `recover` command

andremagui:
Since the introduction of incremental backups in PostgreSQL 17, we created a new recovery executor to handle the operation. The plan was to reuse the `recovery_staging_path` argument/configuration that was designed to recover compressed backups PR #955 . After further review, it was decided that we would create a new argument/configuration specific to recover from incremental backups called `local_staging_path`.

This PR:

1. Add a new configuration option to the global/server configurations and a new command line argument for the recover command.
2. Change the name of a method called `parse_recovery_staging_path` to `parse_staging_path` so we have a generic name to parse staging_paths and keep the same operation for both.
3. Modify all unit tests that used `parse_recovery_staging_path` with the new naming.
4. Change the implementation of the recover CLI method to accommodate the new configuration option.
5. Modify the unit test for the recover CLI method according to the new implementation

References: Bar-269


andremagui:
I thought about using the new field as an alias for the recovery_staging_path but i think it would also make things a little confusing. Let me know about other ways of implementing this if you have any ideas.

barman/cli.py
```diff
@@ -890,6 +890,18 @@ def rebuild_xlogdb(args):
             help="The name of the AWS region containing the EC2 VM and storage "
             "volumes for recovery of a snapshot backup",
         ),
+        argument(
+            # TODO: change name of flag after definition
+            "--local-staging-path",
+            dest="local_staging_path",
```
andremagui, line 896: I thought about another implementation that would reuse the `recovery_staging_path` here in the `dest=recovery_staging_path` instead of using the new field. We could make --local-stating-path required for the recover of incremental backups. Then if there is a recovery_staging_path set in the configuration files, we would ignore it for incremental backups and warn the user that it needs to set `--local-staging-path`. By doing this, it would be hidden from the user that we are actually reusing the recovery_staging_path and we would have a new flag.
This way we would not need to create a new configuration field and it would be something like aliasing the existing configuration field. 

Let me know what you guys think, please! @barthisrael @gcalacoci @martinmarques @gustabowill  @bleidens

andremagui: we already discussed and the best approach would be to have another configuration field, instead of reusing it

andremagui: We are going with `--local-staging-path` after discussing this in the checkpoint call we had today.

barman/cli.py
```diff
@@ -899,6 +899,18 @@ def rebuild_xlogdb(args):
             help="The name of the AWS region containing the EC2 VM and storage "
             "volumes for recovery of a snapshot backup",
         ),
+        argument(
+            # TODO: change name of flag after definition
```
barthisrael, line 903:
```suggestion
```
We have a definition 😄

barman/cli.py
```diff
@@ -899,6 +899,18 @@ def rebuild_xlogdb(args):
...
+            "--local-staging-path",
+            dest="local_staging_path",
```
barthisrael, line 905:
```suggestion
```

`dest` is not needed in this case as by default `argparse` replaces `-` with `_` to build the variable in the namespace object, so the name would be equivalent.

(see https://github.com/python/cpython/blob/main/Lib/argparse.py#L1562-L1573)

barman/cli.py
```diff
@@ -918,47 +930,66 @@ def recover(args):
...
-            server.config.recovery_staging_path = recovery_staging_path
+    # If the user sets both flags, we output an error and exit.
+    # This prevents from staging_path being overwritten in the
+    # code "block" below.
+    if args.local_staging_path is not None and args.recovery_staging_path is not None:
+        output.error(
+            "Cannot set `recovery_staging_path` and "
+            "`local_staging_path` to recover.",
+        )
+        output.close_and_exit()
```
barthisrael, line 941: I think we should refactor this whole "staging path" validation here.

It might be clearer if we simply use two different `if` blocks, one for handling compression/recovery_staging_path, another for `incremental/local_staging_path`.

We used to have this code in Barman 3.10.1 for handling the former case: https://github.com/EnterpriseDB/barman/blob/release/3.10.1/barman/cli.py#L898-L925.
I think we should have that back, and add a similar block for the latter case.

barman/cli.py
```diff
@@ -899,6 +899,18 @@ def rebuild_xlogdb(args):
             help="The name of the AWS region containing the EC2 VM and storage "
             "volumes for recovery of a snapshot backup",
         ),
+        argument(
```
barthisrael, line 902: Maybe we could put this argument near `--recovery-staging-path`. They serve similar purposes. Currently they are shown like this in the console:

```console
  --recovery-staging-path RECOVERY_STAGING_PATH
                        A path to a location on the recovery host where compressed backup files will be staged during the recovery. This location must have enough available space to temporarily hold the full compressed backup. This option is *required*
                        when recovering from a compressed backup.
  --recovery-conf-filename RECOVERY_CONF_FILENAME
                        Name of the file to which recovery configuration options will be added for PostgreSQL 12 and later (default: postgresql.auto.conf).
  --snapshot-recovery-instance SNAPSHOT_RECOVERY_INSTANCE
                        Instance where the disks recovered from the snapshots are attached
  --snapshot-recovery-zone SNAPSHOT_RECOVERY_ZONE
                        Zone containing the instance and disks for the snapshot recovery (deprecated: replaced by --gcp-zone)
  --gcp-zone GCP_ZONE   Zone containing the instance and disks for the snapshot recovery
  --azure-resource-group AZURE_RESOURCE_GROUP
                        Azure resource group containing the instance and disks for recovery of a snapshot backup
  --aws-region AWS_REGION
                        The name of the AWS region containing the EC2 VM and storage volumes for recovery of a snapshot backup
  --local-staging-path LOCAL_STAGING_PATH
                        A path to a location on the local host where incremental backups will be combined during the recovery. This location must have enough available space to temporarily hold the new synthetic backup. This option is *required* when
                        recovering from an incremental backup.
```

I think it would be better if they are put near each other.

tests/test_cli.py
```diff
@@ -842,9 +844,9 @@ def test_recover_recovery_staging_path(
         mock_recover_args,
         backup_is_incremental,
         backup_is_compressed,
-        recovery_staging_path_arg,
```
barthisrael, line 845: I think we should have different tests for dealing with `recovery_staging_path` and with `local_staging_path`.

barthisrael (changes requested):

tests/testing_helpers.py
```diff
@@ -314,7 +314,7 @@ def build_config_dictionary(config_keys=None):
         "gcp_zone": None,
         "immediate_checkpoint": False,
         "incoming_wals_directory": "/some/barman/home/main/incoming",
-        "keepalive_interval": 60,
```
barthisrael, line 317: This doesn't look like an expected change.

barthisrael:
Posted some suggestions.
Note that unit tests are failing.

tests/testing_helpers.py
andremagui: this was unfortunate! My modification was exactly at the same line and i thought this was retained in the rebase! Sorry, my bad!

barthisrael: np :)

barman/cli.py
```diff
@@ -918,47 +928,58 @@ def recover(args):
...
-                    "staging path is provided. Either set recovery_staging_path "
-                    "in the Barman config or use the --recovery-staging-path "
-                    "argument.",
-                    args.backup_id,
-                    server.config.name,
-                    backup_id.compression,
-                )
+    # If the backup to be recovered is incremental then there are additional
+    # checks to be carried out
+    if backup_id.parent_backup_id is not None:
```
barthisrael, line 960:
```suggestion
    if backup_id.is_incremental:
```

We now have a property :)

barthisrael:
This works as expected.

* Errors out when the option is not set:

```console
$ barman recover postgres dev_postgres_incremental ~/data_recovery
ERROR: Cannot recover from backup 'dev_postgres_incremental' of server 'postgres': backup will be combined with pg_combinebackup in the barman host but no local staging path is provided. Either set local_staging_path in the Barman config or use the --local-staging-path argument.
```

* Succeeds when the option is set:

```console
$ barman recover postgres dev_postgres_incremental ~/data_recovery --local-staging-path /tmp/local_staging_path
Start combining backup via pg_combinebackup for backup 20240814T182116 on /tmp/local_staging_path
End combining backup via pg_combinebackup for backup 20240814T182116
Starting local restore for server postgres using backup 20240814T182116
Destination directory: /home/vagrant/data_recovery
Copying the base backup.
Copying required WAL segments.
Generating archive status files
Identify dangerous settings in destination directory.

Recovery completed (start time: 2024-08-19 12:55:40.870383+02:00, elapsed time: less than one second)
Your PostgreSQL server has been successfully prepared for recovery!
```

```console
# Checked while `pg_combinebackup` was being executed
$ ls /tmp/local_staging_path/
20240814T182116
```

barman/cli.py
barthisrael: We will address this through a separate PR as there are more occurrences of this same situation in other pieces of the code.

barthisrael (approved):
Looks good!

gcalacoci (approved):

PR #985, opened 2024-08-12
fixes and improvements on  `barman diagnose` command

gustabowill:
Opening a new PR as I got the wrong branch naming in #978

What this PR is doing:

c05da293ca531248d4f8a7ba47dbf81d78459592 fixes an issue with the barman diagnose command, where it was actually getting the Python version from the first Python interpreter found in the PATH environment variable instead of the one the Barman is actually using to run. This could lead to wrong outputs if the one found in PATH is different from the one that Barman is using. This was fixed by getting the version from the sys module instead of the python --version command which was previously.

c935a76ce380f19f4ca831caa24d0281bae6d17a adds also the Python executable path to the Barman diagnose output, as suggested by @gcalacoci

ba6956f1c99076d4a7557c666ba5285dde4c90b9 adds tests to the `get_system_info` method of the `UnixLocalCommand` as they were currently missing and is related to both previous modifications. These tests got bigger than expected for a short method because there are many cases for the `release` field of the output and i had to test them all.

References: BAR-232

### DIFF
--- a/barman/fs.py
+++ b/barman/fs.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Barman.  If not, see <http://www.gnu.org/licenses/>.
 
+import sys
 import logging
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


barthisrael (approved):
This works well:

```console
$ barman diagnose | jq '.global.system_info |{python_executable,python_ver}'
{
  "python_executable": "/home/vagrant/.pyenv/versions/3.9.19/envs/barman/bin/python",
  "python_ver": "Python 3.9.19"
}
```

Code and tests looks good.

gcalacoci (approved):

PR #986, opened 2024-08-13
Add doc about keep restriction on incremental backups

andremagui:
The use of keep annotations should be restricted to full backups only as it could lead to having orphan backups in a case where a Postgres incremental backup has its parent deleted while having the keep annotation. Such an incremental backup will then be useless for recovery.

This PR is adding a note to the documentation of the barman `keep` command

References: BAR-258

### DIFF
--- a/doc/manual/43-backup-commands.en.md
+++ b/doc/manual/43-backup-commands.en.md
@@ -76,6 +76,21 @@ the server then you can make it an archival backup with:
 barman keep <server_name> <backup_id> [--target TARGET, --status, --release]
 ```
 
+> **NOTE:**
+> To ensure the integrity of your backup system, incremental backups
+> cannot use the keep annotation in Barman. This restriction is due to
+> the way incremental backups depend on each other. Using the keep 
+> annotation with incremental backups could result in orphaned backups,
+> which means that certain backups might exist without their necessary
+> parent backups.
+>
+> In simpler terms, if you were allowed to apply the keep annotation to
+> an incremental backup, there would be a risk that parts of the backup
+> chain would be retained without their required predecessors. This 
+> situation could create backups that would be no longer useful or
+> complete, as they would be missing the essential parent backups needed
+> to restore them properly.
+
 Possible values for `TARGET` are:
 
 * `full`: The backup can always be used to recover to the latest point in


doc/manual/43-backup-commands.en.md
```diff
@@ -76,6 +76,12 @@ the server then you can make it an archival backup with:
 barman keep <server_name> <backup_id> [--target TARGET, --status, --release]
 ```
 
+> **NOTE:**
```
gcalacoci, line 79: @andremagui reword this. needs to be clearer. this is just a suggestion:
```
> `Incremental backups` are not eligible for the use of the `keep` annotation. 
> Due to the dependency that `incremental backups` have on each other, 
> the use of the `keep` annotation could lead to wrongly retaining portions of an
> incremental backups chain, de facto creating orphan incremental backups without a parent.
```
@barthisrael , thoughts?

andremagui: Could this be too much?

>Incremental backups cannot use the keep annotation. This restriction is due to the way incremental backups depend on each other. Using the keep annotation with incremental backups could result in orphaned backups, which means that certain backups might exist without their necessary parent backups.

>In simpler terms, if you apply the keep annotation to an incremental backup, there is a risk that parts of the backup chain may be retained without their required predecessors. This situation can create backups that are no longer useful or complete, as they are missing the essential parent backups needed to restore them properly.

>To ensure the integrity of your backup system, avoid using the keep annotation with incremental backups.

gcalacoci: LGTM
tagging @barthisrael for another opinion

barthisrael: I think we should change it a bit.

That explanation gives me the impression that we don't recommend doing that, but that Barman would allow it. In fact that's not true:

```console
$ barman keep postgres dev_postgres_incremental --target full
ERROR: Unable to execute the keep command on backup 20240814T182116: is an incremental backup.
Only full backups are eligible for the use of the keep command.
```

I would prefer something like this:

> Incremental backups cannot use the keep annotation. This restriction is due to the way incremental backups depend on each other. Using the keep annotation with incremental backups could result in orphaned backups, which means that certain backups might exist without their necessary parent backups.
> 
> In simpler terms, if you were allowed to apply the keep annotation to an incremental backup, there would be a risk that parts of the backup chain would be retained without their required predecessors. This situation could create backups that would be no longer useful or complete, as they would be missing the essential parent backups needed to restore them properly.
> 
> To ensure the integrity of your backup system, Barman disallow using the keep annotation with incremental backups.

What do you think about it?

barthisrael (changes requested):
Commented in the thread.

I suggested an enhancement to the existing suggestion.

barthisrael (approved):
Looks good!

gcalacoci (approved):

barthisrael:
Let's squash this, so we can merge.

PR #987, opened 2024-08-14
Add docs for the deletion of incremental backups

gustabowill:
Adds some explanation in the docs about how dependent Postgres incremental backups are going to be deleted along with their parent as they become useless for recovery in such cases.

References: BAR-251

### DIFF
--- a/doc/manual/43-backup-commands.en.md
+++ b/doc/manual/43-backup-commands.en.md
@@ -67,6 +67,11 @@ barman delete <server_name> <backup_id>
 
 The `delete` command accepts any [shortcut](#backup-id-shortcuts) to identify backups.
 
+> **IMPORTANT:**
+> If the specified backup has dependent block-level incremental backups,
+> those backups and all their dependents will also be deleted during this operation
+> as they would effectively become unusable for recovery with a missing parent in its chain.
+
 ## `keep`
 
 If you have a backup which you wish to keep beyond the retention policy of


doc/manual/43-backup-commands.en.md
```diff
@@ -67,6 +67,12 @@ barman delete <server_name> <backup_id>
 
 The `delete` command accepts any [shortcut](#backup-id-shortcuts) to identify backups.
 
+Note that if the specified backup has dependent Postgres incremental backups
+(i.e., backups taken with `--incremental <backup_to_delete>`), those backups and all their
+descendants will also be deleted during this operation, as they become unusable for recovery
+having a missing parent in its chain. This logic does not apply to Rsync incremental backups,
+which are independent on their own.
+
```
andremagui, line 75: Check the parts where there are NOTES in this doc file. It would be good to keep standards with notes, like below:

A note with capitalized letters starting with a > symbol and bold.
Example:
> **NOTE:**

Also maybe a reword like the following:
if the specified backup is a `postgres` backup and has dependent incremental backups, this backup and all their following dependents will be deleted. This happens because a backup without its parent makes the backup chain invalid and a further recovery of any of the dependent backups would not be possible.
This only applies to backups with backup_method="postgres".

My thought is that we should be less wordy and try to use the same words like dependent OR descendant (not both). Also i think that part (i.e., backups taken with `--incremental <backup_to_delete>`) is not so important and might be confusing.

The reword is just a suggestion, you do not need to follow it!

gustabowill: I agree, tbh i don't like this part either "(i.e., backups taken with --incremental <backup_to_delete>)" hahahaha.

I have two points about this:

1 - Do you think the term "postgres backup" is well-understood by the users? I mean, at the end, everything is a postgres backup, right? I've been avoiding this as I think it is mostly used by us to differentiate it from rsync backups but i'm not sure if it is the correct term.

2 - I have the impression that notes are mostly used when you're explaining something "extra", like something that didn't fit in the context but that's still important to mention. In this case, i think explaining this behavior is just a continuation of the existing explanation. What do you think?

andremagui: 1 - good point! If this is misleading to users, we might need to use backup_method="postgres" instead of just "postgres backup". Something like this!
2 - Yeah, i guess i saw the Note in the beginning of the sentence and i thought it would be better to put as a big NOTE. But i guess that explanation could be an **IMPORTANT** flag or **NOTE** because in fact it is some kind of note or important information. 

But of course, we can wait for more suggestion from the others!

gustabowill: Sure, I'll try a better version later today and ping you here to see what you think.

gcalacoci: when adding a NOTE or an IMPORTANT section to the doc, use
```
**NOTE:**
```
or 
```
**IMPORTANT:**
```

Not commenting on the content yet as seems that some changes are going to happen @gustabowill

barthisrael (changes requested):
The PR looks good, I only posted a fix for a typo.

gustabowill:
> The PR looks good, I only posted a fix for a typo.

Thanks, just updated the PR.

barthisrael (approved):
Looks good!

gcalacoci (approved):
LGTM

PR #989, opened 2024-08-15
Add docs about the new keep-alive mechanism

gustabowill:
Add in the docs a description of the new `keepalive_interval` server configuration as well as your equivalent CLI argument `--keepalive-interval` that can be specifided together with the `backup` command.

References: BAR-256

### DIFF
--- a/doc/barman.1.d/50-backup.md
+++ b/doc/barman.1.d/50-backup.md
@@ -75,6 +75,11 @@ backup *SERVER_NAME*
     :   the time, in seconds, spent waiting for the required WAL
         files to be archived before timing out
 
+    --keepalive-interval
+    :   an interval, in seconds, at which a hearbeat query will be sent to the
+        server to keep the libpq connection alive during an Rsync backup. Default
+        is 60. A value of 0 disables it.
+
     --manifest
     :   forces the creation of a backup manifest file at the end of a backup. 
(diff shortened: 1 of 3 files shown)


gcalacoci (approved):

gcalacoci:
this LGTM, approved.

barthisrael (changes requested):
The changes introduced by this PR look good.

I'm only missing some "note" in the `Backup with rsync/SSH` section providing more context about this setting. Maybe something like this:

```md
> **NOTE:**
> Starting with Barman 3.11.0, Barman uses a keep-alive mechanism when taking
> rsync-based backups. It keeps sending a simple `SELECT 1` query over the
> libpq connection where Barman runs `pg_backup_start`/`pg_backup_stop`
> low-level API functions, and it's in place to reduce the probability of a firewall or
> a router dropping that connection as it can be idle for a long time while the base
> backup is being copied. You can control the interval of the hearbeats, or even
> disable the mechanism, through the `keepalive_interval` configuration option.
```

Thoughts @gcalacoci @gustabowill ?

gustabowill:
> Thoughts @gcalacoci @gustabowill ?

Looks awesome, thanks! I added that at the end of the Rsync section.

barthisrael (approved):
Looks good!

PR #992, opened 2024-08-19
Change docs to introduce the new --incremental option

gustabowill:
Change docs to explain the new `--incremental` option that can be used together with the `backup` command. Some sections which mentioned incremental backups are being something specific to Rsync were also changed to take the new block-level incremental option into account.

References: BAR-249

### DIFF
--- a/doc/barman.1.d/50-backup.md
+++ b/doc/barman.1.d/50-backup.md
@@ -19,6 +19,11 @@ backup *SERVER_NAME*
         Overrides value of the parameter `immediate_checkpoint`, if present
         in the configuration file.
 
+    --incremental [BACKUP_ID]
+    :   performs a block-level incremental backup. A `BACKUP_ID` or [backup ID shortcut](#shortcuts)
+        of a previous backup must be provided, which references a previous backup
+        in the catalog to be used as the parent backup from which the incremental is taken.
+
     --reuse-backup [INCREMENTAL_TYPE]
     :   Overrides `reuse_backup` option behaviour. Possible values for
         `INCREMENTAL_TYPE` are:
--- a/doc/manual/50-feature-details.en.md
+++ b/doc/manual/50-feature-details.en.md
@@ -13,23 +13,20 @@ common patterns.
 
 ### Incremental backup
 
-Barman implements **file-level incremental backup**. Incremental
-backup is a type of full periodic backup which only saves data changes
-from the latest full backup available in the catalog for a specific
-PostgreSQL server. It must not be confused with differential backup,
+Incremental backup is a type of backup which uses an already existing backup as reference
+for copying only necessary data changes from the PostgreSQL server. It must not be confused with differential backup,
 which is implemented by _WAL continuous archiving_.
-
-> **NOTE:** Block level incremental backup will be available in
-> future versions.
-
-> **IMPORTANT:** The `reuse_backup` option can't be used with the
-> `postgres` backup method at this time.
 
 The main goals of incremental backups in Barman are:
 
 - Reduce the time taken for the full backup process
 - Reduce the disk space occupied by several periodic backups (**data
   deduplication**)
+
+Barman currently supports **file-level incremental backups** (using `rsync`) as well
+as **block-level incremental backups** (using `pg_basebackup`).
+
+### File-level incremental backups
 
 This feature heavily relies on `rsync` and [hard links][8], which
 must therefore be supported by both the underlying operating system
@@ -41,15 +38,15 @@ relevant savings in disk usage. This is particularly true of VLDB
 contexts and of those databases containing a high percentage of
 _read-only historical tables_.
 
-Barman implements incremental backup through a global/server option
+Rsync incremental backups can be enabled through a global/server option
 called `reuse_backup`, that transparently manages the `barman backup`
 command. It accepts three values:
 
 - `off`: standard full backup (default)
-- `link`: incremental backup, by reusing the last backup for a server
+- `link`: file-level incremental backup, by reusing the last backup for a server
   and creating a hard link of the unchanged files (for backup space
   and time reduction)
-- `copy`: incremental backup, by reusing the last backup for a server
+- `copy`: file-level incremental backup, by reusing the last backup for a server
   and creating a copy of the unchanged files (just for backup time
   reduction)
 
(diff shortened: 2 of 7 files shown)


doc/manual/50-feature-details.en.md
```diff
@@ -13,24 +13,22 @@ common patterns.
 
 ### Incremental backup
 
-Barman implements **file-level incremental backup**. Incremental
-backup is a type of full periodic backup which only saves data changes
+Incremental backup is a type of full periodic backup which only saves data changes
```
gcalacoci, line 16: mmm, I would avoid using the "full" world.. as is sorta true when you use hard links with rsync but is not true when talking of incremental.

I think that we should reword this:
```
Incremental backup is a type of full periodic backup which only saves data changes
from the latest full backup available in the catalog for a specific
PostgreSQL server.
```
 in:
```
Incremental backup is a special type of backup which uses an already existing full backup as reference
for copying only necessary data changes from the PostgreSQL server.
```
@gustabowill @barthisrael

doc/manual/50-feature-details.en.md
```diff
@@ -13,24 +13,22 @@ common patterns.
...
-> **IMPORTANT:** The `reuse_backup` option can't be used with the
-> `postgres` backup method at this time.
-
 The main goals of incremental backups in Barman are:
 
 - Reduce the time taken for the full backup process
 - Reduce the disk space occupied by several periodic backups (**data
   deduplication**)
 
+Barman currently supports **file-level incremental backups** (using `rsync`) as well
```
gcalacoci, line 27: should we move this higher? @barthisrael @gustabowill

doc/manual/50-feature-details.en.md
gustabowill: Agree! I found that a little strange too at first glance too, but this is how it is in the current docs, so I just kept it.

gustabowill: Updated this.

doc/manual/50-feature-details.en.md
gustabowill: Humm I like it here because it leaves an opening to the following sections which discuss these topics in more detail. But Idk, would like to hear more opinions as well.

barthisrael (commented):
I think it would be cool to merge the other PRs first, which are smaller, so we can rebase this on top of the new `master` branch.
That way we can build a new document and go over it as a whole.

doc/manual/50-feature-details.en.md
barthisrael: Looks good.

doc/manual/50-feature-details.en.md
barthisrael: I think it's ok in the current location.

gcalacoci (approved):
DALL·E 2024-08-20 16 40 29 - A round badge-like image with a cartoonish seal in the center  The seal should have a cute and playful appearance, with large round eyes, a smiling fa

doc/manual/50-feature-details.en.md
```diff
@@ -41,15 +38,15 @@ relevant savings in disk usage. This is particularly true of VLDB
...
 called `reuse_backup`, that transparently manages the `barman backup`
 command. It accepts three values:
 
 - `off`: standard full backup (default)
-- `link`: incremental backup, by reusing the last backup for a server
+- `link`: file-level incremental backup, by reusing the last backup for a server
   and creating a hard link of the unchanged files (for backup space
   and time reduction)
-- `copy`: incremental backup, by reusing the last backup for a server
+- `copy`: file-level incremental backup, by reusing the last backup for a server
```
barthisrael, line 49: I wonder that we can add a note after this explaining that users of `rsync` don't need to take full and incremental backups, i.e. they can take all backups with `reuse_backup = link` if they want, and the first backup in the catalog will copy all files while the 2nd onwards only the deduplicated files.

With this, I'm willing to avoid a recurrent situation where people schedule some backups with `off` and others with `link` e.g.

barthisrael (changes requested):
I like the changes introduced by this PR.
I think we can leverage of them to add a note to the rsync section.

doc/manual/50-feature-details.en.md
gustabowill: Just added a note about that at the end of the Rsync section.

doc/barman.1.d/50-backup.md
```diff
@@ -19,6 +19,11 @@ backup *SERVER_NAME*
         Overrides value of the parameter `immediate_checkpoint`, if present
         in the configuration file.
 
+    --incremental [BACKUP_ID]
+    :   performs a block-level incremental backup. A `BACKUP_ID` or [backup ID shortcut](#shortcuts)
+        of a previous backup must be provided, which references a previous backup
+        in the catalog to be used as the parent backup for deduplication.
```
martinmarques, line 25: Do we want to use the word deduplication? I would go with something like this:
```suggestion
        in the catalog to be used as the parent backup from which the incremental is taken.
```

doc/manual/50-feature-details.en.md
```diff
@@ -73,6 +70,52 @@ incremental backup as follows:
 barman backup --reuse-backup=link <server_name>
 ```
 
+> **NOTE:** Unlike Postgres block-level incremental backups, Rsync file-level incremental backups are independent
+> on their own, meaning that a backup that reused a previous backup for deduplication is not compromised
+> in any way if that backup is deleted. As mentioned, deduplication in Rsync backups is implemented with the use
```
martinmarques, line 75:
```suggestion
> in any way if the parent backup is deleted. As mentioned, deduplication in Rsync backups is implemented with the use
```

doc/manual/50-feature-details.en.md
```diff
@@ -73,6 +70,52 @@ incremental backup as follows:
...
+> will still have all files and can be reused without any concerns.
+
+> **IMPORTANT:** The `reuse_backup` option must be used along with
+> `rsync` or `local-rsync` as backup method.
+
+
+### Block-level incremental backups
+
+Since version 3.11, Barman introduces support for block-level incremental
+backups, leveraging the native [incremental backup support introduced in PostgreSQL 17](https://www.postgresql.org/docs/17/continuous-archiving.html#BACKUP-INCREMENTAL-BACKUP).
```
martinmarques, line 88: We can leave this for now, but once 17 is out and is the current version, we should change 17 for current. Let's do that as part of the next release.
This means we need a JIRA ticket so we don't forget ;)

doc/manual/50-feature-details.en.md
```diff
@@ -73,6 +70,52 @@ incremental backup as follows:
...
+barman backup --incremental <backup_id> <server_name>
+```
+
+To be able to perform a block-level incremental backup in Barman you must:
+
+- Have PostgreSQL 17 or later.
+- Have `summarize_wal` enabled.
+- Have `postgres` as your backup method.
+
+> **NOTE:** Compressed backups are not yet eligible for incremental backups in Barman.
```
martinmarques, line 113: I would highlight the "yet" there ;-)

martinmarques (changes requested):
Minor changes to perform. Overall this looks great!

doc/manual/50-feature-details.en.md
```diff
@@ -73,6 +70,52 @@ incremental backup as follows:
 barman backup --reuse-backup=link <server_name>
 ```
 
+> **NOTE:** Unlike Postgres block-level incremental backups, Rsync file-level incremental backups are independent
+> on their own, meaning that a backup that reused a previous backup for deduplication is not compromised
+> in any way if that backup is deleted. As mentioned, deduplication in Rsync backups is implemented with the use
+> of hard links, so when a previously reused backup is deleted, files shared with other backups will
+> still remain on disk, only being removed when the last backup using those files is also deleted. It also
+> means that there’s no need to take new full backups when a previous full backup is deleted, each backup
+> will still have all files and can be reused without any concerns.
```
barthisrael, line 79:
```suggestion
> still remain on disk, only being removed when the last backup using those files is also deleted. It also
> means that there’s no need to take full backups when a previous full backup is deleted, each backup
> will still have all files and can be reused without any concerns.
> Along the same lines, there is no need to ever take a full backup with rsync by using `reuse_backup = off`. If it is the first backup being taken with `reuse_backup = link`, in essence
> it behaves like `off` because there are no existing files to create hard-links on.
```

barthisrael (commented):
Overall looks good to me!
I think after applying Martin's suggestions we are good.
I posted another suggestion just expanding on your note.

barthisrael (approved):
Looks good to me!

gustabowill:
Just finished applying all suggestions.

PR #993, opened 2024-08-19
Replace checks on `parent_backup_id` with the property `is_incremental`

andremagui:
In the beginning of the implementation of the incremental backups when released with PostgreSQL 17, we introduced a `parent_backup_id` field to the BackupInfo class, and we had been using that to distinguish between incremental and non-incremental backups.

At some point we introduced a new property `is_incremental` which abstracts that logic. However, that property started being used in new code after that, but was not used in the places where there were already checks on `parent_backup_id`.

This PR is taking care of that and replacing all checks made with `parent_backup_id` for the property `is_incremental`. 
Also the unit tests were all modified according to the changes in implementation.

References: BAR-276

### DIFF
--- a/barman/infofile.py
+++ b/barman/infofile.py
@@ -903,14 +903,15 @@ class LocalBackupInfo(BackupInfo):
 
     def get_parent_backup_info(self):
         """
-        If the backup has a parent (incremental backup), build the LocalBackupInfo
-        object for the parent backup and return it.
-        if the parent backup does not exist or its status is `EMPTY`, return None.
+        If the backup is incremental, build the :class:`LocalBackupInfo` object
+        for the parent backup and return it.
+        If the backup is not incremental OR the status of the parent
+        backup is ``EMPTY``, return ``None``.
 
         :return LocalBackupInfo|None: the parent backup info object,
             or None if it does not exist or is empty.
         """
-        if self.parent_backup_id:
+        if self.is_incremental:
             backup_info = LocalBackupInfo(
                 self.server,
                 backup_id=self.parent_backup_id,
@@ -1011,7 +1012,7 @@ class LocalBackupInfo(BackupInfo):
 
             :return bool: ``True`` if it is consistent, ``False`` otherwise.
         """
-        if self.data_checksums != "on" or not self.parent_backup_id:
+        if self.data_checksums != "on" or not self.is_incremental:
             return True
         for backup in self.walk_to_root(return_self=False):
             if backup.data_checksums == "off":
@@ -1020,8 +1021,8 @@ class LocalBackupInfo(BackupInfo):
 
     def is_full_and_eligible_for_incremental(self):
         """
-        Used to filter out backups that have a parent backup id and are not
-        considered as FULL backups.
+        Check if this is a full backup taken with `postgres` method and which is
+        eligible to be a parent for an incremental backup.
 
         .. note::
             Only consider backups which are eligible for Postgres core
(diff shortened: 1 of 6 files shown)


barman/infofile.py
```diff
@@ -1020,7 +1021,7 @@ def is_checksum_consistent(self):
 
     def is_full_and_eligible_for_incremental(self):
         """
-        Used to filter out backups that have a parent backup id and are not
+        Used to filter out backups that are incremental and are not
```
gcalacoci, line 1024:
```suggestion
        Used to filter out backups that incremental backups that are not
```

gcalacoci: @andremagui i think makes more sense

gcalacoci (changes requested):
Besides a small change everything else LGTM

barman/infofile.py
```diff
@@ -903,14 +903,15 @@ def get_backup_manifest_path(self):
 
     def get_parent_backup_info(self):
         """
-        If the backup has a parent (incremental backup), build the LocalBackupInfo
-        object for the parent backup and return it.
-        if the parent backup does not exist or its status is `EMPTY`, return None.
+        If the backup is incremental, build the LocalBackupInfo object
+        for the parent backup and return it.
+        If the backup is not incremental OR the status of the parent
+        backup is `EMPTY`, return None.
```
barthisrael, line 909:
```suggestion
        If the backup is incremental, build the :class:`LocalBackupInfo` object
        for the parent backup and return it.
        If the backup is not incremental OR the status of the parent
        backup is ``EMPTY``, return ``None``.
```

barman/infofile.py
```diff
@@ -1020,7 +1021,7 @@ def is_checksum_consistent(self):
 
     def is_full_and_eligible_for_incremental(self):
         """
-        Used to filter out backups that have a parent backup id and are not
+        Used to filter out backups that are incremental and are not
         considered as FULL backups.
```
barthisrael, line 1025:
```suggestion
        Check if this is a full backup taken with `postgres` method and which is
        eligible to be a parent for an incremental backup.
```

I wonder if this makes things clearer?
I looked a few times at the current summary, and it sounds confusing.

tests/test_infofile.py
```diff
@@ -1212,15 +1221,16 @@ def test_is_checksum_consistent(
...
             assert incremental3.is_checksum_consistent() is expected
 
-    def test_true_is_full_and_eligible_for_incremental(self):
+    @patch("barman.infofile.LocalBackupInfo.is_incremental", new_callable=PropertyMock)
+    def test_true_is_full_and_eligible_for_incremental(self, mock_is_incremental):
         """
         Test that the function applies the correct conditions for a full backup
         that is eligible for incremental mode. The backup_method should be `postgres`,
-        the summarize_wal should be `on` and parent_backup_id should be `None`
+        the summarize_wal should be `on` and backup should be incremental.
```
barthisrael, line 1229:
```suggestion
        the summarize_wal should be ``on`` and backup should be incremental.
```

barthisrael (changes requested):
Looks mostly good.
Posted a few suggestions.

barman/infofile.py
andremagui: I applied the suggestion made by @barthisrael on the same part of the docstring that sounded clearer to me! I think the **must not be part** is kinda confusing in here.

barthisrael (approved):
Looks good!

barthisrael:
Let's squash before merging.

gcalacoci (approved):
Approved. Squash before merging

PR #994, opened 2024-08-19
Add docs about the behavior of --target-time with no timezone

gustabowill:
Adds a note in the docs about the behavior of `--target-time` when no timezone is specified, in which case the Barman host timezone is used instead.

References: BAR-257


barthisrael (approved):
Looks good.

PR #1000, opened 2024-08-22
Rename `Resource savings` into `Resources saved`

gcalacoci:
Resources saved is a better and clearer name


barthisrael (approved):
Looks good!

martinmarques (approved):

PR #1012, opened 2024-09-09
Check for USAGE (instead of MEMBER) privilege in all pg_has_role occurrences

RealGreenDragon:
To work correctly Barman database user should be included in some roles ad this is checked by `pg_has_role` function.
In all `pg_has_role` functions `MEMBER` privilege is checked but in Barman code no `SET ROLE` is present so it better to check in `pg_has_role` functions `USAGE` privilege instead (that includes `INHERIT` check).

`pg_has_role` function documentation: https://www.postgresql.org/docs/16/functions-info.html#id-1.5.8.32.4.4.2.2.14.1.1.1

PR already discussed with @martinmarques here: https://github.com/EnterpriseDB/repmgr/pull/807#discussion_r1750197750

### DIFF
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -610,11 +610,11 @@ class PostgreSQLConnection(PostgreSQL):
           OR
           (
             (
-              pg_has_role(CURRENT_USER, 'pg_monitor', 'MEMBER')
+              pg_has_role(CURRENT_USER, 'pg_monitor', 'USAGE')
               OR
               (
-                pg_has_role(CURRENT_USER, 'pg_read_all_settings', 'MEMBER')
-                AND pg_has_role(CURRENT_USER, 'pg_read_all_stats', 'MEMBER')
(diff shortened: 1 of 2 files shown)


martinmarques:
The code itself looks good. I would like to see a better description in the commit message, something closer to what you put in the description of the PR. Usually, if the commit message is written in a way that has a line subject and description in subsequent paragraphs, GitHub creates the PR with the user not needing to intervene in any way.

RealGreenDragon:
@martinmarques Thanks for the advice, I will do it at next PRs. I also rebased the branch as requested.

martinmarques:
@RealGreenDragon you can amend the commit messages using git rebase --interactive

RealGreenDragon:
@martinmarques: commits squashed, but I see a linter fail (python so I thinks is not for my changes...)

martinmarques:
That linker error seems like something on our side.

martinmarques:
Seems we are testing python black and that'¡s failing. But it has to do with code that is not yours.

PR #1016, opened 2024-09-18
Remove black from the barman repository

andremagui:
This PR removes the black workflow from barman repo.

A super linter workflow will be added that is going to be in charge of linter scans, `black` being one of them

References: BAR-362


martinmarques (approved):
Same as with the Sonarqube PR, amend the commit message and then merge

Issue #949, opened 2024-06-24
barman-cli-cloud does not work with IMDSv2


smcaine:
Hi,

I want to be able to ship backups to S3. In AWS we use recommended IMDSv2 (http token required) on all of our EC2 instances, we do not store AWS credentials on any instances or containers, this is not recommended, we only use IRSA to auth to other AWS services.

barman-cli-cloud only works like this if i set `http-token` as optional in my EC2 metadata config, see here for more details: 

[https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-v2-how-it-works.html](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-v2-how-it-works.html)

Please can some logic be added to fix this to work with a token required?

I think the below just needs to be passed to the `boto3.session()`  replacing ` boto3.Session(profile_name=profile_name)`:

```python
response = client.get_session_token(
    # some arbitrary ttl:
    DurationSeconds=43200,
)

token = response['Credentials']['SessionToken']
boto3.session(aws_session_token=token, )
```

thanks!

smcaine:
is anyone able to ick this one up? i created a PR for this here https://github.com/EnterpriseDB/barman/pull/1021

Thanks

martinmarques:
Hi, we will look at this before the next release.

martinmarques:
This can be doen with the AWS ASK without any need of coding in Barman

PR #1021, opened 2024-10-09
Add aws irsa auth

smcaine:
Adding this flag to use the service account iam role, for when using eks: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html

Using aws access keys is bad practice and typically prohibited in most production environments.

This relates to issue i raised https://github.com/EnterpriseDB/barman/issues/949


paul-carlton (approved):
lgtm

barman/cloud_providers/aws_s3.py
```diff
@@ -102,6 +103,7 @@ def __init__(
         self,
         url,
         encryption=None,
+        aws_irsa=False,
```
gcalacoci, line 106: please relocate this new argument at the end of the argument list to preserve the original signature @smcaine

gcalacoci:
@smcaine thanks for your contribution.

I've put a suggestion in the code during my review of the pr.

The code seems correct but your patch has broken the Unit Tests for Barman, because of this we cannot merge the PR as it is.

Could you fix the failing tests?
Would also be of help for you to add some specific Unit Tests for the code you introduced.

FYI we are working on some contribution guidelines for PRs and what I'm asking will be part of this new documentation section.

Thanks.

martinmarques:
Hi Stuart, as mentioned before, your PR is failing for tox tests. Can you check these and push a fix that makes them pass?

https://github.com/EnterpriseDB/barman/actions/runs/11720727101/job/32805402247?pr=1021

There are also flack errors that need addressing: 

https://github.com/EnterpriseDB/barman/actions/runs/11720727101/job/32805400982?pr=1021

smcaine:
Can these tests be re-run? Hopefully I've covered everything. Ive just revised PR so its clear this is just for EKS and IAM role Service accounts

PR #1028, opened 2024-10-28
Use isoformat when producing json output in cloud-show-backup

gcalacoci:
This is an hotfix for support ticket 40034

This patch modifies the `to_json()` method of the BackupInfo class substituting the `ctime` method from the datetime class with the `isoformat` method, when printing `begin_time` and `end_time` in json format.

The new format preserves all the relevant informations of the datetime object, including the timezone.

References: 40034


gcalacoci:
duplicate of #1029

PR #1029, opened 2024-10-29
Use isoformat for dates when producing json output in cloud commands

gcalacoci:
This patch modifies the `to_json()` method of the BackupInfo class
substituting the `ctime` method from the datetime class with the
`isoformat` method, when printing `begin_time` and `end_time`
in json format.

The new format preserves all the relevant informations of the datetime
object, including the timezone.

References: bar-316


barthisrael (approved):
From what I can see, the `begin_time` and `end_time` are always set with a time zone. With that in mind, I think we can simply proceed they way you are doing, i.e. replacing `ctime` with `isoformat`.

The changes look good.

As a side comment, I was wondering if we should take the opportunity to add unit tests for `barman-cloud-backup-list` with JSON output? Currently we have only for console mode.

PR #1030, opened 2024-10-29
Run isort on code to remove unused imports and blank lines

gcalacoci:
Execute isort on the whole projet cleaning up and sorting imports in both tests and code.

References: bar-429


barthisrael (approved):
The changes look good.

PR #1037, opened 2024-12-04
Add docs about running barman commands as postgres user

defilippomattia:
Docs about configuring sudo privileges for the postgres user so that it can run the get-wal command as the barman user without needing to enter a password.
 
I know the goal of Barman docs is not to teach Linux, but it could still be helpful.


martinmarques:
Thank you Mattia for the contribution. There's consensus on this PR, but I'd like to have a look at it without the pressure of having to jump on a call in the next 5 minutes 🤣 

I'll look at this next week and get back. 

I really appreciate people taking the time to amend documentation.

martinmarques:
Closing here as this was already committed through our internal process. Commit is 1c98d061313e6dda5b40944d729e68e0c3b3757a

PR #1044, opened 2024-12-18
Proof of concept fix for `barman backup --wait` on pg17

cameronmurdoch:
pg_walfile_name_offset() has been changed in pg17 to return the current segment if on a segment boundary.

Implement previous_segment_name() to calulate the previous segment and use this to restore the pre pg17 behaviour.

Tested with pg16 and pg17.

### DIFF
--- a/barman/postgres.py
+++ b/barman/postgres.py
@@ -55,6 +55,7 @@ from barman.infofile import Tablespace
 from barman.postgres_plumbing import function_name_map
 from barman.remote_status import RemoteStatusMixin
 from barman.utils import force_str, simplify_version, with_metaclass
+from barman.xlog import previous_segment_name
 
 try:
     from queue import Empty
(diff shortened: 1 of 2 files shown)


andremagui (changes requested):
The job of `current_xlog_info` is to get the `location`, `file_name`, `file_offset` and `timestamp`  by running `pg_walfile_name_offset` on the `pg_current_wal_lsn`. This must stay as it is because we do not care about the way it handles boundaries in the property itself, but we do care about this when some part of the code is calling this property, probably.

For example, for every backup a `backup_info` object is created which will retain all the information needed from a backup. In that object, we have fields like end_wal, begin_wal, etc... When `barman backup` command is run, it calls the `current_xlog_info` property to populate `begin_wal`, `end_wal`, `begin_lsn`, `end_lsn`, `begin_offset`, `end_offset`.

For `begin_wal`, we just get the current WAL. It is unlikely to start a backup at a boundary but even if it happens, we do not want `begin_wal` as the previous WAL.

So, this is one reason to move away the logic from the property.

So, as a start guide, check the link i shared in the review message that will send you to the location where stop information is created. Also, provide or refactor unit tests for this implementation and fix possible broken tests.

vnghia:
Hi I would like to ask is there any update on this one ? Thank you very much!

andremagui:
Hi @cameronmurdoch ,

Thank you for your contribution and the work you put into this PR — we really appreciate it!

As we no longer heard from you, we're going to close this PR now. We're going to incorporate this change through our internal development process.

You'll be credited appropriately once it's merged.

Thanks again for helping improve the project!

Issue #1049, opened 2024-12-30
Don't fail if the replication slots already exists during creation


antonag32:
I am setting up barman automatically through some scripts and it would be helpful to provide some idempotency when creating replication slots. I am specifically talking about the following command:

` barman receive-wal --create-slot server-name`

If the slot exists the command fails. I was thinking about adding a new flag so the command won't fail if its because the replication slot already exists.

What do you think? I can make a PR if you agree to it.

barthisrael:
Hello @antonag32,

We don't think that adding a flag to `barman recevei-wal` to ignore slot creation issues is the way to go.

A possibly better way of handling that would be having Barman to exit with different return codes depending on the issue that it faces. Right now it exits with `0` on success, or with `1` in most cases of errors. But a change in that mechanism requires a deeper analysis.

In any case, even with the current Barman code, which contains no classification of errors with different return codes, you should still be able to provide idempotency through your scripts when running `barman recevei-wal --create-slot server-name`. You could ignore a failure of `barman receive-wal --create-slot server-name` by checking that stderr contains a pattern like `Replication slot 'SLOT_NAME' already exists` when the return code is not `0`. 

Best regards,
Israel.

barthisrael:
Hello @antonag32,

After a deeper look around, we think it actually makes sense to add a new flag to `barman receive-wal` in this specific case.

As we can see in the [pg_receivewal docs](https://www.postgresql.org/docs/current/app-pgreceivewal.html), there is an option called `--if-not-exists`:

```
--if-not-exists
    Do not error out when --create-slot is specified and a slot with the specified name already exists.
```

We could create a similar option in `barman receive-wal` to call `pg_receivewal` with that flag.

> What do you think? I can make a PR if you agree to it.

Would you still be willing to contribute a PR to Barman, adding `--if-not-exists` option to `barman receive-wal`?
If not, we will take that task over within the development team.

Best regards,
Israel.

antonag32:
Hello @barthisrael 

I will be happy to contribute. I am a bit busy this week but should be able to start working on a PR next week. Thanks for taking the time to review the issue.

barthisrael:
Hey @antonag32 ,

> I will be happy to contribute. I am a bit busy this week but should be able to start working on a PR next week. Thanks for taking the time to review the issue.

Sure, thank you!

martinmarques:
@antonag32 did you have a chance to look into this?

antonag32:
[...]

PR #1065, opened 2025-02-09
Add --if-not-exists flag to receive-wal

crazybolillo:
This flag prevents `receive-wal` from exiting with an error when a slot with the same name already exists. This is the same behavior as calling `pg_receivewal` with the same flag.

References: #1049.

### DIFF
--- a/barman/server.py
+++ b/barman/server.py
@@ -2908,7 +2908,7 @@ class Server(RemoteStatusMixin):
                 "on server %s. Skipping to the next server" % self.config.name
             )
 
-    def create_physical_repslot(self):
+    def create_physical_repslot(self, ignore_duplicate=False):
         """
         Create a physical replication slot using the streaming connection
         """
@@ -2951,7 +2951,13 @@ class Server(RemoteStatusMixin):
             self.streaming.create_physical_repslot(self.config.slot_name)
             output.info("Replication slot '%s' created", self.config.slot_name)
         except PostgresDuplicateReplicationSlot:
-            output.error("Replication slot '%s' already exists", self.config.slot_name)
+            if ignore_duplicate:
+                output.info(
+                    "Replication slot '%s' already exists and --if-not-exists was specified. Skipping creation",
+                    self.config.slot_name,
+                )
+            else:
+                output.error("Replication slot '%s' already exists", self.config.slot_name)
         except PostgresReplicationSlotsFull:
             output.error(
                 "All replication slots for server '%s' are in use\n"
--- a/barman/cli.py
+++ b/barman/cli.py
@@ -1734,6 +1734,11 @@ def archive_wal(args):
             action="store_true",
         ),
         argument(
+            "--if-not-exists",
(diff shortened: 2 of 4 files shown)


crazybolillo:
I'm antonag32 but ended up doing this outside work hours so I'm using my personal account. Let me know if I have to make any changes.

barman/server.py
```diff
@@ -2852,7 +2852,11 @@ def create_physical_repslot(self):
             self.streaming.create_physical_repslot(self.config.slot_name)
             output.info("Replication slot '%s' created", self.config.slot_name)
         except PostgresDuplicateReplicationSlot:
-            output.error("Replication slot '%s' already exists", self.config.slot_name)
+            args = "Replication slot '%s' already exists", self.config.slot_name
+            if ignore_duplicate:
+                output.info(*args)
```
martinmarques, line 2857: I don't think you'll ever get to this block. If you configured create slot with `--if-not-exists` then you never get an exception of PostgresDuplicateReplicationSlot, at least I can't think of how that could happen from the top of my head.

martinmarques (changes requested):
There's missing documentation for this new option.
Not 100% sure of this, but maybe we want to think about adding this as an option to barman.conf as well,

barman/server.py
crazybolillo: I am not changing anything inside https://github.com/EnterpriseDB/barman/blob/53b00b03e917a4bc88766bf5f76063602439bb6b/barman/postgres.py#L405-L410

So the command sent to Postgres is still the same, which should result in `PostgresDuplicateReplicationSlot` still being thrown and catched by the new code. I think this is the easiest way, and it also lets us know that the replication slot existed.

crazybolillo:
Added the new flag to the RST docs.

barman/server.py
martinmarques: Sorry for not getting back earlier.

You are right. For some reason I was thinking that this patch would do something different (should have taken more time to go over the whole flow).

This is correct, but I think we might want to change the message we send when we ignore the fact that the slot exists and when we don't. Else it can lead to confusion when reading the logs quickly and not checking the detail that one is an ERROR and the other INFO.

martinmarques: Let me set this as resolved. The output message we can discuss in a separate thread

barman/server.py
```diff
@@ -2852,7 +2852,11 @@ def create_physical_repslot(self):
...
+            else:
+                output.error(*args)
```
martinmarques, line 2859: Regarding the output message and how it's being set here. I would prefer a slightly different message when the user accepts the existing slot without failing than when it doesn't.

Also, follow the same process for calling the output module as in other parts of the code, meaning, don't pass a variable where you store the message string, but rather have the string be formatted when calling `info()` or `error()` methods.

martinmarques (changes requested):

barman/server.py
crazybolillo: Went ahead and changed the message so what happened is more clear to the user. Let me know what you think.

PR #1079, opened 2025-04-16
Update pre_requisites.rst

yuki-tei:
Hi team,
Greetings. Hope you are doing well and healthy.

Prerequisite for user privilege seems to lack granting pg_checkpoint. Or you will receive this error:
[barman@edbsup04 ~]$ barman switch-wal --force --archive barman-test
ERROR: Barman switch-wal --force requires superuser rights or the 'pg_checkpoint' role

Kind Regards,
Yuki Tei


martinmarques:
There's a `note` in line 71 of that file that mentions granting `pg_checkpoint` role for checkpoints. We didn't include it in the code snippet because that is only available in Postgres 13 and above, while the SQL snippet is for all versions (not all, but at least from PG10 up)

yuki-tei:
Hi,
Apologies for overlooking. I would close this pull request.

Kind Regards,
Yuki Tei

PR #1094, opened 2025-07-03
Change the S3 connectivity test to be done with ListObjectsV2 rather than HeadBucket to allow for more fine-grained permissions

DimitriosLisenko:
In S3, there is no way to grant permissions to do the HeadBucket API call without also granting permissions to list directories at the root level of the S3 bucket.

For example, if your S3 bucket has the following structure:
- x/backups
- y/backups

and you want backups to be done to "x/backups/db_one" via an S3 user that only has permissions to read/write to "x/backups/*", it won't work because this S3 user can't HeadBucket or list the directories in the root of the bucket (so the connectivity test fails).

The directories in the root of the bucket might not be something you necessarily want to expose to S3 users that should only know about their particular subdirectory.

This PR changes the S3 connectivity test to be done via ListObjectsV2 rather than HeadBucket, which *does* work for S3 users that only have access to a particular subdirectory in the bucket without exposing what's in the root of the bucket.

I have updated the relevant tests, as well as ran manual tests in CNPG (that uses barman internally) to take backups to a subdirectory the S3 user has access to (works) and does not have access to, which fails with the following error:
```
ERROR: Barman cloud WAL archiver exception: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: User: arn:aws:iam::... is not authorized to perform: s3:ListBucket on resource: "arn:aws:s3:::..." because no identity-based policy allows the s3:ListBucket action
```

### DIFF
--- a/barman/cloud_providers/aws_s3.py
+++ b/barman/cloud_providers/aws_s3.py
@@ -213,7 +213,7 @@ class S3CloudInterface(CloudInterface):
         """
         try:
             # Search the bucket on s3
-            self.s3.meta.client.head_bucket(Bucket=self.bucket_name)
+            self.s3.meta.client.list_objects_v2(Bucket=self.bucket_name, MaxKeys=1, Prefix=self.path)
             return True
         except ClientError as exc:
             # If a client error is thrown, then check the error code.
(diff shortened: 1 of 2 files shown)


martinmarques:
This will be released later this week, and the commit from this PR will be included

PR #1109, opened 2025-08-27
Resolve Python 3.13 functools.partial warning with staticmethod

mjlshen:
Resolve Python 3.13 warning:

FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior

I ran into this as a user of CloudNative Postgres, however its also visible in CI logs within this repo, e.g. https://github.com/EnterpriseDB/barman/actions/runs/16815312469/job/47630556427#step:6:157

### DIFF
--- a/barman/cloud.py
+++ b/barman/cloud.py
@@ -175,8 +175,10 @@ class CloudTarUploader(object):
     # This is the method we use to create new buffers
     # We use named temporary files, so we can pass them by name to
     # other processes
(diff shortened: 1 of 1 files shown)


martinmarques:
We are going to provide a different fix for this, namely, something similar to what was proposed in https://github.com/EnterpriseDB/barman/pull/1143

PR #1113, opened 2025-09-03
Fix: RemoteGetWal using proper thread management

### LINKED ISSUE #1080: barman-wal-restore uses only one of several ssh connections for get-wal
Hi everyone,

we use barman as a central WAL file store for a postgres replication. 
On the standby database we use the following command as restore_command:

`/usr/bin/barman-wal-restore -p 100`

barman-wal-restore restores received WAL files in parallel, but uses only one of several parallel open ssh connections (opened by barman-wal-restore) to copy the WAL files from the barman server to the standby database server.

e.g. ssh -q -T barman@barmanhost barman get-wal ‘db’ ‘0000000100072C950000008F’

The existing barman/ sshd configuration worked with barman version 2.19, but no longer after an update to barman version 3.13.1

Best regards.

MiBUl-eu:
I created a thread base implementation for the RemoteGetWal Class. 

This will fix #1080

I tested it in our environment. 

What do you thing. Did i missed anything.

### DIFF
--- a/barman/clients/walrestore.py
+++ b/barman/clients/walrestore.py
@@ -30,6 +30,7 @@ import os
 import shutil
 import subprocess
 import sys
+import threading
 import time
 from tempfile import NamedTemporaryFile
 
@@ -430,14 +431,15 @@ def parse_arguments(args=None):
 
 
 class RemoteGetWal(object):
-    processes = set()
+    threads = set()
+    instances = set()
     """
-    The list of processes that has been spawned by RemoteGetWal
+    The list of threads and instances that have been spawned by RemoteGetWal
     """
 
     def __init__(self, config, wal_name, dest_file):
         """
-        Spawn a process that download a WAL from remote.
+        Spawn a thread that downloads a WAL from remote.
 
         If needed decompress the remote stream on the fly.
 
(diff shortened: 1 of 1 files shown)


gustabowill:
Hey @MiBUl-eu , thanks for the patch, I recognize this was a non-trivial work.

I analyzed and tested your patch and at first it looks very good. Though unfortunately we had already implemented a fix internally for this very same issue when you posted your PR. The approach we implemented uses processes instead of threads. As decompression is a cpu-bound task, it performs better with processes when decompression is done client-side (when `--keep-compression` is used).

Aside from that, sometimes I got this error when testing your patch, which might be related to not handling errors correctly:

```
  File "/var/lib/pgsql/barman-internal/venv/bin/barman-wal-restore", line 33, in <module>
    sys.exit(load_entry_point('barman', 'console_scripts', 'barman-wal-restore')())
  File "/var/lib/pgsql/barman-internal/barman/clients/walrestore.py", line 114, in main
    os.unlink(process.dest_file)
AttributeError: 'RemoteGetWal' object has no attribute 'dest_file'

```

I was also confused on the purpose of a few things, like the `_wait_for_header_bytes` which does not seem to be used anywhere in your code.

In general, the patch looks good and performs very similarly to the one we implemented. But since we already have our implementation merged and tested, we will not be moving forward with this one. Please feel free to collaborate on other tasks/issues.

Issue #1128, opened 2025-11-03
Resource leak in CloudWalUploader due to file handles not being closed


ethiebautgeorge-nasuni:
There are 2 potential resource leaks in CloudWalUploader due to file handles not being closed:
* file_object is not closed
* wal_file is not closed

When CloudWalUploader is run in a one-shot call from the CLI, the leak is not noticeable.
However, when run in sequence from a single process, if the Python gc doesn't trigger in a while, then the handles are leaked.
In our case, this leads to a handle exhaustion in our WAL uploader service.

PR: https://github.com/EnterpriseDB/barman/pull/1129

ethiebautgeorge-nasuni:
@martinmarques @barthisrael Please have a look.

barthisrael:
> [@martinmarques](https://github.com/martinmarques) [@barthisrael](https://github.com/barthisrael) Please have a look.

Thanks for the PR!

PR #1129, opened 2025-11-03
[1128] Fixed resource leak in CloudWalUploader

ethiebautgeorge-nasuni:
Fixed resource leak in CloudWalUploader in order to address https://github.com/EnterpriseDB/barman/issues/1128

All unit tests pass, except the 2 usual ones that depend on `/dev/sdf`, which I do not have on my Mac.


barthisrael:
@ethiebautgeorge-nasuni thanks for this PR!

We pulled it to our internal pipelines.
We might accept this patch and make it available in a future release.

PR #1142, opened 2025-12-19
Support alternative GCP universes

mgarstecki:
In multiple countries, local hosting companies are partnering with GCP to offer EU-sovereign GCP-like environments. Such partnerships include [one with T-Systems in Germany](https://www.t-systems.com/de/en/sovereign-cloud/solutions/sovereign-cloud-powered-by-google-cloud) or one with Thales in France: [S3NS](https://www.s3ns.io/en).

For that, Google introduced the notion of `universe` in their SDKs and CLI tools, to basically point them at non-Google infrastructure.

We're currently porting our platform to S3NS, which includes CNPG with the Barman Cloud plugin for backups, which doesn't work there right now because we can't configure the Google universe.

To connect to other universes, official clients need to be configured with a universe value, that overrides the base API domain. GOOGLE_CLOUD_UNIVERSE_DOMAIN is the official environment variable used to target non-GCP universes: https://documentation.s3ns.fr/docs/overview/tpc-key-differences#key_differences_for_developers

Unfortunately Barman uses the v1 client that doesn't read this variable, and only accepts the universe as part of the `client_options` field in the constructor, so we have to pass it in like this.

### DIFF
--- a/barman/cloud_providers/google_cloud_storage.py
+++ b/barman/cloud_providers/google_cloud_storage.py
@@ -134,7 +134,7 @@ class GoogleCloudInterface(CloudInterface):
         client_options = None
         universe_domain = os.getenv("GOOGLE_CLOUD_UNIVERSE_DOMAIN")
         if universe_domain:
-            client_options={"universe_domain": universe_domain}
+            client_options = {"universe_domain": universe_domain}
 
         self.client = storage.Client(client_options=client_options)
         self.container_client = self.client.bucket(self.bucket_name)
--- a/barman/cloud_providers/google_cloud_storage.py
+++ b/barman/cloud_providers/google_cloud_storage.py
@@ -131,7 +131,12 @@ class GoogleCloudInterface(CloudInterface):
         Creates a client using "GOOGLE_APPLICATION_CREDENTIALS" env.
         An error will be raised if the variable is missing.
         """
-        self.client = storage.Client()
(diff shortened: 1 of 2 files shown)


mgarstecki:
This is blocking us from deploying on S3NS, since without backups are DBs are nothing :slightly_smiling_face: 

Happy to discuss this further. S3NS is very new (they went GA less than two months ago) but will certainly grow in usage.

Note that the v2 Storage client reads the environment variable by default: https://github.com/googleapis/python-storage/blob/f7095faf0a81239894ff9d277849788b62eb6ac5/google/cloud/_storage_v2/services/storage/client.py#L464

Switching to it would of course fix the issue, but the fix in this PR should be enough to support S3NS without a big upgrade.

gcalacoci:
@mgarstecki, thanks for your PR.
This is interesting, I've just started a thread in our internal chat.

mgarstecki:
@martinmarques @gcalacoci hello :wave: 
I don't see anything in the release note for v3.18.0, are you still working on this ?

barthisrael:
> @martinmarques @gcalacoci hello 👋 I don't see anything in the release note for v3.18.0, are you still working on this ?

Hello @mgarstecki we weren't able to include this patch into the previous release cycle of 3.18.0.

This patch is still present in our internal pipeline, and we might make it for 3.19.0.

PR #1147, opened 2026-01-13
Add client encryption to barman cloud backup and restore

### LINKED ISSUE #1155: Add client encryption to barman cloud backup and restore
Barman cloud is used with cloudnative-pg to backup the managed engines. Backups are sent unencrypted to an s3 endpoint.

When using insecure ( public ) s3 endpoints, it is necessary to encrypt client side.

Options :
* client-side encryption in boto3 - not available but an issue exists
* an encrypting s3 proxy : exists, but is unmaintained. There's also a commercial one but it is very complex and pricy.

pvbiesen-vaph:
Adds client encryption for barman cloud backup and restore.

At this point, only XChaCha20-poly1305 is implemented. Key  rotation is possible and implemented.

Fixes #1155

### DIFF
--- a/.github/workflows/github-to-jira.yml
+++ b/.github/workflows/github-to-jira.yml
@@ -1,0 +1,24 @@
+name: Automation for GitHub -> Jira
+
+on:
+  issues:
+    types:
+      - labeled
+
+jobs:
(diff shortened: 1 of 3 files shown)


martinmarques:
Could you please create an issue in Github that tracks this PR?

Issue #1115, opened 2025-09-09
Barman to barman sync failed for incremental backups


adamw-linadm:
I have two barman servers. One makes a backup from Postgres, the other downloads the backup from barman to itself as an additional backup.
I recently deployed PostgreSQL 17.5 and checked the operation of incremental backups. Everything works great on the backup side directly from postgres.
However, the second barman server, which downloads backups to another location, cannot handle incremental backups.

Backups on first barman server:
```
[barman@backup ~]$ barman list-backup test
test 20250909T085055 - I - Tue Sep  9 08:51:33 2025 - Size: 171.3 MiB - WAL Size: 960.0 MiB
test 20250909T022647 - F - Tue Sep  9 03:26:10 2025 - Size: 291.4 GiB - WAL Size: 160.0 MiB
test 20250908T022857 - F - Mon Sep  8 03:29:12 2025 - Size: 291.3 GiB - WAL Size: 4.8 GiB
test 20250907T022601 - F - Sun Sep  7 03:24:41 2025 - Size: 291.3 GiB - WAL Size: 800.0 MiB
```
Backup on secondary barman server:
```
[barman@ovh-backup1 ~]$ barman list-backups test
test 20250909T022647 - F - Tue Sep  9 03:26:10 2025 - Size: 291.4 GiB - WAL Size: 960.0 MiB
test 20250908T022857 - F - Mon Sep  8 03:29:12 2025 - Size: 291.3 GiB - WAL Size: 4.3 GiB
test 20250907T022601 - F - Sun Sep  7 03:24:41 2025 - Size: 291.3 GiB - WAL Size: 800.0 MiB
test 20250906T022346 - F - Sat Sep  6 03:23:49 2025 - Size: 291.3 GiB - WAL Size: 576.0 MiB
```

And error on secondary server:
```
2025-09-09 10:06:37,847 [3780339] barman.cli ERROR: Unable to run cron on server 'test', please look in the barman log file for more details.
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/barman/cli.py", line 357, in cron
    server.cron(keep_descriptors=args.keep_descriptors)
  File "/usr/lib/python3.12/site-packages/barman/server.py", line 2884, in cron
    self.sync_cron(keep_descriptors)
  File "/usr/lib/python3.12/site-packages/barman/server.py", line 4001, in sync_cron
    self.check_sync_required(backup_id, remote_info, local_backup_info)
[...]

PR #1159, opened 2026-02-15
fix: barman-to-barman sync for incremental backups (#1115)

jonatas-rp:
Fix barman-to-barman sync for incremental backups #1115 

When syncing incremental backups from a primary barman server to a passive barman server, a KeyError was raised in is_backup_locally_obsolete because the parent backup info did not contain the reference to the children_backup_ids and the field was not loaded correctly when building a BackupInfo object.

Changes:
- Added sync_parent_backup_info method to Server class
- Call sync_parent_backup_info in sync_cron before check_sync_required
- Fixed children_backup_ids deserialization in BackupInfo.from_json
- Added tests for sync_parent_backup_info


mike-broughton:
I think this change does not deal with the situation where you have synced backups that are beyond the retention period on the primary? So even when this change gets merged, the only way to cleanup an existing barman backup server is to rerun the sync from scratch?

barthisrael:
> I think this change does not deal with the situation where you have synced backups that are beyond the retention period on the primary? So even when this change gets merged, the only way to cleanup an existing barman backup server is to rerun the sync from scratch?

@mike-broughton the code on `master` has an issue with synching block-level incremental backups: whenever `barman cron` runs on the passive node and attempts to synchronize a block-level incremental backup, it faces a `KeyError` exception because the children metadata of the parent backup was not being handled correctly by that code. As a consequence, you would never get block-level incremental backups in the passive node.

The fix will allow incremental backups to be synched from now own.

Could you please expand on what are your concerns? Is that related with block-level incremental backups?

barthisrael:
Thanks @jonatas-rp for this patch.

It has been included in the release 3.20.0 as you can see in this commit: https://github.com/EnterpriseDB/barman/commit/d6f8eff8519b40fbeee451bd683fdbc6e6f6b611

PR #1161, opened 2026-02-21
fix(cloud): emit fixed-size multipart upload parts for R2 compatibility

### LINKED ISSUE #954: Backups to Cloudflare R2 (S3 Compatible) storage fail
I'm backing up to S3 compatible storage (Cloudflare R2) from Cloud Native PG (CNPG) which uses barman-cloud-backup.

Backups are failing with an error of:

`ERROR: Upload error: An error occurred (InvalidPart) when calling the CompleteMultipartUpload operation: All non-trailing parts must have the same length. (worker 1)`

The above error is surfaced from barman-cloud-backup when attempting to complete the weekly full backup.

```
$ barman-cloud-backup --version
barman-cloud-backup 3.10.1
```

I've come across what seems like a similar issue when using Cloudflare R2 as a backend for a Docker registry that was resolved by setting a fixed chunk size for multipart uploads, however I don't know enough about the barman codebase to know if this is the same issue.

https://community.cloudflare.com/t/all-non-trailing-parts-must-have-the-same-length/552190/7

And this discussion here:

https://github.com/distribution/distribution/pull/3940#issuecomment-1638356120

alleneubank:
## Summary

- Refactor `CloudTarUploader` to fill each disk-backed temp file to exactly `chunk_size` bytes before uploading, so every non-trailing multipart part is uniform in size
- Cloudflare R2 requires all non-trailing parts to have identical sizes; the previous implementation flushed whenever the buffer exceeded `chunk_size`, producing variable-sized parts
- The trailing part (uploaded via `flush`/`close`) may be smaller, which is allowed by the multipart upload spec
- Backward compatible with AWS S3 and other providers

## Changes

- `barman/cloud.py`: Extract `_write_data()` (fills fixed-size part buffers) and `_upload_part_buffer()` (uploads a single part) from the existing `write()`/`flush()` methods
- `tests/test_cloud.py`: Add two new tests verifying partial-buffer behavior and fixed-size part emission; fix existing throttle test to set up buffer state correctly

## Test plan

- [ ] `test_write_uses_disk_buffer_for_partial_data` — partial writes stay in buffer, no upload triggered
- [ ] `test_write_emits_fixed_size_parts_and_retains_tail` — 10 bytes at chunk_size=4 produces two 4-byte parts + 2-byte tail
- [ ] Existing `TestCloudTarUploader` tests continue to pass

Fixes: EnterpriseDB/barman#954

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+# Keep image build contexts lean; setup.py needs barman/ plus the prebuilt
+# man pages under docs/_build/man (data_files), so docs stays in context.
+.git
+.github
+.venv
+.pytest_cache
+build
+tests
+barman.egg-info
--- a/.github/workflows/cnpg-image.yml
+++ b/.github/workflows/cnpg-image.yml
@@ -1,0 +1,45 @@
+---
+name: CNPG operand image
+
+# Publishes the CloudNativePG postgres operand image with the forked barman
+# overlaid (R2 multipart fix). Consumed by misfits-postgres clusters via
+# cluster.imageName in the canton monorepo.
+on:
+  push:
+    branches:
+      - bb/fix-r2-multipart-uniform-parts
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-push:
+    name: Build and push ghcr.io/0xsend/cnpg-postgresql
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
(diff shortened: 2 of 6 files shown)


martinmarques:
Hello, we had an initial look at the PR, but couldn't get to it due to conflicting priorities. We want to have a closer look at the proposal. We'll get back to you here after the next release is over.

linwalth:
is this still on the table?

PR #1172, opened 2026-04-30
Add support for S3 storage classes

### LINKED ISSUE #1104: Add support for S3 Storage Classes
Hi everyone,

I'm creating this issue as per the conversation on the mailing list: https://groups.google.com/g/pgbarman/c/ipyNJgOZk5g/m/2W-UPHvWBwAJ?utm_medium=email&utm_source=footer

## What

By default, backups are uploaded to S3 Standard StorageClass. Infrequent Access StorageClass would be better for us, as it would allow us to save half of the storage costs ($0.023 per GB vs. $0.0125 per GB). Unfortunately, it is not currently supported.

There was a PR attempt here, but it was never completed: https://github.com/EnterpriseDB/barman/pull/338

NOTE: We use `barman` through CloudNativePG, in case this is relevant information.

## Why

This is an important cost optimization for us.

## Questions

- Is this a feature the `barman` team is willing to prioritize? If so, on what timeline could we expect a new release of Barman that might include this feature? I'm asking so I can manage expectations internally.
- The ParadeDB team (where I work) might be able to help get this feature built, but we don't have the expertise with barman and would need to understand the release cadence to decide how to prioritize this.

Thank you!

jonatas-rp:
Closes #1104 

Changes:

- Add --storage-class to barman-cloud-backup, --storage-class-wals to barman-cloud-wal-archive, and aws_storage_class aws_storage_class_wals server config keys, all forwarded to S3 uploads as the StorageClass header.
- Raises ObjectIsArchived exception when files are archived with explicit message that hints users to issue an RestoreObject request
- Adds tests to validate configuration variables are correctly being forwarded
- Included documentation for the new added config keys and warnings related to archived objects


martinmarques:
Fixes #1104

PR #1174, opened 2026-05-03
Add missing .zst extension for ZSTD compression in _build_dest_name

hobostay:
## Summary
- The `_build_dest_name` method in `CloudUploadController` (barman/cloud.py) handles compression extensions for `gz`, `bz2`, `snappy`, and `lz4`, but was missing `zstd`
- When ZSTD compression is used, the resulting tar archives are created without the `.zst` extension, which can confuse downstream decompression logic that relies on file extensions
- The `ALLOWED_COMPRESSIONS` dict at line 86 already maps `.zst` to `zstd`, confirming this extension should be used

## Bug details
```python
# Before (missing zstd):
elif self.compression == "lz4":
    components.append(".lz4")
return "".join(components)

# After (fixed):
elif self.compression == "lz4":
    components.append(".lz4")
elif self.compression == "zstd":
    components.append(".zst")
return "".join(components)
```

## Test plan
- [ ] Confirm ZSTD-compressed archives get the `.zst` extension

### DIFF
--- a/barman/cloud.py
+++ b/barman/cloud.py
@@ -414,6 +414,8 @@ class CloudUploadController(object):
             components.append(".snappy")
         elif self.compression == "lz4":
             components.append(".lz4")
+        elif self.compression == "zstd":
+            components.append(".zst")
         return "".join(components)
 
     def _get_tar(self, name):


martinmarques:
zstd is not supported for cloud backups. It can only be used with WAL archiving.

I would really encourage you not to open unverified PRs written by Claude or any other AI tool.

PR #1178, opened 2026-05-03
Fix negative file_count when streaming WAL directory is empty

hobostay:
## Summary
- In `_check_wal_queue` (barman/server.py), `file_count` is decremented by 1 when `dir_name == "streaming"` to account for the `.partial` file
- If the streaming directory is empty (0 files), `file_count` becomes `-1`, which can lead to incorrect comparison results
- The `file_count > 0` check at line 838 will correctly skip for `-1`, but `file_count > max_incoming_wal` at line 854 could produce confusing negative values in error messages

## Bug details
```python
# Before (buggy) - can go negative:
if dir_name == "streaming":
    file_count -= 1  # -1 when directory is empty!

# After (fixed) - guard against negative:
if dir_name == "streaming" and file_count > 0:
    file_count -= 1
```

## Test plan
- [ ] Confirm that an empty streaming directory results in `file_count == 0` instead of `-1`

### DIFF
--- a/barman/server.py
+++ b/barman/server.py
@@ -826,7 +826,7 @@ class Server(RemoteStatusMixin):
 
         # Subtract one from the count because of .partial file inside the
         # streaming directory
-        if dir_name == "streaming":
+        if dir_name == "streaming" and file_count > 0:
             file_count -= 1
 
         # If this archiver is disabled, check the number of files in the


martinmarques:
This PR adds zero value. The fact that file_count is negative doesn't affect the check process at all. I do see a different problem here, which is that we assume that the server is streaming the WALs, which might not be the case.

For the future, please start a discussion or open an issue asking if this is correct. Also, don't vibe-code. Check context and verify that the code that Claude delivers is correct and aligns with the architecture. You should also read the commit that added this code piece, and what the method accomplishes.

If you had followed these advises, you wouldn't have opened this PR.

PR #1179, opened 2026-05-03
Fix UnboundLocalError for wal_info in check_wal_validity

hobostay:
## Summary
- In `check_wal_validity` (barman/server.py), `wal_info` is only assigned inside the `if backup_info is not None:` block (line 1495), but is referenced at lines 1499 and 1515
- When `backup_info is None` and `self.config.last_wal_maximum_age is None`, the code at line 1515 accesses `wal_info["wal_until_next_size"]` which raises `UnboundLocalError`
- Similarly, when `backup_info is None` and `last_wal_maximum_age is not None`, line 1499 accesses `wal_info["wal_last_timestamp"]` which would also fail

## Bug details
```python
# Before (buggy):
backup_info = self.get_backup(backup_id)
if backup_info is not None:
    wal_info = self.get_wal_info(backup_info)  # Only assigned here!
# ...
else:
    if backup_info is None or wal_info["wal_until_next_size"] is None:  # UnboundLocalError!

# After (fixed):
wal_info = None
backup_info = self.get_backup(backup_id)
if backup_info is not None:
    wal_info = self.get_wal_info(backup_info)
# ...
else:
    if backup_info is None or wal_info is None or wal_info["wal_until_next_size"] is None:
```

## Test plan
- [ ] Confirm no `UnboundLocalError` when `backup_info` is `None`

### DIFF
--- a/barman/server.py
+++ b/barman/server.py
@@ -1491,12 +1491,13 @@ class Server(RemoteStatusMixin):
         check_strategy.init_check("wal maximum age")
         backup_id = self.backup_manager.get_last_backup_id()
         backup_info = self.get_backup(backup_id)
+        wal_info = None
         if backup_info is not None:
             wal_info = self.get_wal_info(backup_info)
         # first check: check wal maximum age
         if self.config.last_wal_maximum_age is not None:
             # get maximum age information
-            if backup_info is None or wal_info["wal_last_timestamp"] is None:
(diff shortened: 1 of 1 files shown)


martinmarques:
This is another incorrect PR. There's a short-circuit there, so if `backup_info` is None, the condition on the right of the `or` is not evaluated.

PR #1180, opened 2026-05-03
Fix file handle leak on exception in RemoteGetWal

hobostay:
## Summary
- In `RemoteGetWal.__init__` (barman/clients/walrestore.py), when `dest_file` is passed as a string, a file handle is opened at line 476
- If any subsequent operation (compression identification, decompression) raises an exception before `dest_file.close()` at line 502, the file handle is leaked
- Wrapped the operations in a `try/finally` block to ensure the file is always closed when it was opened from a string path

## Bug details
```python
# Before (buggy) - no cleanup on exception:
dest_file = open(dest_file, "wb+")
# ... operations that can raise ...
dest_file.close()  # Never reached if exception occurs above

# After (fixed) - always close on exception:
dest_file = open(dest_file, "wb+")
try:
    # ... operations that can raise ...
finally:
    if self.dest_file is not None:
        dest_file.close()
```

The `if self.dest_file is not None` check ensures we only close the file when we opened it ourselves (i.e., when it was originally a string path).

## Test plan
- [ ] Confirm file handle is properly closed even when compression operations raise exceptions

### DIFF
--- a/barman/clients/walrestore.py
+++ b/barman/clients/walrestore.py
@@ -475,31 +475,33 @@ class RemoteGetWal(object):
             self.dest_file = dest_file
             dest_file = open(dest_file, "wb+")
 
-        # Spawn a remote get-wal process and direct its output to the destination file
-        self.ssh_process = subprocess.Popen(
-            build_ssh_command(config, wal_name), stdout=dest_file
-        )
-        self.ssh_process.wait()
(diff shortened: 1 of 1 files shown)


martinmarques:
This PR makes a bit more sense, but I see other unaddressed issues. I'll check this internally today.

PR #1190, opened 2026-05-22
Fix cloud-wal-restore on compressed WAL + sibling backup label

adam8157:
- `CloudWalDownloader._get_wals_to_download` matched the requested WAL by `path.startswith(requested_wal_path)`, which also matched the sibling `<wal>.<offset>.backup` label file. Combined with the early-break introduced in 5625881a, recovery against a compressed backup would exit the iteration on the `.backup.gz` entry (which sorts before `.gz` lexicographically) and never reach the real WAL — `FATAL: could not locate required checkpoint record`.
- Match by basename equality after stripping any compression suffix; accept both `<wal>` and `<wal>.partial`. Sibling `.backup` files are now harmlessly skipped via `_validate_wal_path`. The "requested WAL missing → return `[]`" invariant from 5625881a is preserved via a `found_requested` flag enforced after the loop.
- Adds a regression test covering the compressed-WAL + sibling-backup-label case.

References: BAR-1305.

### DIFF
--- a/barman/cloud.py
+++ b/barman/cloud.py
@@ -3002,20 +3002,44 @@ class CloudWalDownloader(object):
 
         wals_to_download = []
         count = 0
+        found_requested = False
         for path in sorted(self.cloud_interface.list_bucket(prefix)):
-            is_requested_wal = path.startswith(requested_wal_path)
+            # Match the requested WAL by basename (after stripping any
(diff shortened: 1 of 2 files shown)


adam8157:
Somehow these two commits are already merged, closing

PR #1194, opened 2026-05-29
Exit with network error code on cloud-wal-restore connectivity loss

### LINKED ISSUE #1193: `barman-cloud-wal-restore`: exit with code 2 on network failure
`barman-cloud-wal-restore` always exits with the generic code 4 (GeneralErrorExit) when a WAL download fails, regardless of cause.

This prevents callers (e.g. the Barman Cloud plugin in CloudNativePG) from distinguishing a transient connectivity loss from a genuine failure. `barman-cloud-check-wal-archive` already makes this distinction, exiting with code 2 (`NetworkErrorExit`) when the provider is unreachable. `barman-cloud-wal-restore` should be consistent: if a download fails because connectivity was lost, exit with 2; otherwise keep exiting with 4.

This lets automation apply different retry policies for network issues versus permanent errors.

gbartolini:
## Description

When a WAL download fails, `barman-cloud-wal-restore` always exited with the generic code `4` (`GeneralErrorExit`), regardless of the cause. This prevented callers such as PostgreSQL's `restore_command` from telling a transient loss of connectivity apart from a genuine failure.

This change classifies the **original** exception rather than re-probing connectivity after the fact:

- Adds an `is_connectivity_error(exc)` method to `CloudInterface`, implemented by each provider in terms of the connectivity exceptions it already recognises:
  - AWS: `EndpointConnectionError`
  - Azure: `HttpResponseError`, `ServiceRequestError`
  - GCS: `GoogleAPIError`
- Refactors each provider's existing `test_connectivity()` to reuse the same predicate (DRY).
- In the `cloud-wal-restore` error path, if the original exception is a connectivity error it exits with code `2` (`NetworkErrorExit`); otherwise it keeps exiting with code `4`.

This brings `barman-cloud-wal-restore` in line with `barman-cloud-check-wal-archive`, which already distinguishes network failures, and lets automation apply different retry policies for transient network issues versus permanent errors.

## Testing

- `test_fails_on_network_error_during_download`: a connectivity error exits with code `2`.
- `test_fails_on_download_exception`: a non-connectivity failure exits with code `4`.
- Existing `tests/test_cloud.py` provider tests continue to pass.

Closes #1193

### DIFF
--- a/barman/clients/cloud_walrestore.py
+++ b/barman/clients/cloud_walrestore.py
@@ -22,6 +22,7 @@ from contextlib import closing
 from barman.clients.cloud_cli import (
     CLIErrorExit,
     GeneralErrorExit,
+    NetworkErrorExit,
     create_argument_parser,
 )
 from barman.cloud import CloudWalDownloader, configure_logging
@@ -47,6 +48,7 @@ def main(args=None):
         _logger.error("%s is an invalid name for a WAL file" % config.wal_name)
         raise CLIErrorExit()
 
+    cloud_interface = None
     try:
         cloud_interface = get_cloud_interface(config)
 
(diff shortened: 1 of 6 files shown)


martinmarques:
4f43dc59df5e598de002fea8c8aaa393e68c4dd6

Issue #1101, opened 2025-07-27
Reducing expensive S3 operations


CppBunny:
Currently, barman calls HeadBucket on every write to test connectivity. For a lot of S3 providers, HeadBucket is among the most expensive operations you can perform. For a busy database with WAL archiving enabled, S3 costs can skyrocket just because of the connectivity check.

Would be useful to look into alternatives or provide a way to disable the connectivity check.

martinmarques:
We have a change that is scheduled for release next week, which removes calls to Headbucket, but replacing it with list_objects_v2(). I'm not sure if that will eliminate the overhead.

Please read https://github.com/EnterpriseDB/barman/issues/929 for more information.

martinmarques:
We found that the work we had done was going to exacerbate the expenses, so we have reverted it. We will analyze how to reduce the checks we do with the bucket

CppBunny:
> We found that the work we had done was going to exacerbate the expenses, so we have reverted it. We will analyze how to reduce the checks we do with the bucket

The check itself is not necessarily bad, especially for one-time operations. An idea would be to have a flag that let's you skip the connectivity check (granted the bucket exists) and potentially provide a separate cli command just for verifying connectivity. This way, applications or scripts that integrate Barman can call the connectivity check once at startup or before a looping operation. Feel free to let me know if any help is needed, definitely willing to contribute.

alexklibisz:
+1 for a command line argument or some other configuration to skip this connection check. I don't understand what benefit it adds. If it fails, what do we do differently based on that result? Why not just fail the object upload?

martinmarques:
The latest release of Barman 3.16 has removed the head bucket check from all non-check-related commands. This means that we no longer do a head bucket check on WAL archive and backup runs

> +1 for a command line argument or some other configuration to skip this connection check. I don't understand what benefit it adds. If it fails, what do we do differently based on that result? Why not just fail the object upload?

I -1 this ☝️  because we don't want more command-line arguments. As you mentioned, there wasn't much benefit in having that check execute over and over for each command. 

The reason for it being there was that these kinds of things were more common when it was initially developed, but times have changed.

Hopefully everyone will be happy with this 😄

alexklibisz:
Thanks @martinmarques !

Just to tie this up for others following the issue: 

Looks like the release is here: https://github.com/EnterpriseDB/barman/releases/tag/release%2F3.16.0

Relevant release note:

[...]

Issue #929, opened 2024-06-06
Utilize aws s3 ls instead of s3api head-bucket to check for access before barman-cloud-backup execution


benjamin-schilling-csq:
barman-cloud-backup currently uses the head-bucket operation of the s3api to check for existence and access to the targeted S3 bucket. For those trying to write restricted policy to prefixes within the bucket, the only thing that allows head-bucket to properly work is to allow ListBucket to the entirety of the bucket. If using the equivalent of aws-cli s3 ls, the policy can allow for checking that the root of the bucket exists while also restricting prefixes levels below the root of the bucket, permitting for a more secure solution. 

If this is a feasible feature request it would be greatly appreciated if it could be implemented.

mwijngaard:
I also just ran into this issue as I restrict s3:ListBucket to prefixes. Since the current implementation merely uses HeadBucket to check for bucket existence, the same could also be achieved by using ListObjectsV2 with a `prefix` parameter. This returns a 404 for non-existent buckets, but is compatible with prefix-restricted s3:ListBucket permissions. This seems like a simple method replace in `_check_bucket_existence` without any downside, or is there something I'm missing @martinmarques ?

mwijngaard:
@andremagui I would probably opt against regarding AccessDenied as a 'bucket exists' and just require the bucket to be listable. Since this is also the current requirement (head-bucket requires ListBucket without prefix, which also always allows ListObjectsV2), that would be backwards compatible, and it just makes more sense to me. 

Also, it seems to me that an explicit 'deny' would not be necessary in any case, since deny is the default? Merely having an allow role with prefix condition would be sufficient?

andremagui:
Hi @mwijngaard, we're currently discussing this internally and will get back to you by early next week at the latest.

DimitriosLisenko:
Hi all! I've implemented this change because we need it internally - you can see the update in https://github.com/EnterpriseDB/barman/pull/1094. Happy to make any changes that might be necessary for this to be merged.

martinmarques:
Has anyone looked at the replies to this issue from Terraform?

https://github.com/hashicorp/terraform-provider-aws/issues/17195

mwijngaard:
@martinmarques which part exactly? The comment mentioning the null condition? Like the last comment says, that would indeed allow listing the full bucket, which is the opposite of what we're trying to achieve.

mwijngaard:
[...]

PR #1198, opened 2026-07-03
feat: add flag to disable bucket-level actions (e.g. HeadBucket) when checking WAL archive

AlessandroZanatta:
## What?
I've added a simple CLI flag for the `barman-cloud-check-wal-archive` command that allows disabling the bucket-level actions (e.g. HeadBucket and CreateBucket) that require bucket-level permissions. This change would also be a first step towards fixing #929, which was previously fixed and then reverted due to #1101 considerations.

## Why?
When dealing with strict security requirements, or multi-tenant buckets, this flag can be used to have the aforementioned command skip the `test_connectivity` and `setup_bucket` calls, which would require bucket-level access from the executing principal (e.g. unprefixed ListBucket). Removing these actions would allow plugins, such as CNPG barman cloud plugin, to skip such checks, while retaining safety (check if bucket path is empty) and security (allowing only prefixed ListBucket calls).

## How?
I think the implementation is straightforward. However, I would point out that specifying `--test` and `--assume-bucket-exists` together would now result in `--test` being a no-op. I personally think this is fine since `--assume-bucket-exists` is disabled by default, thus making the change non-breaking with existing users of this project.

## Testing
I've added a couple of test cases that should cover the new branches I've created. The first ensures that, when the flag is enabled, that no bucket-level operation is called. The latter verifies the behavior when `-t/--test` and `--assume-bucket-exists` are both passed.

## Docs
I've also added the CLI flag to the docs. Please let me know if more work shall be done there.

### DIFF
--- a/barman/clients/cloud_check_wal_archive.py
+++ b/barman/clients/cloud_check_wal_archive.py
@@ -47,15 +47,20 @@ def main(args=None):
 
     try:
         cloud_interface = get_cloud_interface(config)
-        if not cloud_interface.test_connectivity():
-            # Deliberately raise an error if we cannot connect
-            raise NetworkErrorExit()
-        # If test is requested, just exit after connectivity test
+        if not config.assume_bucket_exists:
+            if not cloud_interface.test_connectivity():
+                # Deliberately raise an error if we cannot connect
+                raise NetworkErrorExit()
+            # If test is requested, just exit after connectivity test
+            elif config.test:
+                raise SystemExit(0)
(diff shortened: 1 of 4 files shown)


AlessandroZanatta:
@barthisrael @joao00001 @gustabowill
Could you please have a look at this? 
Thanks!

AlessandroZanatta:
@barthisrael @joao00001 @gustabowill
Could you please have a look at this?
Thanks!

PR #1209, opened 2026-07-28
Fix TypeError on snapshot restore: forward copy_partial through SnapshotRecoveryExecutor

eScrappy:
## Problem

Every snapshot restore fails with:

```
TypeError: recover() got an unexpected keyword argument 'copy_partial'
```

`SnapshotRecoveryExecutor.recover()` overrides the base `RecoveryExecutor.recover()`, but its signature was never updated when `copy_partial` was added to the base method. Because `restore` always passes `copy_partial` down the call chain (`cli.py` → `server.recover` → `backup_manager.recover` → `executor.recover`, all via `**kwargs`), the argument lands on the snapshot executor, which doesn't accept it — raising the `TypeError` for any backup whose `snapshots_info` is set.

The non-snapshot path (`MainRecoveryExecutor`) is unaffected because it inherits the base `recover()` unchanged.

## Fix

- Add `copy_partial=False` to `SnapshotRecoveryExecutor.recover()`'s signature (plus docstring).
- Forward `copy_partial=copy_partial` to `super().recover(...)`. Without this, even after accepting the kwarg the `.partial` WAL handling would be silently dropped for snapshot recoveries.

## Tests

- Added `test_recover_copy_partial_forwarded` asserting a truthy `copy_partial` reaches `super().recover()`.
- Updated `test_recover_success`'s superclass-call assertion to include the now-forwarded `copy_partial=False`.

Full `tests/test_recovery_executor.py` suite passes (196 passed).

### DIFF
--- a/src/barman/recovery_executor.py
+++ b/src/barman/recovery_executor.py
@@ -1893,6 +1893,7 @@ class SnapshotRecoveryExecutor(RemoteConfigRecoveryExecutor):
         recovery_conf_filename=None,
         recovery_option_port=None,
         custom_restore_command=None,
+        copy_partial=False,
         recovery_instance=None,
     ):
         """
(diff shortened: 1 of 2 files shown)


eScrappy:
Closing this one: the fix is being landed through EDB's internal 3.x branch instead, and will reach the community repo from there.

For anyone who lands here from the traceback: `barman restore` from a snapshot backup fails with `TypeError: recover() got an unexpected keyword argument 'copy_partial'` because `SnapshotRecoveryExecutor.recover()` overrides the base `RecoveryExecutor.recover()` but was never updated when `copy_partial` was added to it. The fix is to accept `copy_partial` in the override's signature and forward it to `super().recover(...)` — forwarding matters, otherwise `.partial` WAL handling is silently dropped for snapshot recoveries.

PR #1215, opened 2026-08-31
docs: fix the barman-recover man page

mnencia:
The `barman-recover` entry was missing from `man_pages` in `docs/conf.py`, so the shipped man page was stale and out of sync with the actual `recover` command. Add the missing entry and regenerate the page.


barthisrael:
Thanks, Marco!

I pulled this PR to our internal repository.
This patch will come as part of version 3.20.1

PR #1216, opened 2026-08-31
chore: drop executable bit from client modules

mnencia:
These modules are installed as console scripts via setuptools entry points and are never invoked directly, so the executable bit serves no purpose.


barthisrael:
Thanks, Marco!

I pulled this PR to our internal repository.
This patch will come as part of version 3.20.1

PR #1219, opened 2026-09-03
Warn instead of raising when a WAL directory is already gone

### LINKED ISSUE #1213: "barman delete" fails if Wals-directory does not exist
On
`Debian GNU/Linux 12 (bookworm)`

`$ barman -v`
3.19.1 Barman by EnterpriseDB (www.enterprisedb.com)

### barman fails to delete a backup, if an expected Wals-directory doesn't exist

`$ barman delete pg 20260720T001356`
`Deleting backup 20260720T001356 for server pg`
`Delete associated WAL segments:`
`EXCEPTION: [Errno 2] No such file or directory: '/opt/barman/backups/pg/wals/0000000900003E12'`
`See log file for more details.`

My quick solution: repeat execution unitl backup is deleted of
`$ barman delete pg 20260720T001356 > /tmp/xxx 2>&1 || grep EXCEPTION /tmp/xxx | awk -F ':' '{print $3}' | awk -F "'" '{print $2}' >> /tmp/yyy && for i in `cat /tmp/yyy`; do mkdir $i; chown barman:barman $i; echo $i ; done`

Suggested solution: Ignoring non existing Wals Directories for `backup delete`

Thank you

rawsun007:
Closes #1213

## The cause

The traceback in #1213 names a *directory*, not a file:

```
EXCEPTION: [Errno 2] No such file or directory: /opt/barman/backups/pg/wals/0000000900003E12
```

That comes from `os.listdir(wal_dir)` in `LocalWalStorageStrategy.delete()`. The listing happens before the choice between removing the hash directory wholesale and unlinking files one by one, so if the directory has been removed outside Barman the delete aborts there — after the WALs are dropped from the xlogdb but before `delete_basebackup()` and `delete_backupinfo_file()`. The backup is left half removed, which is why the reporter resorted to re-running `barman delete` in a loop and `mkdir`-ing the missing directories back.

## The change

`delete()` now checks the directory first. A missing one is warned about and its WALs are reported as deleted, since they are in fact gone.

@martinmarques said on the issue:

> My feeling is that that should be a WARNING and not an exception

and the reporter agreed, so that is what this does. The wording and structure follow `_delete_wal_file()`, which already swallows an `OSError` on an individual file, warns with `Ignoring deletion of WAL file ...`, and lets the caller count it as deleted. Hook scripts still run for every WAL, and the post-delete script receives the error, matching the file-level path.

## Tests

`test_delete_missing_directory` covers the new branch: nothing is listed, no directory or file removal is attempted, both WALs come back in the return value, the warning names the directory, and both hook scripts run with the error passed on.

Two existing tests, `test_delete_wal_files_individually` and `test_delete_whole_directory`, use fictional paths like `/server/wals/0000000100000001` and patch only `os.listdir`. They tripped the new `os.path.isdir` check and failed, so they now patch it too. Flagging that explicitly since it is a change to existing tests rather than only additions.

## Verification

- `pytest tests/test_wal_archiver.py -k delete` — 11 passed
- `pytest tests/test_wal_archiver.py tests/test_backup.py` — 283 passed, 2 failed
- Those 2 failures are **pre-existing**: `test_delete_cloud_backup_data_check_object_lock[True]` and `test_delete_cloud_backup_data_check_object_lock_non_s3_warning` fail identically with my changes stashed on a clean checkout
- Reverting only `src/barman/wal_archiver.py` and keeping the tests fails exactly one test, `test_delete_missing_directory`
- `ruff check` and `ruff format --check` clean on both files. The `E501` lines a stray `flake8` run reports in that file are pre-existing and untouched

I could not install the project with its pinned `psycopg2` (no `pg_config` on this machine), so I ran the suite against `src/` with `--no-deps`. Nothing in this change touches the PostgreSQL client path.

## AI disclosure

Prepared with Claude Opus 5 in Claude Code: it traced the traceback to the `os.listdir` call, wrote the guard and the test, and drafted this description. I reviewed the diff and ran everything listed above. Per the AI policy in this repository, I can explain the change and stand behind it.

### DIFF
--- a/tests/test_wal_archiver.py
+++ b/tests/test_wal_archiver.py
@@ -1944,7 +1944,10 @@ class TestLocalWalStorageStrategy:
 
     @patch("barman.wal_archiver.LocalWalStorageStrategy._delete_wal_file")
     @patch("barman.wal_archiver.os.listdir")
-    def test_delete_wal_files_individually(self, mock_listdir, mock_delete_file):
+    @patch("barman.wal_archiver.os.path.isdir", return_value=True)
+    def test_delete_wal_files_individually(
+        self, _mock_isdir, mock_listdir, mock_delete_file
+    ):
         """
         Test that :meth:`delete` correctly deletes specified WAL files individually
         when the whole WAL directory can not be deleted altogether.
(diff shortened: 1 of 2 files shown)


tests/test_wal_archiver.py
```diff
@@ -1968,9 +1971,60 @@ def test_delete_wal_files_individually(self, mock_listdir, mock_delete_file):
...
+        Test that :meth:`delete` warns and carries on when the WAL directory has
+        already been removed by other means, instead of raising (GH-1213).
+        """
+        # GIVEN two WALs whose directory no longer exists
+        wal_info1, wal_info2 = MagicMock(), MagicMock()
+        wal_info1.name = "000000010000000000000001"
+        wal_info2.name = "000000010000000000000002"
+        wals_to_delete = {"/server/wals/0000000100000001": [wal_info1, wal_info2]}
+        wal_storage = LocalWalStorageStrategy(
+            build_backup_manager(name="TestServer"), None
```
Mavadocasamigos: https://www.postgresql.org

